PR #331, opened 2020-01-23
add command setlocalvar and checklocalvar

KnowOne134:
**_I affirm:_**

add the commands to check and set local vars on any entity in normal field and in instanced zones

### DIFF
--- a/scripts/commands/checklocalvar.lua
+++ b/scripts/commands/checklocalvar.lua
@@ -1,5 +1,5 @@
 ---------------------------------------------------------------------------------------------------
--- func: checklocalvar <varName> <player/mob/npc> <ID>
+-- func: checklocalvar <varName> {'player'/'mob'/'npc'} {name/ID}
 -- desc: checks player or npc local variable and returns result value.
 ---------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


scripts/commands/checklocalvar.lua
```diff
@@ -0,0 +1,70 @@
...
+
+    if (varName == nil) then
+        error(player, "You must provide a variable name.");
+        return;
+    end
+
+    if arg2 ~= nil and arg3 == nil then
+        error(player, "no ID given for target")
+        return
+    end
```
ibm2431: Might be able to make this a little more clear by bringing this up to form an `else` for the `arg2 == nil` check above:

```lua
    if arg2 == nil then
        targ = player:getCursorTarget()
        if targ == nil then
            error(player, "no target")
            return
        end
    elseif arg3 == nil then
        error(player, "no name or ID given for target")
        return
    end
```

scripts/commands/checklocalvar.lua
```diff
@@ -0,0 +1,70 @@
...
+
+    if zone == 55 or zone == 56 or zone == 60 or zone == 63 or zone == 66 or zone == 69 or (zone >= 73 and zone<= 77) then
+        local instance = player:getInstance()
+        if entity == 'npc' then
+            targ = instance:getEntity(bit.band(arg3, 0xFFF), dsp.objType.NPC)
```
ibm2431: I'm not too familiar with how instance functions work, but if `arg3` is defined as a string (`parameters = 'sss'`), does it need to be cast to a number before doing bitwise operations on it?

(And `dsp.` is now `tpz.`)

scripts/commands/setlocalvar.lua
```diff
@@ -0,0 +1,79 @@
...
+            targ = instance:getEntity(bit.band(Name, 0xFFF), dsp.objType.MOB)
+        elseif entity == 'player' then
+            targ = GetPlayerByName(Name)
+        end
+        if targ == nil then
+            error(player, "Target is invalid")
+            return
+        end
+        targ:setLocalVar(varName, varNumber)
+        player:PrintToPlayer(string.format("%s's variable '%s' : %i", targ:getName(), varName, varNumber));
```
ibm2431: Can save _six_ lines by breaking this outside of the parent if (so below the `end` for the is-in-instance check) and cutting lines L72-L77 below~!

scripts/commands/checklocalvar.lua
```diff
@@ -0,0 +1,70 @@
...
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "sss"
+};
+
+function error(player, msg)
+    player:PrintToPlayer(msg);
+    player:PrintToPlayer("!checklocalvar <variable name> optional <player/mob/npc> optional <ID>");
```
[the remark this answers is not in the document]
project-topaz: Let's stick to existing convention for now. When we start the documentation project we might adopt another method for specifying optional variables. Staying consistent will help us find and update them later.

KnowOne134:
All looks good to me

KnowOne134:
Tested and working . Approved

PR #339, opened 2020-02-14
Bst af3 changes

KnowOne134:
**_I affirm:_**

mobs should pop unclaimed, A bst can charm the 2 other mobs and use them in aid of fight, should despawn after some time if you cant beat the fight


KnowOne134:
unfortunately looks like mobMod CHARMABLE isnt working

Wiggo32:
```lua
UPDATE `mob_pools` SET `mobType`='16' WHERE `poolid`='3834';
UPDATE `mob_pools` SET `mobType`='16' WHERE `poolid`='4038';
```
This worked for me... see if it helps

KnowOne134:
The reason that works is it removes its NM tag where the mob mod should make it so a mob is charmable when it's set to not be in database

ibm2431:
It's hitting the `MOBTYPE_NOTORIOUS` check for bind and returning before doing anything else.

https://github.com/project-topaz/topaz/blob/master/src/map/utils/battleutils.cpp#L3855-L3860

Might need to rework it to be an `AND NOT` `mobmod_charmable`.

KnowOne134:
charm and gauge now work on NMs that have the lua mobmod.CHARMABLE set

PR #340, opened 2020-02-15
Tutorial NPCs

Xephera:
**_I affirm:_**
- [X] that I agree to Project Topaz's [Limited Contributor License Agreement](https://github.com/project-topaz/topaz/blob/master/CONTRIBUTOR_AGREEMENT.md), as written on this date


ibm2431:
```
ibm2431: Well, this branch says it's two commits ahead
ibm2431: If those items are really supposed to be there
ibm2431: We'd take those too
ibm2431: So if it's easier for you, you can just fork/PR from that 
and that shop will become updated on our master too as a side-effect
ibm2431: I'll add a note on the PR saying I approved
```

ibm2431:
I have also already stated that I would make style-guide edits, so any reviews from others can ignore those~!

ibm2431:
Pushed style guide associated changes!

I'll give this a proper review tomorrow~

scripts/zones/Bastok_Markets/npcs/Gulldago.lua
```diff
@@ -4,16 +4,116 @@
...
-    player:startEvent(535);
-end;
+function onTrigger(player, npc)
+    local tutorial = player:getCharVar("TutorialProgress")
+    if tutorial == 11 then
+        player:startEvent(535)
+    else
+        local isSkilled = false
+        local mLevel = player:getMainLvl()
+        local hasCrystal = player:hasKeyItem(tpz.ki.DEM_GATE_CRYSTAL)
```
KnowOne134: No need to local this since it only occurs once

scripts/zones/Bastok_Markets/npcs/Gulldago.lua
```diff
@@ -4,16 +4,116 @@
...
+        player:addItem(4376, 6)
+        player:messageSpecial(ID.text.ITEM_OBTAINED, 4376)
+    elseif csid == 522 then
+        player:setCharVar("TutorialProgress", 3)
+    elseif csid == 524 then
+        npcUtil.giveItem(player, {{4096, 1}, {926, 1}, {4370, 1}})
+        player:setCharVar("TutorialProgress", 4)
+    elseif csid == 526 then
+        player:addKeyItem(tpz.ki.CONQUEST_PROMOTION_VOUCHER)
+        player:messageSpecial(ID.text.KEYITEM_OBTAINED, tpz.ki.CONQUEST_PROMOTION_VOUCHER)
```
KnowOne134: No need for the message when using npcutil

scripts/zones/Bastok_Markets/npcs/Gulldago.lua
```diff
@@ -4,16 +4,116 @@
...
+        player:setCharVar("TutorialProgress", 6)
+    elseif csid == 528 then
+        player:addItem(16003)
```
KnowOne134: NpcUtil

scripts/zones/Bastok_Markets/npcs/Gulldago.lua
```diff
@@ -4,16 +4,116 @@
...
+function onTrade(player, npc, trade)
+end
 
-function onTrigger(player,npc)
-    player:startEvent(535);
-end;
+function onTrigger(player, npc)
+    local tutorial = player:getCharVar("TutorialProgress")
+    if tutorial == 11 then
+        player:startEvent(535)
```
KnowOne134: Should this last even have some sort of ending var or complete progress so you dont get stuck in a loop here. Or is this last cs its default text when done forever?

scripts/zones/Konschtat_Highlands/mobs/Amber_Quadav.lua
```diff
@@ -7,4 +7,7 @@ require("scripts/globals/regimes")
 
 function onMobDeath(mob, player, isKiller)
     tpz.regime.checkRegime(player, mob, 83, 1, tpz.regime.type.FIELDS)
+    if player:getCharVar("TutorialProgress") == 7 then
```
KnowOne134: Should we move this to a function inside of the global onmobdeathex?

Xephera:
I think I left the commented old code I was referencing in the bottom of Gulldago's code as well, oops! Didn't notice it until now.

KnowOne134:
I also see these 3 npcs being 1 file in a tutorial global whith just a difference of csid base and shift. They all have the same code with just different csids and in same order correct?

scripts/zones/Bastok_Markets/npcs/Gulldago.lua
```diff
@@ -4,16 +4,116 @@
...
+        local hasCrystal = player:hasKeyItem(tpz.ki.DEM_GATE_CRYSTAL)
+        if tutorial == 3 then
+            for i = 1, 12 do
+                if player:getSkillLevel(i) >= 5 then
+                    isSkilled = true
+                    break
+                end
+            end
+        end
+        if tutorial == 0 or tutorial == nil then
```
ibm2431: `getCharVar` will always return 0 if the variable isn't in the DB, so you don't need to check if it's `nil` here.

(But good looking out for this sort of thing! The binding is just a little weird 😉 )

scripts/zones/Bastok_Markets/npcs/Gulldago.lua
```diff
@@ -4,16 +4,116 @@
...
+    elseif csid == 526 then
+        player:addKeyItem(tpz.ki.CONQUEST_PROMOTION_VOUCHER)
+        player:messageSpecial(ID.text.KEYITEM_OBTAINED, tpz.ki.CONQUEST_PROMOTION_VOUCHER)
+        player:setCharVar("TutorialProgress", 6)
+    elseif csid == 528 then
+        player:addItem(16003)
+        player:messageSpecial(ID.text.ITEM_OBTAINED, 16003)
+        player:setCharVar("TutorialProgress", 7)
+    elseif csid == 530 then
+        player:addExp(1200)
```
ibm2431: One thing we have inherited - until we get a functional modding system set up - are a list of [global settings in a lua file that server operators can modify](https://github.com/project-topaz/topaz/blob/master/scripts/globals/settings.lua) to adjust things like the rate characters earn Exp and Gil.

These settings have had a history of contention - developers having to worry about them adds to the settings' accruing technical debt, but server operators have a valid need to be able to easily adjust things from just the one file. So to avoid another fight over settings, I think that until we get the modding system, we should plan to keep using them where we know they might be expected.

_That backstory out of the way_, what I'm trying to say is:
`player:addExp(1200 * EXP_RATE)` is preferred, so a server admin can increase the Exp this (and all other) quests grant without having to edit all the individual files.

(And `player:addGil(1000 * GIL_RATE` too)

scripts/zones/Konschtat_Highlands/mobs/Amber_Quadav.lua
TeoTwawki: if you did a zone check, that check would still exec to determine true/fall for _every kill game wide._

drk quest kills can be done on any mob in the entire game so that does make sense. it could have been done like breaking a latent except for the 2nd quest that includes kills from the first one making that annoying. In the future maybe a pseudo additional effect on that item. But I'm rewriting those and that's a global too after that, so I don't think you want to go from _every mob killed_ into _every swing_ checking quest stuff.

The best answer for tutorial mobs is probably to either wrap them like regimes do, or use a mixin :disappointed:

scripts/zones/Bastok_Markets/npcs/Gulldago.lua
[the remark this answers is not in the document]
TeoTwawki: you mean variable

variables are good when the data you are variablizing:
* changes
* gets used multiple times
* would make the line where its used wider

Since this data is static the line doesn't change and it is used only once, its cool to do it in-line.

PR #341, opened 2020-02-15
remove temp items after leaving an instanced zone

KnowOne134:
**_I affirm:_**

found that temp items were only removing from zoning out of an instance in retail captures. logging out and in from or crashing in an instance retained items

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -306,11 +306,6 @@ void SmallPacket0x00A(map_session_data_t* session, CCharEntity* PChar, CBasicPac
     PChar->pushPacket(new CZoneInPacket(PChar, PChar->m_event.EventID));
     PChar->pushPacket(new CZoneVisitedPacket(PChar));
 
-    if (!PChar->loc.zoning)
-    {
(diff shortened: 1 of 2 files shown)


TeoTwawki:
temp items clear out in normal zones to. the only place they retain is in abyssea to my knowledge. I am not sure they even have retail reloading behavior yet (didn't look) but I suspect they should just always clear out on leaving a zone and for the other places (abyssea, besieged) they get loaded in based on a condition of some sort - be it a variable or what have you saying.

KnowOne134:
My intention here and has been tested that any assault where you dc or logout and back into an instance where you havent changed zone you do retain every temp item you have, in abyssea it reloads a new set of temp items. And I believe anywhere else in the game they just clear out. Test in nyzul buying items and ones loaded from box. Force dc and logout and in and all temp items retained. In assault you always keep your fireflies.

TeoTwawki:
I think I misread this
```
 if (m_zoneType != ZONETYPE_DUNGEON_INSTANCED)
```
as == actually

for some derpy reason I was thinking of nyzul's temp items, and those teleports probably don't "zone" you.

PR #343, opened 2020-02-16
globaliz all the cells for in use in salvage.

KnowOne134:
…vage global file i will add to for future use on entering, exit, ect

**_I affirm:_**

the global file will also play a big part for entering exiting and many other functions to come that are already planned

### DIFF
--- a/scripts/globals/abilities/activate.lua
+++ b/scripts/globals/abilities/activate.lua
@@ -22,15 +22,5 @@ function onAbilityCheck(player, target, ability)
 end
 
 function onUseAbility(player, target, ability)
-    local effect = tpz.effect.DEBILITATION
-
-    player:spawnPet(tpz.pet.id.AUTOMATON)
-
-    if player:hasStatusEffect(effect) then
-        local pet = player:getPet()
-        local statusEffect = player:getStatusEffect(effect)
-        local power = statusEffect:getPower()
-        local duration = math.floor(statusEffect:getTimeRemaining()/1000)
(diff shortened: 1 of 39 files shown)


scripts/globals/abilities/activate.lua
```diff
@@ -11,16 +11,26 @@ require("scripts/globals/pets")
...
     end
 end
 
-function onUseAbility(player,target,ability)
+function onUseAbility(player, target, ability)
+    local effect = tpz.effect.DEBILITATION
+
     player:spawnPet(tpz.pet.id.AUTOMATON)
+
+    if player:hasStatusEffect(effect) then
```
TeoTwawki: it might be worth a helper function to perform this task. do the pets have to be resummoned to update their status in salvage?  know there is a potentially similar situation down the road in voidwatch and also abyssea with buffs getting copied to pets.

KnowOne134: You start salvage with the debuffs when entering. So you summon after the effect is already in place. And when so the pet will gain what ever effect and power you have. Also using a cell will remove the effect from the player amd the pet at same time.

TeoTwawki: I mean synchronizing pets with masters in general though. it might be worthwhile to just have a reusable something we can call to sync their effects up in these situations rather than copying same block code to each pet script. A global func you feed which effects are being copied to and it do the work all in one spot.

TeoTwawki: I'm probably not describing it well. think mixin like.

PR #345, opened 2020-02-16
add mob's target as an optional parameter ("attacker") to OnCriticalHit and its related hook.

TeoTwawki:
**_I affirm:_**

nothing currently really _needs_ this yet, but something I was working on could have been made simpler instead of my using `mobb:getTarget()` to grab a table of everything on its hatelist yesterday when I really just need the current target...and I would wager a retail use case for it will show up eventually.

Oddly the hook (listener) is _inside the binding_ instead of in the spot in battleutils that the binding is triggered from. When the listeners were made they were intended to phase out the bindings, though I personally like the bindings better. ¯\\\_(ツ)\_/¯ 

So..I did this all at the binding itself. It now occurs to me I'm getting the target of a target..The alternative is I go in battleutils where its in the context of the players own attack making the mob itself "PTarget" and adding a param for that entity there like..`luautils::OnCriticalHit(PTarget, this);` and then updating all the calls to match while passing it along as `mob, attacker`. I decided to stop and get some input first, but I am leaning toward doing that instead.

### DIFF
--- a/scripts/zones/Wajaom_Woodlands/mobs/Iriz_Ima.lua
+++ b/scripts/zones/Wajaom_Woodlands/mobs/Iriz_Ima.lua
@@ -15,7 +15,7 @@ function onMobSpawn(mob)
     mob:setLocalVar("BreakChance", 5)
 end
 
-function onCriticalHit(mob, player)
+function onCriticalHit(mob, attacker)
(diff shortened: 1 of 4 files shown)


TeoTwawki:
Redid it the way I was talking about last night

PR #346, opened 2020-02-16
Add `Pinecone Bomb` blue magic spell.

mrhappyasthma:
Now that the order of operations has been cleaned up for spells,
Pinecone Bomb can be properly implemented. It does damage to the mob
(which may wake it up from a previous slept state), but then can reapply
sleep as described here: https://ffxiclopedia.fandom.com/wiki/Pinecone_Bomb


ibm2431:
Hello, @mrhappyasthma! Thank you for contributing to Project Topaz!

So that we don't use our parent project's brand, we've changed some function and table names to reflect adopting a new brand. I remember seeing that this PR was started prior to Topaz and originally submitted to our parent project - I took the liberty of updating these references for you; you don't need to sweat over them!

Although, I realize after I submitted the push that I should have asked first to make sure I didn't step on your toes. I'm sorry if I did! My intent was to make your contribution experience as painless as possible! 😅 

That said, I haven't had the chance to take a look at the code yet. We try to be relatively swift with reviews though, so I look forward to working with you soon!

mrhappyasthma:
No worries. I look forward to your review.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,65 @@
...
+        local sleepParams = {}
+        sleepParams.attribute = tpz.mod.INT
+        sleepParams.bonus = 0
+        sleepParams.effect = tpz.effect.SLEEP_II
+        sleepParams.skillType = tpz.skill.BLUE_MAGIC
+        local resist = applyResistanceEffect(caster, target, spell, sleepParams)
+        if (resist > 0.5) then -- Apply the sleep
+            local power = 2
+            local tick = 0
+            local duration = 90 * resist
```
ibm2431: Looking at the description for the spell, it claims that duration is supposed to vary depending on TP. At the moment, this is basing the duration off of attacker INT versus target resistance.

1) Is it supposed to _always_ be affected by TP?
2) Is TP -> duration effect only applicable when under Chain Affinity?
3) What's the base duration of this type of sleep? With or without Chain Affinity?
4) What's the sleep duration at certain TP values? Minimum, maximum?

@Wiggo32 I looked at your capture of you learning Pinecone Bomb and testing it out, but there was no explicit duration testing (the sapling was auto-attacked awake, and the draugar was unaffected)

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,65 @@
...
+function onSpellCast(caster,target,spell)
+    local params = {}
+    -- This data should match information on http://wiki.ffxiclopedia.org/wiki/Calculating_Blue_Magic_Damage
+    params.tpmod = TPMOD_CRITICAL
+    params.dmgtype = DMGTYPE_PIERCE
+    params.scattr = SC_LIQUEFACTION
+    params.numhits = 1
+    params.multiplier = 1.25
+    params.tp150 = 1.25
+    params.tp300 = 1.25
```
ibm2431: [Bgwiki has these ftp values as 2.25](https://www.bg-wiki.com/bg/Pinecone_Bomb). My reading of [BluefTP](https://github.com/project-topaz/topaz/blob/master/scripts/globals/bluemagic.lua#L302) has these values coming out as 1.25~!

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,65 @@
...
+-----------------------------------------
+
+function onMagicCastingCheck(caster,target,spell)
+    return 0
+end
+
+function onSpellCast(caster,target,spell)
+    local params = {}
+    -- This data should match information on http://wiki.ffxiclopedia.org/wiki/Calculating_Blue_Magic_Damage
+    params.tpmod = TPMOD_CRITICAL
```
ibm2431: I was going to say that the TP mod for this spell doesn't seem to be critical hit chance, but then I looked at [the blue magic global](https://github.com/project-topaz/topaz/blob/3df25d52158be6d44dfe84f198194b6822b5ebb8/scripts/globals/bluemagic.lua) and it doesn't look like tpmod params are used _at all_.

I did some searches and [found a definition for `params.tpmod = TPMOD_DURATION` for Battle Dance](https://github.com/project-topaz/topaz/blob/3df25d52158be6d44dfe84f198194b6822b5ebb8/scripts/globals/spells/bluemagic/battle_dance.lua), which seems like what we should be using here (again, this param is never checked), but [that specific global value doesn't seem to be defined anyway](https://github.com/project-topaz/topaz/blob/master/scripts/globals/bluemagic.lua#L4-L9).

I'm not going to expect you to fix that fact that these mods don't seem to do anything, but could you change this to at least _claim_ it's supposed to effect duration?

ibm2431 (changes requested):
I tested this out and it does seem to work!

There a few questions over fTP values, and how the sleep duration is supposed to work. Tagging @Wiggo32 in case he's played with the spell beyond the one capture I could find.

ibm2431:
#269

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: I believe these values are stale from when I branched this code from another spell.

It should be `2.25`, I believe. FFXIclopedia article also agrees. It has it listed as `multiplier` for pinecone bomb: https://ffxiclopedia.fandom.com/wiki/Calculating_Blue_Magic_Damage

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: This was my first time adding a BLU spell, so I'm open to any improvements.

I just copied how sleep was applied in other spells, so it may not be 100% accurate. If you have a reference for how this should be applied, please point me in the right direciton!

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: A lot of these BLU enums seems to be in bad shape. I'm happy to clean them up as needed, but need some guidance.

Would you prefer which/any of the following:

- Remove the TPMOD param since it's unused
- Add TPMOD_DURATION value since it's missing, and then reference it here
- Add usage of `params.tpmod` in the global bluemagic.lua (if so, where)

I think the bottom two are probably the *right* approach, assuming this value is supposed to be used somewhere. But if we felt this value is obsolete or unnecessary, perhaps removing it until it's implemented would be less ambiguous?

Also we should probably file an issue for this problem to track these changes, which can be applied ahead of this PR as needed.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: Also `params.azuretp` does not seem to be used anywhere. From what I can tell. But I'll just set it as `2.25` as well since the damage multiplier seems to be linear.

mrhappyasthma: Added a commit to address this comment.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: Filed https://github.com/project-topaz/topaz/issues/354 to track.

Let me know what your preferences are. I'm happy to do an audit or cleanup based on what is preferred. (Although if we want to use the parameters, I could use some pointers since I'm not that familiar with the code and have no retail FFXI)

scripts/globals/spells/bluemagic/pinecone_bomb.lua
ibm2431: The commit was on the branch for your PR to DSP 😉 

I'll go ahead and mark this comment as resolved though, assuming a similar commit will make it to this PR.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
ibm2431: Yeah, the bottom two are the more time consuming, but overall the best choice. It will require a bit of rework for the Blue Magic global, which desperately needs it.

For the purpose of _this_ PR though, go ahead and add TPMOD_DURATION to both the global and this script (matching the one for Battle Dance), and then after a separate PR hooking up the params to be used, we can work on having spells being able to make use of the mods.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
ibm2431: What we'll probably have to do is find some information - or get a capture - on the basic duration of the effect. Even if we can't fully pick apart how TP affects the duration for this (or other) Blue spells, we can at least apply the bare minimum duration for now and add a note in the script (and create a new Issue) that it's still missing duration bonuses from TP.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: Sounds like a plan. Will update this in a commit.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: I added a commit to both branches. I could have sworn I pushed it, but I must have forgot to `git push` or had a typo and didn't notice it failed.

Should be there now.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: Commit added. Please take a look.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: Sounds good.

Should I wait on someone to do a "capture"? Or should I go forward with a TODO note and reference a github issue to update the TP modification of the sleep duration?

scripts/globals/bluemagic.lua
```diff
@@ -7,6 +7,7 @@ TPMOD_CRITICAL = 1;
 TPMOD_DAMAGE = 2;
 TPMOD_ACC = 3;
 TPMOD_ATTACK = 4;
+TPMOD_DURATION = 5;
```
mrhappyasthma: I left the semicolon here for now. I'm gonna make a new PR to remove all semicolons from this file, since it violates the style recommendations for Lua.

mrhappyasthma:
Question: Do you want me squashing these commits? Or do the mergers handle that for this repo?

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,65 @@
...
+
+function onMagicCastingCheck(caster,target,spell)
+    return 0
+end
+
+function onSpellCast(caster,target,spell)
+    local params = {}
+    -- This data should match information on http://wiki.ffxiclopedia.org/wiki/Calculating_Blue_Magic_Damage
+    params.tpmod = TPMOD_DURATION
+    params.dmgtype = DMGTYPE_PIERCE
```
mrhappyasthma: Need to update this to `damageType` after https://github.com/project-topaz/topaz/commit/b4ed738e8901e0127ffe19b855d6a886f82c476e

ibm2431:
We can go ahead and squash for you~!

mrhappyasthma:
Awesome. Just wanted to verify! :)

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: Done.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
mrhappyasthma: We have some captures. I'll check them out and see what I can determine. Then hopefully someone more experienced can double check me.

scripts/globals/spells/bluemagic/pinecone_bomb.lua
```diff
@@ -0,0 +1,65 @@
...
+        sleepParams.attribute = tpz.mod.INT
+        sleepParams.bonus = 0
+        sleepParams.effect = tpz.effect.SLEEP_II
+        sleepParams.skillType = tpz.skill.BLUE_MAGIC
+        local resist = applyResistanceEffect(caster, target, spell, sleepParams)
+        if (resist > 0.5) then -- Apply the sleep
+            local power = 2
+            local tick = 0
+            local duration = 90 * resist
+            target:addStatusEffect(sleepParams.effect, power, tick, duration)
```
mrhappyasthma: We also need to scale this duration by the resist value.

mrhappyasthma: Oh wait, we take account into that in the local above. So it should be good.

scripts/globals/bluemagic.lua
mrhappyasthma: Rebasing and removing semicolons.

mrhappyasthma:
Rebased and closed out some of the outstanding comments.

This PR is still waiting on a more accurate TP-based sleep duration.

mrhappyasthma:
Corrected the attackType to be `RANGED`.

Still need to do a bit of research for the TP-based sleep. Been crazy with work and will be having more travel over the coming weeks. Whenever I can find some downtime I can plot some of the values from the captures and find a reasonable equation.

ibm2431:
Compiled a spreadsheet from the two captures in which Chain Affinity was used:
[Pinecone-Bomb.zip](https://github.com/project-topaz/topaz/files/4363102/Pinecone-Bomb.zip)

mrhappyasthma:
BTW I still haven't forgotten about this. Just a crazy time we live in these days... Whenever I get some time I'll dig into those spreadsheets and compile any other data that is needed.

ibm2431 (approved):
Pushed a commit based on added findings from Wiggo's most recent "base" capture:
<https://docs.google.com/spreadsheets/d/1lJafURoeyiZ6NAz6A8NpdSk9VMjD0JLqjtxIa190RWs/>

As strange as it seems, it's looking like it's close to a simple inverse bell curve between 15 and 60 seconds. This is based on the Qufim and Yuhtunga testing, where Wiggo is way over the cap of the mobs, so resists aren't coming into play. And when the spell fails, it fails outright - no damage at all. I'm relying on the damage check here.

It may seem weird, but an inverse bell fits the duration data well - and I'm fairly confident in the accuracy of this determination - so that's what I'm going with absent evidence to the contrary. I broke it out into it's own function in case we come across similar data later and end up needing to copy-paste the function into the blue global.

How TP affects duration can be determined at a later date; the blue magic global doesn't utilize TP modifications for _any_ blue spells at the moment anyway.

Approved. Sorry for the wait, mrhappyasthma.

mrhappyasthma:
Wow thanks for digging so deeply into this. I know my contributions have been sporadic and irregular. So I really appreciate you doing all this work on behalf of this PR.

PR #348, opened 2020-02-16
move tractor fail message to onMagicCastingCheck()

TeoTwawki:
**_I affirm:_**

(cherry picked from commit 695a2bcf646b4a0b98e5017774d6c2938358a566 in my personal project, under GPL)

The existing tractor check in magic_state.cpp is crashy. It has been the cause of another servers recent crashes. This morning I sorted out the cause, and decided to move the check into lua anyway since not all the spell functions are bound yet and we normally set the message right there in the onMagicCastingCheck() anyway and no longer have to check if the object we're sending the message to is a player or not as that's baked into the function. 

Also included: 
kill magic number in alexander.lua
translated a comment to english

### DIFF
--- a/scripts/globals/spells/alexander.lua
+++ b/scripts/globals/spells/alexander.lua
@@ -13,7 +13,7 @@ function onMagicCastingCheck(caster,target,spell)
     if (not caster:canUseMisc(tpz.zoneMisc.PET)) then
         return tpz.msg.basic.CANT_BE_USED_IN_AREA
     elseif (not caster:hasStatusEffect(tpz.effect.ASTRAL_FLOW)) then
-        return 581
+        return tpz.msg.basic.MAGIC_MUST_ASTRAL_FLOW
     elseif (caster:hasPet()) then
         return tpz.msg.basic.ALREADY_HAS_A_PET
     elseif (caster:getObjType() == tpz.objType.PC) then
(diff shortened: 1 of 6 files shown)


TeoTwawki:
So guys, the new binding was not at all needed because turns out there was something in the core I had missed and I cannot remember what else I thought I'd use it for. Its really easy to create..Should I remove that until a use case exists?

ibm2431:
I'd be okay with adding the binding now even if there's nothing using it if there's an example of what _might_ use it in the future.

I'm trying to think of spells that might not always be allowed when general spellcasting is. Trusts? Geomancy?

ibm2431:
Although now I'm wondering; [we have player:canUseMisc](https://github.com/project-topaz/topaz/blob/3df25d52158be6d44dfe84f198194b6822b5ebb8/src/map/lua/lua_baseentity.cpp#L5065). Is there a distinction to be made between checking the flags "as a player" versus checking it "as the spell"?

TeoTwawki:
Contexts in which we have no player but do have a spell? 
But this turned out how it did because reused existing darkstar code [which acts on the spell](https://github.com/project-topaz/topaz/blob/2943fab983cc1457ce75f15dd9dca9d3bfa21b05/src/map/spell.cpp#L177) so I just rolled with it.

TeoTwawki:
only one I could think of was mazurka, which is presently handled elsewhere but I didn't see where.

PR #349, opened 2020-02-18
Fix starburst and sunburst element param always being the same due to being overwritten

TeoTwawki:
**_I affirm:_**

[Spotted some derp in sunburst.lua](https://github.com/project-topaz/topaz/blob/2943fab983cc1457ce75f15dd9dca9d3bfa21b05/scripts/globals/weaponskills/sunburst.lua#L24)
`params.ele = tpz.magic.ele.DARK params.ele = tpz.magic.ele.LIGHT`
its just always going to be light because its being replaced immediately.
Also reformatted a bit because I dislike wide lines.


TeoTwawki:
same bug was in starburst so just amended to include it to.

PR #355, opened 2020-02-19
Audit all damage type values in BLU spells.

mrhappyasthma:
Corrects /adds any inaccurate or missing damage types.

Also updated the param value in the Lua table to use `damageType`
instead of `dmgtype`. This better matches the enum value naming
convention for `tpz.damageType`.

**_I affirm:_**


mrhappyasthma:
NOTE: This will conflict with https://github.com/project-topaz/topaz/pull/346, but there's some open comments still on that PR so this can probably go in head of it since it should be more straightforward.

PR #356, opened 2020-02-19
RecastIDs were wrong for higher Curing Waltz Tiers.

### LINKED ISSUE #353: Curing waltzes on DNC apply cool down incorrectly
**_I have:_**

server branch: master

Steps:  Be DNC or /DNC
use any tier Curing Waltz other than the first one.  The cool down will apply to Curing Waltz instead of the correct ability.  The cool down applied appears to be correct, however.  This is exploitable in that it would allow a DNC to spam higher level Curing Waltzes as long as the have the TP.

Expected Behavior:  The cooldown should be applied to the skill used in this case.  For example, If I use Curing Waltz III, that should go on CD, not Curing Waltz.

SirGouki:
**_I affirm:_**


ibm2431 (approved):
Hello, SirGouki! Thank you for contributing to Project Topaz!

Everything looks good. Just wanted to officially welcome you, and say that we look forward to working with you in the future!

PR #357, opened 2020-02-19
Fix various issues with linkshells

cocosolos:
This adds several features to make linkshells match retail behavior. The biggest change is using the 9th byte of a linkshell items extra data to hold the linkshell type, rather than using an offset of the item id as is currently done (this appears to be how retail does it). Also added is the ability to change /lsmes privilege level to LS holder only, pearlsack holders, or everyone. These changes should require no input to existing shells for compatibility. Breaking the linkshell group has also been added. This will kick and break the pearls/sacks of all equipped members, then set the linkshell as broken in the database. Any offline or unequipped members will receive a message when they next try equipping which will then break the item. There has been a !breaklinkshell command added for GMs to break any linkshell group. ***This change DOES require that you add a column to the linkshells table.***

* break linkshell and all pearls/sacks by dropping linkshell or with gm command
* pearlsack kick breaks equipped pearl, linkshell kick breaks all pearls/sacks in all inventories
* lsmes privilege (/lsmes level ls, /lsmes level ps, /lsmes level all, /lsmes clear)
* equipable from any local inventory (inventory, satchel, sack, case)

Fixes pretty much everything in #78 except trading pearls.

**_I affirm:_**

### DIFF
--- a/migrations/broken_linkshells.py
+++ b/migrations/broken_linkshells.py
@@ -1,0 +1,22 @@
+import array
+import mysql.connector
+
+def migration_name():
+	return "Adding broken column to linkshells table"
+
+def check_preconditions(cur):
+	return
+
+def needs_to_run(cur):
+	# Ensure broken column exists in linkshells
+	cur.execute("SHOW COLUMNS FROM linkshells LIKE 'broken'")
+	if not cur.fetchone():
+		return True
+	return False
+
+def migrate(cur, db):
+	try:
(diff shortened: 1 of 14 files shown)


KnowOne134 (approved):

KnowOne134:
tested, all works as long as its equipped in LS slot 1, you can promote demote and kick players that have pearl equipped in slot 2. all other tested worked out

PR #358, opened 2020-02-19
Move battlefield capacity check to correct place.

cocosolos:
getBattlefieldIdByBit is being passed a mask instead of a single battlefield bit, which is then being passed into battlefieldAtCapacity. This fix moves battlefieldAtCapacity to the correct place in findBattlefields and each battlefield is checked individually.

**_I affirm:_**

### DIFF
--- a/scripts/globals/bcnm.lua
+++ b/scripts/globals/bcnm.lua
@@ -739,7 +739,7 @@ function findBattlefields(player, npc, itemId)
         return 0
     end
     for k, v in pairs(zbfs) do
-        if v[3] == itemId and checkReqs(player, npc, v[2], true) then
+        if v[3] == itemId and checkReqs(player, npc, v[2], true) and not player:battlefieldAtCapacity(v[2]) then
             mask = bit.bor(mask,math.pow(2,v[1]))
         end
     end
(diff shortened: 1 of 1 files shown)


ibm2431 (commented):
I'll agree the capacity check shouldn't be in the trigger or trade functions. I'm concerned that if it's put in `findBattlefields`, "full" battlefields won't show up in the list presented to the player (ie: you can't get a "battlefield is at capacity" message, because you have no full battlefields to select).

Should the capacity check go in `EventUpdateBCNM` instead?

cocosolos:
The capacity check isn't for players in the battlefield, but if there are any open arenas (normally out of 3) to host a battle. I'm not sure if they're still supposed to show up in the list if all arenas are occupied, but it ends up doing the same thing: blocking a BCNM from starting if there's no arena to host it.

ibm2431:
Letting others know that cocosolos and I spoke; I'll check the all occupied behavior in a few days.

ibm2431 (approved):
Others: I did some experimenting with BCNM behavior during free period. There are _various_ oddities regarding the BCNM system which I'm not about to make _this_ PR start addressing. Hit me up on Discord if you're curious.

@cocosolos So now we're sure that - with a few exceptions (ie: Limbus) - a player is _not_ presented with an option for a BCNM if all arenas are full, so this PR is legit. I don't _think_ the current system is spitting out the "all full, please wait" message that retail does, but that can be addressed later. I'm good with merging this unless you have any other last minute thoughts~!

PR #360, opened 2020-02-20
Bulk remove semicolons from Lua scripts/commands.

mrhappyasthma:
This may not actually be every file, but I did the best I could.

Command applied:
    find . -type f -name "*.lua" | xargs -d '\n' sed -i 's/;//g'

Semicolons are not appropriate in Lua per the style guidelines:
    https://github.com/project-topaz/topaz/blob/master/CONTRIBUTING.md

**_I affirm:_**

### DIFF
--- a/scripts/commands/addallattachments.lua
+++ b/scripts/commands/addallattachments.lua
@@ -7,7 +7,7 @@ cmdprops =
 {
     permission = 1,
     parameters = "s"
-};
+}
 
 local ValidAttachments = {
   8193, 8194, 8195, 8196, 8197, 8198, 8224, 8225, 8226, 8227,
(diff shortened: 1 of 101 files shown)


mrhappyasthma:
There's so many Lua files, I figured we can do it directory by directory. Otherwise it's ridiculous.

(For example: https://github.com/mrhappyasthma/topaz/commit/7277ac15d61230d0093e8cde29057dbefa00bd69)

ibm2431:
And here I was considering doing the same thing and racking up some line diff counts. You beat me to it. 😉

KnowOne134:
That was just commands. Still alot more to do

mrhappyasthma:
Correct. It was too massive of a change to do in a single PR, so there's plenty of room to help out.

I can progressively send out PRs for this but if anyone wants to help out then even better :)

PR #361, opened 2020-02-20
Added Omens Quest (BLU AF2)

kaincenteno:
Changed BEGINNINGS to check correct AF1 global
Position of flans not accurate to retail as there's no captures for a full alliance fighting BCNM
renamed blank npc to Blu_Af2 so it's easier to tell what purpose it serves.
Immortal Flans do not have any special AI or move sets that differences them from other flans.
Thank you @Wiggo32  for capturing this quest.

**_I affirm:_**


scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -2,17 +2,50 @@
...
+require("scripts/globals/npc_util")
 -----------------------------------
 
 function onTrade(player,npc,trade)
 end
 
 function onTrigger(player,npc)
-    player:startEvent(770)
+    -- OMENS
+    if player:hasKeyItem(tpz.ki.SEALED_IMMORTAL_ENVELOPE) then
```
KnowOne134: Just for readability of the code could you add local getqueststatus(..., omens) and then add that to the check with has envelope.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -2,17 +2,50 @@
...
+        if player:getCharVar("BluAf2CS") == 3 then
+            player:startEvent(714) -- BLU AF2: Master location
+        elseif player:getCharVar("BluAf2CS") == 4 then
+            player:startEvent(715) -- BLU AF2: Reminds master location
+        elseif player:getCharVar("BluAf2CS") == 5 then
+            player:startEvent(716) -- BLU AF2: Quest Completed
+        end
+
+    -- DEFAULT DIALOG AFTER COMPLETING OMENS
+    elseif player:getQuestStatus(AHT_URHGAN,tpz.quest.id.ahtUrhgan.OMENS) == QUEST_COMPLETED then
```
KnowOne134: Can use the above local here also then

scripts/globals/bcnm.lua
```diff
@@ -573,6 +573,7 @@ function checkReqs(player, npc, bfid, registrant)
         [1090] = function() return ( player:hasKeyItem(tpz.ki.TOGGLE_SWITCH)                                                                                                ) end, -- Quest: Puppetmaster Blues
         [1091] = function() return ( mjob == tpz.job.COR and mlvl >= 66                                                                                                     ) end, -- Quest: Breaking the Bonds of Fate (COR LB5)
         [1092] = function() return ( toau == mi.toau.LEGACY_OF_THE_LOST                                                                                                     ) end, -- TOAU35: Legacy of the Lost
+        [1122] = function() return ( player:getQuestStatus(AHT_URHGAN,tpz.quest.id.ahtUrhgan.OMENS) == QUEST_ACCEPTED and player:getCharVar('BluAf2CS') == 1                ) end, -- Quest: Omens (BLU AF2)
```
KnowOne134: Can we change the charger to the quest name. Quest_Omens for all scripts? Thoughts?

scripts/zones/Navukgo_Execution_Chamber/mobs/Immortal_Flan.lua
```diff
@@ -0,0 +1,19 @@
+-----------------------------------
```
KnowOne134, line 1: No need for this script as it has no info at this time

sql/npc_list.sql
```diff
@@ -6253,7 +6253,7 @@ INSERT INTO `npc_list` VALUES (17056402,'Wathdeeh','Wathdeeh',0,0.000,0.000,0.00
 INSERT INTO `npc_list` VALUES (17056403,'Wawaroon','Wawaroon',0,0.000,0.000,0.000,0,40,40,0,0,0,6,25,0x0000C70600000000000000000000000000000000,32,'TOAU',1);
 INSERT INTO `npc_list` VALUES (17056404,'qm9','???',130,-405.888,6.514,-440.237,32769,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'TOAU',0);
 INSERT INTO `npc_list` VALUES (17056405,'14','     ',0,0.000,0.000,0.000,0,40,40,0,0,0,2,2075,0x0100000000100020003000400050006000700000,32,'TOAU',1);
-INSERT INTO `npc_list` VALUES (17056406,'15','     ',225,342.129,36.509,-24.856,32769,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'TOAU',1);
+INSERT INTO `npc_list` VALUES (17056406,'Blu_Af2','     ',225,342.129,36.509,-24.856,32769,40,40,0,0,0,0,3,0x0000340000000000000000000000000000000000,0,'TOAU',1);
```
KnowOne134: Is this the only thing this npc in the game does is deal with blu quest?

KnowOne134 (commented):
Thanks for the work done. Have a few critiques before I actually test and to hear from others

sql/npc_list.sql
kaincenteno: As far as i know it is only used for BLU AF2 quest. It's a blank npc on the floor. I can change it back but just named it before as on other PRs i submitted before i was asked to name it something related to what it is used for if its a blank or duplicate. (was just trying to avoid having to have discussions about it lol but again no issues changing it back just need confirmation :cat:  )

scripts/zones/Navukgo_Execution_Chamber/mobs/Immortal_Flan.lua
kaincenteno: Should I just delete it?  :scream_cat:

scripts/globals/bcnm.lua
kaincenteno: Charger? what does that mean? :sweat_smile:

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
kaincenteno: For sure but the check for envelope is only being used once. From old PRs I have submitted on the other project that shall not be named, I was told to do that only if it is being used more than once for readability :cat: 

Tagging @ibm2431  for second opinion.

sql/npc_list.sql
kaincenteno: Tagging @TeoTwawki  for his insight in naming unnamed npcs :cat:

scripts/globals/bcnm.lua
KnowOne134: Sorry. Phone and auto correct. 
CharVar is what that was meant to be

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
KnowOne134: Just the readability part. Like just looking at the code I wouldn't know it's for this quest because the check is just an envelope ki that I don't know where it comes from

scripts/zones/Navukgo_Execution_Chamber/mobs/Immortal_Flan.lua
KnowOne134: Git delete yes. Unless you want to add something into the lua for the mob

scripts/globals/bcnm.lua
kaincenteno: Sure we can, that will help with readability :cat:
I'll change it today

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
ibm2431: General rule was that you could assign a variable if it makes an unreadably long line shorter, but...

I feel bad for `player:hasCompletedQuest()`, everyone always forgets it exists.

(edit: This is in reference to assigning a local to quest status to shorten lines when checking it against QUEST_COMPLETE)

sql/npc_list.sql
TeoTwawki: it has no other uses, name doesn't much matter since blank in game except for being readable/recognizable to us humans.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -29,9 +29,10 @@ function onTrigger(player,npc)
     local anEmptyVesselProgress = player:getCharVar("AnEmptyVesselProgress")
     local divinationReady = vanaDay() > player:getCharVar("LastDivinationDay")
     local beginnings = player:getQuestStatus(AHT_URHGAN,tpz.quest.id.ahtUrhgan.BEGINNINGS)
+    local omens = player:getQuestStatus(AHT_URHGAN,tpz.quest.id.ahtUrhgan.OMENS)
```
ibm2431: You're squishing your commas~!

(In lots of places)

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -29,9 +29,10 @@ function onTrigger(player,npc)
...
 
     -- AN EMPTY VESSEL
-    if ENABLE_TOAU == 1 and anEmptyVessel == QUEST_AVAILABLE and anEmptyVesselProgress <= 1 and player:getMainLvl() >= ADVANCED_JOB_LEVEL then
+    if ENABLE_TOAU == 1 and anEmptyVessel == QUEST_AVAILABLE and anEmptyVesselProgress <= 1 and player:getMainLvl() >= AF1_QUEST_LEVEL then
```
ibm2431: Empty Vessel is the BLU unlock quest

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -73,6 +74,24 @@ function onTrigger(player,npc)
             player:startEvent(706,player:getGil()) -- clue about the five staging points, costs you 1000 gil
         end
 
+    -- OMENS
+    elseif beginnings == QUEST_COMPLETED and omens == QUEST_AVAILABLE and player:getMainJob() == tpz.job.BLU and player:getMainLvl() >= AF2_QUEST_LEVEL then
+        if not divinationReady then
+            player:startEvent(63)
+        elseif player:needToZone() then
```
ibm2431: Do me a favor and use `setLocalVar` and `getLocalVar` instead. I don't know if Waould is already using one if you've refused a divination (and if not, could you adjust that?), but `needToZone` should be phased out and eventually removed because it's "zone global" without a specified reason.

edit: Or, alternatively, be refactored to _take_ a reason and keep them straight.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -73,6 +74,24 @@ function onTrigger(player,npc)
...
+            player:startEvent(78,player:getGil()) -- dummy questions, costs you 1000 gil
+        else
+            player:startEvent(710) -- start AF2 quest
+        end
+    elseif omens == QUEST_ACCEPTED then
+        if player:getCharVar("BluAf2CS") == 1 then
+            player:startEvent(711,player:getGil()) -- clue about bcnm location, costs you 1000 gil
```
ibm2431: Does the event handle denying the player if they don't have enough gil? Or do you need to?

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -158,5 +177,19 @@ function onEventFinish(player,csid,option)
         player:messageSpecial(ID.text.PAY_DIVINATION) -- You pay 1000 gil for the divination.
     elseif csid == 707 then
         npcUtil.completeQuest(player, AHT_URHGAN, tpz.quest.id.ahtUrhgan.BEGINNINGS, {item=17717})
+
+    -- OMENS
+    elseif csid == 710 then
+        player:addQuest(AHT_URHGAN,tpz.quest.id.ahtUrhgan.OMENS)
+        player:setCharVar("BluAf2CS",1)
+    elseif csid == 711 and option == 1 then
+        player:delGil(1000)
```
ibm2431, line 190: Either way, be sure to check the player's gil!

I think delGil won't take gil below 0, but there's still the message below, and it's good to keep the habit~

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
```diff
@@ -0,0 +1,57 @@
...
+function onBattlefieldTick(battlefield, tick)
+    tpz.battlefield.onBattlefieldTick(battlefield, tick)
+end
+
+function onBattlefieldRegister(player, battlefield)
+end
+
+function onBattlefieldEnter(player, battlefield)
+    local mobOffset = (battlefield:getArea() - 1) * 7  -- Offset to spawn correct mob depending on battlefieldNumber
+    local players = battlefield:getPlayers()
```
ibm2431: So question: Do additional flans spawn in front of an inside player's eyes as more players enter the battlefield? Should we be checking party numbers as players enter, or when the battlefield is initially set up?

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -73,6 +74,24 @@ function onTrigger(player,npc)
...
+            player:startEvent(78,player:getGil()) -- dummy questions, costs you 1000 gil
+        else
+            player:startEvent(710) -- start AF2 quest
+        end
+    elseif omens == QUEST_ACCEPTED then
+        if player:getCharVar("BluAf2CS") == 1 then
+            player:startEvent(711,player:getGil()) -- clue about bcnm location, costs you 1000 gil
+        elseif player:getCharVar("BluAf2CS") == 2 then
+            player:startEvent(712) -- gives keyitem to claim armour piece
+        elseif player:getCharVar("BluAf2CS") == 3 then
```
ibm2431: Would you happen to know if Waoud offers a divination after speaking with Lathuya? This startEvent here is correct based off of Wiggo's capture, but he didn't speak with Waoud again between Lathuya and completion. (I wouldn't have either.)

sql/npc_list.sql
ibm2431: Unless ships are involved. Don't ask me why.

(And I'd quip that "Omens" would be more clear than "Af2")

edit: This is in reference to names being whatever you want. Come on, GitHub.

ibm2431 (changes requested):

sql/npc_list.sql
TeoTwawki: > ships

coz moving zone parts.

also certain bcnm entrance (divine might..) and door etc, expect the name to match wtf they reference.

those are usually (nut not always named the hex value of whatever Nth position they exist in zone, prefixed with a symbol like: `_07f` or matching the fourcc of their accompanying animation loop for objects that just sit and sparkle but can't be clicked on (holiday decorations will animate automatically if correctly named).

/endClientTrivia

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
TeoTwawki: even if it does, never trust the event - injections a thing

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
kaincenteno: Fixed. Will be pushing change soon.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
kaincenteno: fixed. Will push change soon.

scripts/zones/Navukgo_Execution_Chamber/mobs/Immortal_Flan.lua
kaincenteno: Deleted. Pushing change soon.

sql/npc_list.sql
kaincenteno: changing it back to 15, until there's a confirmation that this blank npc is not used for something else (can't be proven as most of Aht Ughan quests/misisions are not scripted).

Change will be seen on the next push (in a few minutes)

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: added spaces to commas through the file (most of the file was previously scripted by someone else)
Push coming soon

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: oh crap that's right. Thanks for mentioning it. Reverting this.
Push coming soon

kaincenteno: moved it to begginings, i knew somewhere else needed to be fixed lol

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: @ibm2431  The event handles the check if they dont have enough gil.
No changes done here

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: No need to check the player's Gil as the check is part of the params for it.

No changes made here.

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
kaincenteno: Flans are really far away from the players eyes, They cannot be seen from @Wiggo32 's capture. If the logic on this needs to be changed a capture with 4+ people would be needed to verify how do they pop. 

Let me know if any change is needed here :cat:

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: I'm unsure if it offers divination afterwards. No capture was done for this part.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: created localvar WaoudNeedToZone.

Pushing changes soon

sql/npc_list.sql
TeoTwawki: I'm positive that it doesn't have any missions involved nor any non blu quest.

scripts/globals/bcnm.lua
kaincenteno: resolved

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -13,33 +13,35 @@ require("scripts/globals/npc_util")
...
-function onTrigger(player,npc)
-    local anEmptyVessel = player:getQuestStatus(AHT_URHGAN,tpz.quest.id.ahtUrhgan.AN_EMPTY_VESSEL)
+function onTrigger(player, npc)
+    local anEmptyVessel = player:getQuestStatus(AHT_URHGAN, tpz.quest.id.ahtUrhgan.AN_EMPTY_VESSEL)
     local anEmptyVesselProgress = player:getCharVar("AnEmptyVesselProgress")
     local divinationReady = vanaDay() > player:getCharVar("LastDivinationDay")
-    local beginnings = player:getQuestStatus(AHT_URHGAN,tpz.quest.id.ahtUrhgan.BEGINNINGS)
+    local beginnings = player:getQuestStatus(AHT_URHGAN, tpz.quest.id.ahtUrhgan.BEGINNINGS)
+    local omens = player:getQuestStatus(AHT_URHGAN, tpz.quest.id.ahtUrhgan.OMENS)
+    local needToZone = player:getLocalVar("WaoudNeedToZone")
```
KnowOne134: there is a function player:needToZone()

kaincenteno: ah thanks for the catch @KnowOne134 , let me murder it with kindness

kaincenteno: @KnowOne134 i dont see it on my end. You might be looking at a wrong commit?

TeoTwawki: just generally you don't want a var (in this case `local needToZone`) that has same name as a function. I recall ibm wanting to phase said function out.

kaincenteno: alright. Changing it @TeoTwawki  😸

kaincenteno: changed to waoudNeedToZone :smile_cat: 

resolving comment

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -2,17 +2,53 @@
...
 end
 
-function onTrigger(player,npc)
-    player:startEvent(770)
+function onTrigger(player, npc)
+    local omensProgress = player:getCharVar("OmensProgress")
+    local omens = player:getQuestStatus(AHT_URHGAN, tpz.quest.id.ahtUrhgan.OMENS)
+
+    -- OMENS
+    if player:hasKeyItem(tpz.ki.SEALED_IMMORTAL_ENVELOPE) then
```
ibm2431: I might prefer a:
```lua
if omens >= QUEST_ACCEPTED then
    if omensProgress == 3 then
        ...etc
    elseif omens == QUEST_COMPLETED then
        -- default dialogue
    end
end
```

instead of checking for quest progress by seeing if a character has a key item or not. In my view, key items are just shiny baubles that are given to you as a side effect of your actual progress (the variable) - not the progress themselves.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
ibm2431: Packet injection is a thing 😉 

When params are sent to the client in a `startEvent`, those params only tell a client how to change the display of the event. The resulting "option" that's sent by the client to the server is entirely under the client's control, and there's no way to prevent / guarantee a client can _only_ send certain options back to the server after being given specific params by us.

So instead, we need to be sure to _never trust the client_ and always re-verify that "meaningful" options they send (ie: resulting in giving items, or in this case, deleting gil) are actually valid and that we're not being lied to.

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
ibm2431: While waiting for other players to enter a BCNM, players have a tendency to run up to the edge of the mobs' aggro range.

Going to tag @KnowOne134 and @cocosolos here as they do more work with BCNMs than me - how would we go about setting the number of flans based on registrant party size? Set a "local-but-not-lua-local" var for number of flans needed during `onBattlefieldRegister`? Can we spawn all the required flans at once in `onBattlefieldInitialise`?

ibm2431 (changes requested):

scripts/zones/Navukgo_Execution_Chamber/bcnms/omens.lua
cocosolos: If you wanna use party/alliance size you can use onBattlefieldInitialise and get the party size with ```battlefield:getInitiator():getPartySize()``` or ```battlefield:getInitiator():getAllianceSize()```. Players do get locked out once one of the mobs is engaged so onBattlefieldEnter works as well, depends on how you want the mobs to spawn. For example if someone does an early pull before everyone enters, should they still be fighting the full amount of mobs?

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: ah got it, but in this case it's deleting gil so I would assume if someone is trying to cheat then they must be really dumb to cheat to delete their items lol.

If we are trying to avoid cheating then it would be for other stuff like unlocking this job without meeting the criteria.

I can go and change this and make all the onEventFinish to have the same requirements that started that event :cat: let me know.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
kaincenteno: looks nicer :+1:  thank you @ibm2431

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: after thinking it, you are right @ibm2431 :cat:  it makes sense to not be fooled by those cheaters that like to mess the game experience.

I only included it inside the OMENS onEventFinish. But did it fully on Lathuya. Maybe it could be a good project to be placed on the projects section :sunglasses:

ibm2431 (approved):
Pushed a commit to make the flan spawning behavior accurate to retail (and some miscellaneous fixes and changes).

Approved now, sorry it took me so long to get around to this, kain!

I feel weird merging my own code without a checkmark from _someone_, so I'm holding off on merge until someone-other-than-me says it looks good (or next staff meeting, whichever comes first).

PR #364, opened 2020-02-20
Trust Basics

zach2good:
**_I affirm:_**

This PR topaz-ifys @Omnione's PR (with their blessing) and allows trusts to be cast (but they don't do anything).

With contributions also from @Safhaven's Trust PR.

**GOAL:**
- Take the existing work of others and get it into topaz, so it isn't as daunting to start working on trusts.
- Put in basic plumbing, limit checking etc.

**DONE:**
`trustutils`, `CTrustEntity` and some misc trust changes in core
Bindings: `getTrustID` and `getPartyTrusts`, make `getPartyCount` more usable
Audit the mobpools and mob_spell_lists of all trusts, basic nation trusts have correct spell lists
`isValidHealTarget` helper in `magic.lua` to simplify checks going forward
`spell/trust` entries for a bunch of trusts, which are pretty much empty husks for now
`trust.lua` global with `tpz.trust.canCast` which handles:
- Rejecting casting if you're in an alliance
- Rejecting casting if you're not the party leader
- Rejecting you if you try to cast the same trust twice
- Rejecting you if you try to cast related trusts (See all the Shantotto spell/trust entries).
- Rejecting you if you push the party size above 6 (tested with multiple PCs, trusts and combinations thereof)
- Rejecting you if you try to cast more trusts than you have the ROV KI's for (easily removable)

Bindings for trust Spawn/Despawn/Die messages
Commands:  /returnfaith, /refa, /returntrust and /retr (+all) work 
Fix client crash on casting spells directly on trusts. (Seems hacky, but it's stable at least!)
Trusts despawn and die correctly, when you kill them, or when you die/zone
Trusts will wait for you to engage and start an enmity list before they run in and melee your target
Basic spellcasting if a trust has spells assigned in `mob_spell_lists.sql`

**Starter trust missions are being worked on in a branch off of this branch viewable here:**
https://github.com/zach2good/topaz/compare/trust_list...zach2good:trust_starter_quests

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5340,6 +5340,29 @@ inline int32 CLuaBaseEntity::unlockJob(lua_State *L)
 }
 
 /************************************************************************
+*  Function: hasJob()
+*  Purpose : Check to see if JOBTYPE is unlocked
+*  Example : player:hasJob(BRD)
+*  Notes   :
+************************************************************************/
+
+inline int32 CLuaBaseEntity::hasJob(lua_State *L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+
+    JOBTYPE JobID = (JOBTYPE)lua_tointeger(L, 1);
+
+    TPZ_DEBUG_BREAK_IF(JobID > MAX_JOBTYPE || JobID < 0);
+
+    CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
+
+    lua_pushinteger(L, (PChar->jobs.unlocked >> JobID) & 1);
+    return 1;
+}
+
+/************************************************************************
 *  Function: getMainLvl()
 *  Purpose : Returns the main level of entity's current job
 *  Example : player:getMainLvl()
@@ -5368,6 +5391,29 @@ inline int32 CLuaBaseEntity::getSubLvl(lua_State *L)
     TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype == TYPE_NPC);
 
     lua_pushinteger(L, ((CBattleEntity*)m_PBaseEntity)->GetSLevel());
+    return 1;
+}
+
+/************************************************************************
+*  Function: getJobLevel()
+*  Purpose : Return the levle of job specified by JOBTYPE
+*  Example : player:getJobLevel(BRD)
+*  Notes   :
+************************************************************************/
+
+inline int32 CLuaBaseEntity::getJobLevel(lua_State *L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+
+    JOBTYPE JobID = (JOBTYPE)lua_tointeger(L, 1);
+
+    TPZ_DEBUG_BREAK_IF(JobID > MAX_JOBTYPE || JobID < 0);
+
+    CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
+    lua_pushinteger(L, PChar->jobs.job[JobID]);
+
     return 1;
 }
 
@@ -14202,9 +14248,11 @@ Lunar<CLuaBaseEntity>::Register_t CLuaBaseEntity::methods[] =
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,changeJob),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,changesJob),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,unlockJob),
+    LUNAR_DECLARE_METHOD(CLuaBaseEntity,hasJob),
 
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,getMainLvl),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,getSubLvl),
+    LUNAR_DECLARE_METHOD(CLuaBaseEntity,getJobLevel),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,setLevel),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,setsLevel),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,levelCap),
--- a/src/map/lua/lua_baseentity.h
+++ b/src/map/lua/lua_baseentity.h
@@ -268,9 +268,11 @@ public:
     int32 changeJob(lua_State*);            // changes the job of a char (testing only!)
     int32 changesJob(lua_State*);           // changes the sub job of a char (testing only!)
     int32 unlockJob(lua_State*);            // Unlocks a job for the entity, sets job level to 1
+    int32 hasJob(lua_State*);               // Check to see if JOBTYPE is unlocked for a character
 
     int32 getMainLvl(lua_State*);           // Gets Entity Main Job Level
     int32 getSubLvl(lua_State*);            // Get Entity Sub Job Level
+    int32 getJobLevel(lua_State*);          // Gets character job level for specified JOBTYPE
     int32 setLevel(lua_State*);             // sets the character's level
     int32 setsLevel(lua_State*);            // sets the character's level
     int32 levelCap(lua_State*);             // genkai


Omnione (approved):

ibm2431:
Merging master into a feature branch might pose problems when it comes time to merge the finished feature into stable - all the unrelated commits in master would becoming too~ 😅

ibm2431:
Commits from Omnione were originally based off of sidestream, changing Shantotto's file here threw merge conflicts with our master, one thing led to another, and I felt a straight up rebase off of our current master was in order.

I know how it looks, but I didn't _want_ to attach my name to those commits, I swear!

src/map/entities/battleentity.cpp
```diff
@@ -1180,7 +1180,7 @@ void CBattleEntity::Spawn()
 
 void CBattleEntity::Die()
 {
-    if (CBaseEntity* PKiller = GetEntity(m_OwnerID.targid))
+    if (CBaseEntity * PKiller = GetEntity(m_OwnerID.targid))
```
TeoTwawki: Spaces on BOTH sides?!

We got us a heretic here boys..

scripts/globals/spells/trust/ajido-marujido.lua
```diff
@@ -0,0 +1,19 @@
...
+-- Trust: Ajido-Marujido
+-----------------------------------------
+require("scripts/globals/msg")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    if (not caster:canUseMisc(tpz.zoneMisc.PET)) then
+        return tpz.msg.basic.CANT_BE_USED_IN_AREA
+    elseif (not caster:canCastTrust(spell:getID())) then
+        return
```
TeoTwawki: This is returning nil here, and seems like it could be another misc flag added to the enum instead while leaving the duplicate check to the core before we trigger the casting check (its a different message anyway)

sql/mob_pools.sql
```diff
@@ -5953,130 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
--- INSERT INTO `mob_pools` VALUES (5991,'darrcuiln',
--- INSERT INTO `mob_pools` VALUES (5992,'aahm',
--- INSERT INTO `mob_pools` VALUES (5993,'aaev',
--- INSERT INTO `mob_pools` VALUES (5994,'aamr',
--- INSERT INTO `mob_pools` VALUES (5995,'aatt',
--- INSERT INTO `mob_pools` VALUES (5996,'aagk',
--- INSERT INTO `mob_pools` VALUES (5997,'iroha',
--- INSERT INTO `mob_pools` VALUES (5998,'ygnas',
+
+INSERT INTO `mob_pools` VALUES (5896,'Shantotto','shantotto',153,0x0000B80B00000000000000000000000000000000,4,5,2,240,100,0,2,0,0,0,0,32,0,3,0,0,308,1,0,750);
```
TeoTwawki: It occurs to me we may have the capitalization backwards. IIRC, the 1st field is going to be the script name, and the 2nd is the packet name for instanced objects. So the 2nd would have “proper” name capitalization but no spaces per SE specifications while the first would go full lowercase to avoid posix problems that were run into ages ago (duplicates happened, and git on windows machines got confused)

src/map/lua/lua_baseentity.cpp
```diff
@@ -5339,6 +5340,29 @@ inline int32 CLuaBaseEntity::unlockJob(lua_State *L)
...
 }
 
+/************************************************************************
+*  Function: hasJob()
+*  Purpose : Check to see if JOBTYPE is unlocked
+*  Example : player:hasJob(BRD)
+*  Notes   :
+************************************************************************/
+
+inline int32 CLuaBaseEntity::hasJob(lua_State *L)
```
TeoTwawki: Looks like that rebase didn’t go perfectly as planned

scripts/globals/spells/trust/ajido-marujido.lua
zach2good: I don't know what the desired message is, but is `return tpz.msg.basic.MAGIC_CANNOT_CAST` a better placeholder?

src/map/lua/lua_baseentity.cpp
zach2good: 💣

src/map/entities/battleentity.cpp
zach2good: image

scripts/globals/spells/cure.lua
```diff
@@ -62,7 +62,7 @@ function onSpellCast(caster,target,spell)
         end
     end
 
-    if (target:getAllegiance() == caster:getAllegiance() and (target:getObjType() == tpz.objType.PC or target:getObjType() == tpz.objType.MOB)) then
+    if (target:getAllegiance() == caster:getAllegiance() and (target:getObjType() == tpz.objType.PC or target:getObjType() == tpz.objType.MOB  or target:getObjType() == tpz.objType.TRUST)) then
```
ibm2431: I'm not going to require a solution this particular problem in this PR, but I'll probably want a better way a spell can check for valid entity types, prior to all spells being modified to add trusts to them. It's already a long `or` chain, and if I recall, Fellow branch added its own as well.

Bitflags and an SQL field, masks, I dunno, _something_...!

To quote my past self:
> we'll eventually need a validHealTarget which cuts:
(target:getObjType() == dsp.objType.PC or target:getObjType() == dsp.objType.MOB or target:getObjType() == dsp.objType.TRUST)
Out of every healing script, and also makes sure the trust is a valid target (some trusts can't be targeted with spells).

scripts/globals/spells/trust/aldo.lua
```diff
@@ -0,0 +1,19 @@
+-----------------------------------------
```
ibm2431, line 1: (The following comments apply to all trust spell scripts):

scripts/globals/spells/trust/aldo.lua
```diff
@@ -0,0 +1,19 @@
...
+-- Trust: Aldo
+-----------------------------------------
+require("scripts/globals/msg")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
```
ibm2431, line 7: We can probably save ourselves some effort later by calling a globalized function from the trusts global:
```lua
function onMagicCastingCheck(caster, target, spell)
    local message = tpz.trust.castCheck(caster, spell)
    return message
end
```

This example has a local assignment and return because after the "basic" check, you can still check for "duplicate/distinct" trusts while in the spell script, if the "can cast" logic is taken out of core (see later).

scripts/globals/spells/trust/aldo.lua
```diff
@@ -0,0 +1,19 @@
...
+    if (not caster:canUseMisc(tpz.zoneMisc.PET)) then
```
ibm2431: Trusts are invalid in some areas where pets are (ex: Salvage), so Trusts will need their own misc flags.

scripts/globals/spells/trust/aldo.lua
```diff
@@ -0,0 +1,19 @@
...
+        return tpz.msg.basic.CANT_BE_USED_IN_AREA
```
ibm2431: The message is different, but I don't expect the message IDs to be added in this PR, so just noting it now~

(One benefit of making the check logic a global is that changing the message later should be a one-line edit so someone doesn't need to go through all the scripts.)

scripts/globals/spells/trust/aldo.lua
```diff
@@ -0,0 +1,19 @@
...
+    elseif (not caster:canCastTrust(spell:getID())) then
```
ibm2431: At the moment this is a call to core, but I think we'd be better served if we instead do "can cast" in lua. Core might need a few binds to check trust ID (if they aren't already there), but I believe all other checks can be done without involving core:
```lua
-- pseudo
tpz.trust.canCast = function(caster, spell)
    if not caster:canUseMisc(tpz.zoneMisc.TRUST) then
        return NO_TRUSTS_HERE_MSG
    end

    if not party leader then
        return NOT_LEADER_MSG
    end

    local num_pt, num_trusts
    for players in party do
        if member is trust then
            if member spell id is spell:getID() then
                return DUPLICATE_TRUST_MSG
            else
                num_trusts = num_trusts + 1
            end
        end
        num_pt = num_pt + 1
    end
    if num_pt >= 6 then
        return PT_FULL_MSG
    end
    if num_trusts >= 3 and not player has ROV KI then
        return TOO_MANY_TRUSTS_MSG
    end
end
```

We can also either table "trusts with the same name", or pass the "the other same-trust"'s ID into this call to check for it:
```lua
-- shantotto ii.lua
function onMagicCastingCheck(caster, target, spell)
    tpz.trust.canCast(caster, spell, shantotto_i_spell_id)
end

-- trust global
trust.canCast = function(caster, spell, duplicate_trust_id)
    if (member spell ID is spell:getID) or (member spell ID is duplicate_trust_id) then
        return DUPLICATE_TRUST_MSG
    end
end
```
This gets around core doing the "heuristic" checks based on name; just let lua handle it in the spell script itself~

scripts/globals/spells/trust/aldo.lua
```diff
@@ -0,0 +1,19 @@
...
+        return
+    end
+    return 0
+end
+
+function onSpellCast(caster, target, spell)
+    caster:spawnTrust(930)
+    return 0
```
ibm2431: We _could_ globalize this as well, into a `tpz.trust.spawn(caster, spell)`, which at the moment would just be:
```lua
    caster:spawnTrust(spell:getID())
```
But at least it'd be in a global place in case we need to make additions to onSpellCast behavior later. But I'll be honest and admit I'm currently drawing a blank on "global useful" additions that'd necessitate calling a trust global method from the spell script, instead of calling `caster:spawnTrust(spell:getID())` directly from the script.

Maybe buffing the trust if certain conditions are met? Global checks for current campaigns?

Things like "saying special messages based on other trusts in party", "advancing certain quests", and "attach certain AI" would probably need to be done at the individual script level though, so I'm uncertain how well a lua trust global for "onSpellCast" would work out.

scripts/globals/trust.lua
```diff
@@ -0,0 +1,22 @@
...
+
+tpz.trust.messageParty = function(mob, messageString)
+    local party = {}
+    local master = mob:getMaster()
+
+    party = master:getAlliance()
+
+    for _, member in ipairs(party) do
+        if member:getZoneID() == mob:getZoneID() and member:isPC() then
+            member:PrintToPlayer(messageString, 4)
```
ibm2431: Did we ever figure out how these messages work on retail?

sql/mob_spell_lists.sql
```diff
@@ -2515,6 +2515,62 @@ INSERT INTO `mob_spell_lists` VALUES ('Ghul-I-Beaban_BLM',305,174,1,255); -- fir
 INSERT INTO `mob_spell_lists` VALUES ('Ghul-I-Beaban_BLM',305,179,1,255); -- blizzaga
 INSERT INTO `mob_spell_lists` VALUES ('Saa_Doyi_the_Fervid',306,320,1,255); -- katon_ichi
 INSERT INTO `mob_spell_lists` VALUES ('Frost_Flambeau',307,180,1,255); -- blizzaga_ii
+-- Trust
+INSERT INTO `mob_spell_lists` VALUES ('TRUST_Shantotto',308,159,1,255);  -- Stone
+INSERT INTO `mob_spell_lists` VALUES ('TRUST_Shantotto',308,160,26,255); -- Stone II
+INSERT INTO `mob_spell_lists` VALUES ('TRUST_Shantotto',308,161,51,255); -- Stone III
+INSERT INTO `mob_spell_lists` VALUES ('TRUST_Shantotto',308,162,68,255); -- Stone IV
```
ibm2431: The self-professed master of magic should probably know how to cast Thunder 😉 

Overly picky ibm here: Sorted by spell ID please~! (Fire, Blizzard, Aero, Stone, Thunder, Water)

sql/mob_pools.sql
```diff
@@ -5953,130 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
+INSERT INTO `mob_pools` VALUES (5899,'Excenmille','excenmille',145,0x0000BB0B00000000000000000000000000000000,7,0,8,390,100,0,0,0,0,0,0,32,0,3,0,0,310,1,0,4);
+INSERT INTO `mob_pools` VALUES (5900,'Ayame','ayame',149,0x0000BC0B00000000000000000000000000000000,12,0,10,440,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,5);
+INSERT INTO `mob_pools` VALUES (5901,'NanaaMihgo','nanaa_mihgo',152,0x0000BD0B00000000000000000000000000000000,6,0,2,190,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,6);
+INSERT INTO `mob_pools` VALUES (5902,'Curilla','curilla',145,0x0000BE0B00000000000000000000000000000000,7,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,7,1,0,7);
+INSERT INTO `mob_pools` VALUES (5903,'Volker','volker',149,0x0000BF0B00000000000000000000000000000000,1,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,8,1,0,8);
+INSERT INTO `mob_pools` VALUES (5904,'Ajido-Marujido','ajido-marujido',153,0x0000C00B00000000000000000000000000000000,4,5,1,400,100,0,0,0,0,0,0,32,0,3,0,1,9,1,0,0);
+INSERT INTO `mob_pools` VALUES (5905,'Trion','trion',145,0x0000C10B00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,10,1,0,10);
```
ibm2431: Might be best to zero out the IDs for non-existing spell lists for now~

sql/mob_pools.sql
```diff
@@ -5953,130 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
--- INSERT INTO `mob_pools` VALUES (5994,'aamr',
--- INSERT INTO `mob_pools` VALUES (5995,'aatt',
--- INSERT INTO `mob_pools` VALUES (5996,'aagk',
--- INSERT INTO `mob_pools` VALUES (5997,'iroha',
--- INSERT INTO `mob_pools` VALUES (5998,'ygnas',
+
+INSERT INTO `mob_pools` VALUES (5896,'Shantotto','shantotto',153,0x0000B80B00000000000000000000000000000000,4,5,2,240,100,0,2,0,0,0,0,32,0,3,0,0,308,1,0,750);
+INSERT INTO `mob_pools` VALUES (5897,'Naji','naji',149,0x0000B90B00000000000000000000000000000000,4,5,2,240,100,0,2,0,0,0,0,32,0,3,0,0,0,1,0,750);
+INSERT INTO `mob_pools` VALUES (5898,'Kupipi','kupipi',153,0x0000BA0B00000000000000000000000000000000,3,0,11,230,100,2,0,0,0,0,0,32,0,3,0,1,309,1,0,3);
+INSERT INTO `mob_pools` VALUES (5899,'Excenmille','excenmille',145,0x0000BB0B00000000000000000000000000000000,7,0,8,390,100,0,0,0,0,0,0,32,0,3,0,0,310,1,0,4);
```
ibm2431: Are these skill list IDs intentional?

sql/mob_pools.sql
```diff
@@ -5953,130 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
+INSERT INTO `mob_pools` VALUES (6013,'Lilisette','lilisette_ii',145,0x00000C0C00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,10,1,0,10);
+INSERT INTO `mob_pools` VALUES (6014,'Tenzen','tenzen_ii',145,0x0000190C00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,10,1,0,10);
+INSERT INTO `mob_pools` VALUES (6015,'Mumor','mumor_ii',145,0x0000200C00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,10,1,0,10);
+INSERT INTO `mob_pools` VALUES (6016,'Ingrid','ingrid_ii',145,0x00001E0C00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,10,1,0,10);
+INSERT INTO `mob_pools` VALUES (6017,'Arciela','arciela_ii',145,0x00000D0C00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,10,1,0,10);
+INSERT INTO `mob_pools` VALUES (6018,'Iroha','iroha_ii',145,0x0000280C00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,10,1,0,10);
+INSERT INTO `mob_pools` VALUES (6019,'Shantotto','shantotto_ii',153,0x0000260C00000000000000000000000000000000,4,5,2,240,100,0,2,0,0,0,0,32,0,3,0,0,276,1,0,750);
```
ibm2431: Shantotto and Shantotto II still have their previously defined spell and skill list IDs; Shantotto looks the same, but it seems Shantotto II's spell list of 276 got taken.

Both of their skill list IDs were taken as well.

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -69,14 +69,13 @@ void CTrustController::Tick(time_point tick)
 
 void CTrustController::DoCombatTick(time_point tick)
 {
-    if (!POwner->PMaster->PAI->IsEngaged())
+    if ((POwner->PMaster == nullptr || POwner->PMaster->isMounted() ||
```
ibm2431: I'm _fairly_ certain that Trusts don't die if the owner mounts; they'll keep up with you.

src/map/ai/states/magic_state.cpp
```diff
@@ -96,7 +96,7 @@ bool CMagicState::Update(time_point tick)
 
         action_t action;
 
-        if (!PTarget || m_errorMsg || (HasMoved() && (m_PEntity->objtype != TYPE_PET ||
+        if (!PTarget || m_errorMsg || (HasMoved() && (m_PEntity->objtype != TYPE_PET || m_PEntity->objtype != TYPE_TRUST ||
```
ibm2431: I realize now that I've never seen a trust have spellcasting interruptions other than being silenced. I'm not saying 100% it doesn't happen, just that I haven't seen it (because usually trusts don't move while they're casting). It may be hard to confirm though, as I think the only thing that'd cause a mid-spell trust to move would be knockback, and then the "at same location" check might be coming into play.

src/map/entities/trustentity.cpp
```diff
@@ -94,11 +84,10 @@ void CTrustEntity::Die()
 {
     PAI->ClearStateStack();
     PAI->Internal_Die(0s);
-    luautils::OnMobDeath(this, nullptr);
```
ibm2431, line 97: We will need a way to hook into trust spawn and despawn events, if only for those special party messages. Listeners in the controller, maybe?

src/map/lua/lua_baseentity.cpp
```diff
@@ -11887,6 +11888,100 @@ inline int32 CLuaBaseEntity::spawnPet(lua_State *L)
     return 0;
 }
 
+/************************************************************************
+*  Function: canCastTrust()
```
ibm2431: See above comments regarding if we can potentially implement this in lua; it'd be preferable if so~

src/map/packets/trust_sync.cpp
```diff
@@ -36,13 +37,28 @@ CTrustSyncPacket::CTrustSyncPacket(CCharEntity* PChar, CTrustEntity* PTrust)
...
+    {
+        ref<uint8>(0x04) = 0x03;
+        ref<uint8>(0x05) = 0x05;
+    }
 
     ref<uint16>(0x06) = PTrust->targid;
     ref<uint32>(0x08) = PTrust->id;
-    ref<uint16>(0x0C) = PChar->targid;
+
+    if (PTrust)
```
ibm2431: It may be late, but I'm a little confused here: shouldn't a check like this be done higher in the packet? We've already accessed targid and id at this point~

And are we sending out this packet when we _don't_ have a PTrust?

What's going on with the size change?

src/map/zone.h
```diff
@@ -555,6 +555,7 @@ class CZone
     virtual void    SpawnMOBs(CCharEntity* PChar);                                  // отображаем MOBs в зоне
     virtual void    SpawnPETs(CCharEntity* PChar);                                  // отображаем PETs в зоне
     virtual void    SpawnNPCs(CCharEntity* PChar);                                  // отображаем NPCs в зоне
+    virtual void    SpawnTRUSTs(CCharEntity* PChar);                                // отображаем TRUSTs в зоне
```
ibm2431: I like how even the new functions are getting russian comments now

src/map/utils/petutils.cpp
```diff
@@ -244,6 +316,134 @@ namespace petutils
...
+                mob_pools.name,\
+                modelid,\
+                mobsize,\
+                systemid,\
+                mob_pools.familyid,\
+                mob_pools.mJob,\
+                mob_pools.sJob,\
+                (mob_family_system.HP / 100), \
+                (mob_family_system.MP / 100), \
+                mob_family_system.speed, \
```
ibm2431: Since this query is pulling from multiple tables, could we make sure all columns are prepended with the source table it's coming from? And can we get the selections grouped by their source table? We've got columns from pools, then some from family, then more from pools.

src/map/utils/petutils.cpp
```diff
@@ -1846,4 +2006,32 @@ namespace petutils
         }
         return false;
     }
+
+    bool CheckTrustName(const char* name, uint32 trustID)
```
ibm2431: I don't think we'll need this function if we have each spell script check for it's known "duplicate" spell IDs

src/map/utils/petutils.cpp
```diff
@@ -244,6 +316,134 @@ namespace petutils
...
+            }
+        }
+
+        for (uint32 index = 0; index < g_PTrustIDList.size(); index++)
+        {
+            QueryTrust(g_PTrustIDList.at(index)->spellID);
+        }
+    }
+
+    void QueryTrust(uint32 TrustID)
```
ibm2431: Maybe rename this function so it's more clear that it's not merely querying for a trust, but actually building one~

src/map/ai/helpers/targetfind.cpp
```diff
@@ -241,12 +252,21 @@ void CTargetFind::addAllInAlliance(CBattleEntity* PTarget, bool withPet)
 
 void CTargetFind::addAllInParty(CBattleEntity* PTarget, bool withPet)
 {
-
```
ibm2431, line 244: GitHub won't let me add a review comment to unchanged lines, but will trusts need to be added to `addAllInZone` too?

src/map/ai/helpers/targetfind.cpp
```diff
@@ -475,6 +498,19 @@ CBattleEntity* CTargetFind::getValidTarget(uint16 actionTargetID, uint16 validTa
...
+    if (PTarget->objtype == TYPE_TRUST)
+    {
+        return PTarget;
+    }
+
+    if (validTargetFlags & TARGET_PLAYER_PARTY)
+    {
+        if (PTarget->objtype == TYPE_TRUST)
+        {
+            return PTarget;
```
ibm2431: You've already returned PTarget if it's a Trust up on 503; the check for validTargetFlags if the target is a trust never gets reached.

src/map/packets/party_define.cpp
```diff
@@ -63,18 +62,62 @@ CPartyDefinePacket::CPartyDefinePacket(CParty* PParty)
...
+        uint32 allianceid = 0;
+        if (PParty->m_PAlliance)
+        {
+            allianceid = PParty->m_PAlliance->m_AllianceID;
+        }
 
-        if (PParty->GetLeader() != nullptr && PParty->GetLeader()->objtype == TYPE_PC)
+        uint8 i = 0;
+
+        int ret = Sql_Query(SqlHandle, msg, allianceid, PParty->GetPartyID(), PARTY_SECOND | PARTY_THIRD);
```
ibm2431, line 83: So here's where I say that trusts can't be summoned while in an alliance.

_But_, if it's possible for core to handle trusts in alliances without creating additional overhead or other issues, I would personally be willing to overlook trust compatibility with alliances so long as any `canCast` function rejects the attempt by default~

src/map/zone_entities.cpp
```diff
@@ -962,7 +1079,7 @@ void CZoneEntities::ZoneServer(time_point tick, bool check_regions)
                 CMobEntity* PCurrentMob = (CMobEntity*)PMobIt.second;
                 PCurrentMob->PEnmityContainer->Clear(PPet->id);
             }
-            if (PPet->getPetType() != PETTYPE_AUTOMATON || !PPet->PMaster)
+            if (!PPet->PMaster || PPet->objtype == TYPE_TRUST || PPet->getPetType() != PETTYPE_AUTOMATON)
```
ibm2431: Will a trust be in this pet list if they're being put in the trust list down below?

src/map/zone_entities.cpp
```diff
@@ -973,6 +1090,46 @@ void CZoneEntities::ZoneServer(time_point tick, bool check_regions)
...
+            {
+                CCharEntity* PChar = (CCharEntity*)it->second;
+                if (distance(PChar->loc.p, PTrust->loc.p) < 50)
+                {
+                    PChar->SpawnTRUSTList.erase(PTrust->id);
+                    PChar->ReloadPartyInc();
+                }
+            }
+
+            if (!PTrust->PMaster || PTrust->objtype == TYPE_TRUST)
```
ibm2431: At this point, can a PTrust ever have an objtype _other_ than `TYPE_TRUST`?

src/map/zone_entities.cpp
```diff
@@ -316,6 +369,26 @@ void CZoneEntities::DecreaseZoneCounter(CCharEntity* PChar)
             }
             PChar->PPet = nullptr;
         }
+
+        //remove trust
+        if (PChar->PTrusts.size() > 0)
+        {
+            for (auto& trust : PChar->PTrusts)
+            {
+                trust->PAI->Disengage();
```
ibm2431: I'm not as familiar with AI controllers as I'd _like_, but [looking at `ClearTrusts`, it's already calling Despawn()](https://github.com/project-topaz/topaz/blob/release/src/map/entities/charentity.cpp#L490-L499) on the list.

ibm2431 (changes requested):
Thanks as always, Zach!

(And Omnione!)

I freely admit that mobs aren't my usual wheelhouse. Someone else might have more to say about how their entities are handled.

src/map/utils/trustutils.cpp
```diff
@@ -186,336 +166,345 @@ namespace trustutils
...
-        case 4: race = 1; break;    //Elvaan
-        case 5:
-        case 6: race = 2; break;    //Tarutaru
-        case 7: race = 3; break;    //Mithra
-        case 8: race = 4; break;    //Galka
-        }
+    uint8 race = 0; //Human
 
-        // Расчет прироста HP от main job
+    switch (PTrust->look.race)
```
TeoTwawki: this needs to die, trusts aren't players. there is a trust that is a treant even which would fail here.

TeoTwawki: as an example, a "taru" trust should simply use family ID 153 or 154 (or make a new one if those stats don't match)

src/map/utils/trustutils.cpp
```diff
@@ -239,197 +240,115 @@ void FreeTrustList()
...
+    uint16 mSTR = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 2), mLvl);
+    uint16 mDEX = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 3), mLvl);
+    uint16 mVIT = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 4), mLvl);
+    uint16 mAGI = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 5), mLvl);
+    uint16 mINT = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 6), mLvl);
+    uint16 mMND = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 7), mLvl);
+    uint16 mCHR = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 8), mLvl);
+
+    uint16 sSTR = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetSJob(), 2), PTrust->GetSLevel());
+    uint16 sDEX = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetSJob(), 3), PTrust->GetSLevel());
```
TeoTwawki: shouldn't you just make `PTrust->GetSLevel()` a variable like you did for mainjob's level to avoid all these repeat calls?

TeoTwawki (commented):
Um.. we should not be using the player chat spoofs for retail messages. These already exist in the client and just need called correctly.

zach2good:
@TeoTwawki I had a look in `adventuringfellow` and I see how it's _meant_ to be done, I'll whip all of the fake PT chat out and leave it for my next PR 👍

TeoTwawki:
@zach2good just a heads up when you get there, he used showText and it should be messageSpecial - showText is just wrapping messageSpecial to add a few additional things (things which neither fellows nor trusts will need)

scripts/commands/addalltrusts.lua
```diff
@@ -21,7 +21,7 @@ function onTrigger(player, target)
         921,922,923,924,925,926,927,928,929,930,931,932,933,934,935,936,937,938,939,940,941,942,943,944,945,
         946,947,948,949,950,951,952,953,954,955,956,957,958,959,960,961,962,963,964,965,966,967,968,969,970,
         971,972,973,974,975,976,977,978,979,980,981,982,983,984,985,986,987,988,989,990,991,992,993,994,995,
-        996,997,998,999,1003,1004,1005,1006,1007,1008,1009,1010,1011,1012,1013,1014,1015,1016,1017,1018,1019
+        996,997,998,999,1003,1004,1005,1006,1007,1008,1009,1010,1011,1012,1013,1014,1015,1016,1017,1018,1019,
```
TeoTwawki: The trailing comma has no effect since it is the final element

TeoTwawki (commented):
I’m surprised they even removed Cornelia from the dats! (I just checked)

scripts/globals/mobskills/provoke.lua
```diff
@@ -0,0 +1,15 @@
...
+require("scripts/globals/msg")
+require("scripts/globals/status")
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    target:addEnmity(mob, 1, 1800)
```
ibm2431, line 13: Mainly to everyone, not necessarily you, Zach:
I'm _okay_ with these new scripts for now as a testing bed, but we're going to have to come up with _some_ solution to trusts having all the same abilities that players do! The last thing I want is additional burden to maintain an additional copy of weaponskills.

scripts/globals/spells/trust/aldo_uc.lua
```diff
@@ -0,0 +1,14 @@
+-----------------------------------------
+-- Trust: Aldo UC
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
```
ibm2431, line 7: We might need some additional Unity logic later. I don't know how we'd handle adding/removing Unity trusts. But this is far, far in the future. Just noting it here.

And our favorite head of a criminal organization might need to check for himself.

scripts/globals/spells/trust/arciela_ii.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Balamor
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: Doesn't our favorite princess need to check for copies of herself?

scripts/globals/spells/trust/arciela_ii.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Balamor
```
ibm2431: I'm not certain she would be entirely pleased with this label.

scripts/globals/spells/trust/balamor.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Ayame
```
ibm2431: Balamor would probably be offended as well.

scripts/globals/spells/trust/ayame_uc.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Ayame UC
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: Our second-favorite samurai might need to check for herself too~

scripts/globals/spells/trust/ingrid.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Ingrid
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: Our... "favorite"... exorcist might want to check for herself~!

scripts/globals/spells/trust/iroha.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Iroha
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: As much as I would love to have as many Irohas as possible, our favorite samurai priestess should probably be bound by the same rules.

scripts/globals/spells/trust/lilisette_ii.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Lilisette II
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: Our favorite dancer as well

scripts/globals/spells/trust/lion_ii.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Lion II
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: As well as our favorite pirate princess

scripts/globals/spells/trust/mumor_ii.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Mumor II
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: There can only be one of our favorite pop idol!

scripts/globals/spells/trust/naja_uc.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Naja Salaheem UC
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: The thought of one Naja is horrifying enough!

scripts/globals/spells/trust/nashmeira_ii.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Nashmeira II
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: Could we make an argument that one of our favorite puppeteer's is actually a very life-like puppet being controlled by the other, letting us have both in the party?

scripts/globals/spells/trust/prishe_ii.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Prishe II
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: As much as I would like to double the pain, we can only have one of our favorite brawler

scripts/globals/spells/trust/tenzen_ii.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Tenzen II
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell)
```
ibm2431: And only one of our favorite samurai

scripts/globals/spells/trust/ygnas.lua
```diff
@@ -0,0 +1,13 @@
+-----------------------------------------
+-- Trust: Ygnas
+-----------------------------------------
+require("scripts/globals/trust")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    return tpz.trust.canCast(caster, spell, 1010)
```
ibm2431: Is Zeid too scary for Ygnas to want to be around? 😉

sql/mob_skill_lists.sql
```diff
@@ -3461,3 +3461,4 @@ INSERT INTO `mob_skill_lists` VALUES ('Siren',1010,3512); -- Sonic Buffet
 INSERT INTO `mob_skill_lists` VALUES ('Siren',1010,3513); -- Entice
 INSERT INTO `mob_skill_lists` VALUES ('Siren',1010,3514); -- Hysteric Assault
 INSERT INTO `mob_skill_lists` VALUES ('Siren',1010,3515); -- Clarsach Call
+INSERT INTO `mob_skill_lists` VALUES ('TRUST_Trion',1011,1945); -- Provoke
```
ibm2431, line 3464: I'm thinking that during the "set up abilities" portion of this trust project we may want to consider reserving an entire block for trust skill lists so that they're all together. Not going to worry about it right now though~

sql/mob_skills.sql
```diff
@@ -1765,7 +1765,7 @@ INSERT INTO `mob_skills` VALUES (1941,1305,'string_clipper',0,7.0,2000,1,4,0,0,0
 INSERT INTO `mob_skills` VALUES (1942,1303,'arcuballista',0,15.0,2000,1,4,0,0,0,3,1,0);
 INSERT INTO `mob_skills` VALUES (1943,1306,'slapstick',0,7.0,2000,1,4,0,0,0,8,0,0);
 INSERT INTO `mob_skills` VALUES (1944,1307,'shield_bash',0,7.0,2000,0,4,4,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (1945,1219,'provoke',0,7.0,2000,0,4,4,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (1945,1218,'provoke',0,7.0,2000,0,4,4,0,0,0,0,0);
```
ibm2431, line 1768: And in the future, we might also want to make sure that any skills "added" for trusts are the same IDs as captured from retail (and that the skill list uses the right one). According to the Windower resource database, there's three different copies of Provoke in monster_abilities.

scripts/globals/spells/trust/curilla.lua
```diff
@@ -0,0 +1,44 @@
...
+        if trySentinel == 1 and os.time() >= sentinelReadyTime then
+            trust:useMobAbility(707, target)
+            trust:setLocalVar("trySentinel", 0)
+            trust:setLocalVar("sentinelReadyTime", os.time() + 600)
+        end
+
+        if trust:getTP() >= 1000 then
+            local weaponSkills = {34, 37, 41}
+            local ws = weaponSkills[math.random(1, #weaponSkills)]
+            trust:useMobAbility(ws, target)
```
ibm2431, line 35: You probably recognize this, Zach, so this is mainly for others:

Trust AI will need quite a bit of work to implement some new systems. As examples, cooldowns should be handled by the entity itself, and available weaponskills should be provided by core upon spawn based on skill lists and the trust's level. What's currently here is probably testing to make sure trusts are able to use abilities alright and that listeners work, rather than being a finalized AI.

I'm okay with these limited placeholders being in for merge (so people working on the branch can see a couple trusts in action for testing), but please no one start copy-pasting for all tank trusts and then open a PR~ 😅

scripts/globals/trust.lua
```diff
@@ -1,22 +1,80 @@
...
+tpz.trust.canCast = function(caster, spell, not_allowed_trust_ids)
+
+    -- Trusts not allowed in an alliance
+    if caster:checkSoloPartyAlliance() == 2 then
+        return tpz.msg.basic.TRUST_NO_CAST_TRUST
+    end
+
+    -- Trusts only allowed in certain zones
+    if not caster:canUseMisc(tpz.zoneMisc.TRUST) then
+        -- TODO: Update area flags
```
ibm2431, line 24: This can be a separate PR after this is merged in, but just adding a note here to make sure updating zone flags isn't forgotten~

scripts/globals/trust.lua
```diff
@@ -1,22 +1,80 @@
...
+        --return tpz.msg.basic.TRUST_NO_CALL_AE
+    end
 
-    party = master:getAlliance()
+    -- You can only summon trusts if you are the party leader or solo
```
ibm2431, line 28: There's also the check for if someone has recently joined, but I'm okay with that being put on a TODO list for later. Again, just noting~

scripts/globals/trust.lua
```diff
@@ -1,22 +1,80 @@
...
     end
+
+    -- Max party size
+    if num_pt >= 6 then
+        return tpz.msg.basic.TRUST_NO_CAST_TRUST
+    end
+
+    -- Limits set by ROV Key Items
+    if num_trusts >= 3 and not caster:hasKeyItem(tpz.ki.RHAPSODY_IN_WHITE) then
+        return tpz.msg.basic.TRUST_NO_CAST_TRUST
```
ibm2431, line 72: I believe there's a message for reaching the maximum number of allowed trusts. But as the last two notes: "for later"

src/map/entities/trustentity.cpp
```diff
@@ -48,54 +49,46 @@ void CTrustEntity::PostTick()
...
-        }
 
-        if (updatemask & UPDATE_HP)
+        if (PMaster)
         {
-            if (PMaster->PParty != nullptr)
+            if (updatemask & UPDATE_HP)
             {
-                PMaster->ForParty([this](auto PMember)
+                if (PMaster->PParty != nullptr)
```
ibm2431: Can probably combine these nested ifs with some `&&` 😉

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -87,35 +98,43 @@ void CTrustController::DoCombatTick(time_point tick)
...
+
         POwner->PAI->EventHandler.triggerListener("COMBAT_TICK", POwner, POwner->PMaster, PTarget);
     }
 }
 
 void CTrustController::DoRoamTick(time_point tick)
 {
-    if ((POwner->PMaster == nullptr || POwner->PMaster->isMounted() ||
-        POwner->PMaster->isDead()) && POwner->isAlive())
+    if (POwner->PMaster->PAI->IsEngaged() && GetTopEnmity() == POwner->PMaster)
```
ibm2431, line 121: Checking if the topEnmity is the master might not work if the master isn't the tank, and checking if it's not null would cause trusts to engage before the master has hit the mob. Maybe check to see if the master has it as their claimed mob?

Although, that will still run into issues if the mob was claimed by JA or MA. We might end up needing a "has smacked" 😦

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -127,4 +146,48 @@ bool CTrustController::Ability(uint16 targid, uint16 abilityid)
...
+    };
+
+    position_t final_pos = target_pos;
+    for (auto PTrust : static_cast<CCharEntity*>(POwner->PMaster)->PTrusts)
+    {
+        if (POwner != PTrust && distance(POwner->loc.p, PTrust->loc.p) < 5.0f)
+        {
+            auto angle = getangle(POwner->loc.p, PTrust->loc.p);
+            position_t declumped_pos{ POwner->loc.p.x - (cosf(rotationToRadian(angle)) * 2.0f),
+                POwner->loc.p.y, POwner->loc.p.z + (sinf(rotationToRadian(angle)) * 2.0f), 0, 0 };
```
ibm2431, line 186: Making sure: Does this produce positions that are like retail (all fairly close to where the master was upon engage)?

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -127,4 +146,48 @@ bool CTrustController::Ability(uint16 targid, uint16 abilityid)
...
+        position_t final_pos{
+            lerp(a.x, b.x, ratio),
+            lerp(a.y, b.y, ratio),
+            lerp(a.z, b.z, ratio),
+            0, 0};
+        return final_pos;
+    };
+
+    position_t final_pos = target_pos;
+    for (auto PTrust : static_cast<CCharEntity*>(POwner->PMaster)->PTrusts)
```
ibm2431, line 180: Disclaimer: I'd have to see this in action, but my hot take is that this might have some interesting results with trusts trying to negotiate who stands where all at the same time. 😅

src/map/entities/trustentity.cpp
```diff
@@ -48,54 +49,46 @@ void CTrustEntity::PostTick()
...
 void CTrustEntity::FadeOut()
 {
-    CMobEntity::FadeOut();
+    CBaseEntity::FadeOut();
     loc.zone->PushPacket(this, CHAR_INRANGE, new CEntityUpdatePacket(this, ENTITY_DESPAWN, UPDATE_NONE));
 }
 
 void CTrustEntity::Die()
 {
-    PAI->ClearStateStack();
```
ibm2431, line 95: Do we not need to clear the state stack? Or is it being handled elsewhere?

src/map/party.cpp
```diff
@@ -200,10 +200,18 @@ uint8 CParty::MemberCount(uint16 ZoneID)
...
+        auto* member = members.at(i);
+        if (member->getZone() == ZoneID)
         {
             count++;
         }
+        if (member->objtype == TYPE_PC)
+        {
+            for (auto* trust : static_cast<CCharEntity*>(member)->PTrusts)
+            {
+                count++;
```
ibm2431, line 212: Making sure: Trusts aren't being double-counted, right? I've forgotten if trusts are inserted as actual members of the party object.

src/map/packets/trust_sync.cpp
```diff
@@ -31,34 +31,22 @@ CTrustSyncPacket::CTrustSyncPacket(CCharEntity* PChar, CTrustEntity* PTrust)
...
-        packBitsBE(data + (0x04), (0x18) + PTrust->name.size(), 0, 6, 10); // Message Size
-
-        if(PChar)
-        {
-            ref<uint16>(0x0C) = PChar->targid;
-        }
-
-        memcpy(data + (0x18), PTrust->GetName(), PTrust->name.size());
-    }
+    packBitsBE(data + (0x04), (0x18) + PTrust->name.size(), 0, 6, 10); // Message Size
```
ibm2431: Is name here our script name, or the packet name?

src/map/utils/trustutils.cpp
```diff
@@ -0,0 +1,468 @@
...
+            break;
+        }
+    }
+
+    PTrust->loc              = PMaster->loc;
+    PTrust->m_OwnerID.id     = PMaster->id;
+    PTrust->m_OwnerID.targid = PMaster->targid;
+
+    // spawn me randomly around master
+    PTrust->loc.p = nearPosition(PMaster->loc.p, CTrustController::SpawnDistance, (float)M_PI * tpzrand::GetRandomNumber(2.0f));
```
ibm2431, line 270: I think retail somehow forms a line based off of the master's last position, in that trusts spawn "behind" you, and then behind other trusts. Although it's been a while since I've last watched how they spawn, and this is very clearly "for later", because it's not important.

src/map/utils/trustutils.cpp
```diff
@@ -0,0 +1,468 @@
...
+            if (maxSubSkill != 0)
+            {
+                PTrust->WorkingSkills.skill[i] = maxSubSkill;
+            }
+        }
+    }
+
+    for (int i = SKILL_HAND_TO_HAND; i <= SKILL_STAFF; i++)
+    {
+        uint16 maxSkill = battleutils::GetMaxSkill(3, mLvl > 99 ? 99 : mLvl);
```
ibm2431: `GetMaxSkill(i`?

src/map/utils/trustutils.cpp
```diff
@@ -0,0 +1,468 @@
...
+    Trust_t* trustData   = new Trust_t();
+
+    for (uint32 index = 0; index < g_PTrustList.size(); index++)
+    {
+        if (g_PTrustList.at(index)->trustID == TrustID)
+        {
+            trustData = g_PTrustList.at(index);
+            break;
+        }
+    }
```
ibm2431: Couldn't you find the index by subtracting the base Trust ID from the mob's Trust ID, instead of needing to step through the list?

sql/mob_pools.sql
```diff
@@ -5953,131 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
+INSERT INTO `mob_pools` VALUES (5906,'zeid','Zeid',146,0x0000C20B00000000000000000000000000000000,8,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5907,'lion','Lion',149,0x0000C30B00000000000000000000000000000000,6,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5908,'tenzen','Tenzen',149,0x0000C40B00000000000000000000000000000000,12,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5909,'mihli_aliapoh','MihliAliapoh',151,0x0000C50B00000000000000000000000000000000,3,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5910,'valaineral','Valaineral',145,0x0000C60B00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5911,'joachim','Joachim',149,0x0000C70B00000000000000000000000000000000,10,3,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5912,'naja_salaheem','NajaSalaheem',151,0x0000C80B00000000000000000000000000000000,6,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
```
ibm2431: Normal Naja should be a MNK, not a THF~

sql/mob_pools.sql
```diff
@@ -5953,131 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
+INSERT INTO `mob_pools` VALUES (5926,'mnejing','Mnejing',364,0x0000D60B00000000000000000000000000000000,7,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5927,'sakura','Sakura',55,0x0000D70B00000000000000000000000000000000,21,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5928,'luzaf','Luzaf',145,0x0000D80B00000000000000000000000000000000,17,13,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5929,'najelith','Najelith',149,0x0000D90B00000000000000000000000000000000,11,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5930,'aldo','Aldo',149,0x0000DA0B00000000000000000000000000000000,6,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5931,'moogle','Moogle',185,0x0000DB0B00000000000000000000000000000000,21,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5932,'fablinix','Fablinix',327,0x0000DC0B00000000000000000000000000000000,6,4,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
```
ibm2431, line 5993: Anyone have ideas how to handle "dual mains" like Fablinix and King of Hearts?

sql/mob_pools.sql
```diff
@@ -5953,131 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
+INSERT INTO `mob_pools` VALUES (5986,'teodor','Teodor',149,0x00001F0C00000000000000000000000000000000,4,6,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5987,'ullegore','Ullegore',74,0x0000210C00000000000000000000000000000000,4,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5988,'makki-chebukki','Makki-Chebukki',153,0x0000220C00000000000000000000000000000000,11,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5989,'king_of_hearts','KingOfHearts',61,0x0000230C00000000000000000000000000000000,5,3,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5990,'morimar','Morimar',149,0x0000240C00000000000000000000000000000000,9,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5991,'darrcuiln','Darrcuiln',489,0x0000250C00000000000000000000000000000000,1,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5992,'aahm','ArkHM',149,0x0000290C00000000000000000000000000000000,13,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
```
ibm2431: Wiki is a bit confusing; the main trust page lists AAHM as WAR/NIN (which is what I'd peg him as), but the specific page lists him as NIN/WAR instead.

sql/mob_pools.sql
```diff
@@ -5953,131 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
+INSERT INTO `mob_pools` VALUES (5993,'aaev','ArkEV',145,0x00002A0C00000000000000000000000000000000,7,3,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5994,'aamr','ArkMR',151,0x00002B0C00000000000000000000000000000000,9,6,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5995,'aatt','ArkTT',153,0x00002C0C00000000000000000000000000000000,4,8,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5996,'aagk','ArkGK',146,0x00002D0C00000000000000000000000000000000,12,14,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5997,'iroha','Iroha',149,0x0000270C00000000000000000000000000000000,4,12,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
```
ibm2431: Wiki is similarly confused for Iroha, listing her as SAM on the main page, but weirdly BLM/SAM on her specific page. It also lists her ROV version as WHM/SAM on its individual page, instead of SAM/WHM as it does on the main category listing.

Absent some concrete argument for main mage, I'd very much bet that Iroha is main SAM. Just a feeling 😉

sql/mob_pools.sql
```diff
@@ -5953,131 +5953,131 @@ INSERT INTO `mob_pools` VALUES (5892,'Euryale','Euryale',171,0x06007106000000000
...
+INSERT INTO `mob_pools` VALUES (6010,'zeid_ii','Zeid',146,0x00000E0C00000000000000000000000000000000,8,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (6011,'prishe_ii','Prishe',145,0x00000A0C00000000000000000000000000000000,2,3,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (6012,'nashmeira_ii','Nashmeira',149,0x00000B0C00000000000000000000000000000000,18,3,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (6013,'lilisette_ii','Lilisette',484,0x00000C0C00000000000000000000000000000000,19,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (6014,'tenzen_ii','Tenzen',149,0x0000190C00000000000000000000000000000000,12,11,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (6015,'mumor_ii','Mumor',149,0x0000200C00000000000000000000000000000000,4,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (6016,'ingrid_ii','Ingrid',149,0x00001E0C00000000000000000000000000000000,4,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
```
ibm2431: Ingrid II here is supposed to be WHM though.

Others: I have checked main/sub jobs for all other trusts as of this commit. You don't need to.

ibm2431 (changes requested):
I think we're pretty close to "merge ready"!

To others: With the exception of commits addressing review comments, I intend for this to be my last in-depth review for this PR! So if you think I may have missed something, get those reviews in~

zach2good:
@ibm2431 Thanks for the in-depth review 🙏

src/map/ai/controllers/trust_controller.cpp
zach2good: I remember reading a lot of complaints that JA/MA don't engage trusts anyway. When levelling a caster you still have to run in and take a swing at an enemy, then run away in order to let the trusts do their thing...

zach2good: This would be the sort of thing I'd like to match retail on by default but have a "fix" available in settings

src/map/ai/controllers/trust_controller.cpp
zach2good: The current movement code works about 70% of the time, every time. They do tend to form up like an actual party, but there are still instances of them getting stuck away from the mob, jostling for position, getting trapped by a wall of other trusts and being unable to get to the enemy etc.

src/map/ai/controllers/trust_controller.cpp
zach2good: Like my comment above, there are still instances of trust dance parties and them getting stuck in interesting situations... 

This will be a pretty hefty TODO in the future, but it's fun to play with right now

src/map/utils/trustutils.cpp
zach2good: I'll drop into retail later and check

src/map/utils/trustutils.cpp
zach2good: Unfortunately not, there is a gap in 1000-1003, so the ID's don't all line up perfectly with their position in TrustList. This can be simplified into `std::find`, I'd rather a search than an if/else for the regions either side of the gap

zach2good: To make matters EVEN worse, when SE added Monberaux, they didn't put him at the end of the second grouping of trusts, they put him _in the gap_.  Eugh.

src/map/party.cpp
zach2good: I decided not to treat trusts as full party members (they don't exist inside `members` and they don't return with `getParty()`). This might change in the future, but it'll take a lot more plumbing and loads of testing.

zach2good:
After addressing some points from IBM's review, I thought it might be useful to write up some of my choices about this PR for newcomers to this PR (if there are any):

**The crux of this entire PR:**
The packet logic change in `magic_state.cpp`. I don't know if this is correct, I found this while desperately wrestling with the client crash there. It works, it's stable and I haven't noticed any strangeness.

**Design decisions:**
I wanted Core to be totally unrestricted, and any restrictions to exist in Lua. The Lua restricts things down to a retail-like experience, but if operators wanted to clear out all of the restrictions in `tpz.trust.canCast`, there's nothing stopping them:
- Having trusts in Jeuno
- An alliance full of trusts
- 22 Shantottos nuking things

Trusts are "kind of" treated like party members. I've added `ForPartyWithTrusts` and `:getPartyWithTrusts` for anything that might need to iterate over the party including trusts. I did experiment with treating them as full members, but there were all sorts of explosions with Latents, `OnMobDeathEx`, etc. so I skirted around it.

**Magic, Skills and AI:**
What I've put in at the moment is just a fun frame that allows you to play with a party of Kupipi, Naji and Shantotto.

sql/mob_pools.sql
```diff
@@ -6050,12 +6050,12 @@ INSERT INTO `mob_pools` VALUES (5988,'makki-chebukki','Makki-Chebukki',153,0x000
 INSERT INTO `mob_pools` VALUES (5989,'king_of_hearts','KingOfHearts',61,0x0000230C00000000000000000000000000000000,5,3,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
 INSERT INTO `mob_pools` VALUES (5990,'morimar','Morimar',149,0x0000240C00000000000000000000000000000000,9,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
 INSERT INTO `mob_pools` VALUES (5991,'darrcuiln','Darrcuiln',489,0x0000250C00000000000000000000000000000000,1,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
-INSERT INTO `mob_pools` VALUES (5992,'aahm','ArkHM',149,0x0000290C00000000000000000000000000000000,13,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
+INSERT INTO `mob_pools` VALUES (5992,'aahm','ArkHM',149,0x0000290C00000000000000000000000000000000,1,13,3,240,100,0,0,0,0,0,0,32,0,3,0,0,0,1,0,0);
```
TeoTwawki, line 6053: Don't forget the weapon type in the core for mobs is currently bs, and that field controls if a mob double swings or not. IIRC hume ark angel dual wields. So you'd have to set his type "wrong" here to get correct behavior in that case.

ibm2431: "But /NIN gives Dual Wield!"
"Not until main level is 20."

If that's the case, then Luzaf and a few others might encounter difficulties too.

TeoTwawki: @ibm2431 Mobs, pets, and trusts don't obey those rules so that won't come into play at all. Whatever the pool value says they should do, they will do at any level.

And on retail trusts that dual wield will do so at any level afaik (I haven't seen any example to the contrary).

src/map/ai/controllers/trust_controller.cpp
ibm2431: Well, sometimes you want to pull the mob away from others without your trusts running forward to plant themselves in the middle of danger!

I think a happy medium might be a GM command.

src/map/ai/controllers/trust_controller.cpp
ibm2431: Alright, just making sure we were attempting to match retail! 👍

src/map/utils/trustutils.cpp
ibm2431: Ah yeah, the gap. Good call, then!

ibm2431 (approved):
I believe I am okay with this! There's still more to do (and we may want to compile a semi-official TODO list based off of review comments and things you've noticed), but this PR does what it set out to do and then some.

I am giving it a green checkmark.

PR #365, opened 2020-02-20
No xp if have battlefield status

KnowOne134:
**_I affirm:_**

tested and works


TeoTwawki:
suggestion: set m_giveExp false instead so `charutils::DistributeExperiencePoints(PChar, this);` isn't even called.

PR #369, opened 2020-02-21
if GM, don't annoy us with "you should set homepoint" message after job change.

TeoTwawki:
**_I affirm:_**
- [:crossed_fingers: ] that I've _tested my code_ since the last commit in the PR, and will test after any later commits


TeoTwawki:
this just adds `and player:getGMLevel() == 0` to the checks for if it should play that event or not. the rest is pure formatting to bring the files up to style spec.

PR #372, opened 2020-02-22
if charming a player that may be charmed, or in ballista bind them

KnowOne134:
**_I affirm:_**

tested with logging in 2 chars, having a mob charm 1 then using charm on the charmed char to give them a random bind effect for  to 5 seconds

### DIFF
--- a/scripts/globals/abilities/charm.lua
+++ b/scripts/globals/abilities/charm.lua
@@ -31,7 +31,7 @@ end
 
 function onUseAbility(player, target, ability)
     if target:isPC() then
-        target:addStatusEffect(tpz.effect.BIND, 0, 0, math.random(1,5))
+        ability:setMsg(tpz.msg.basic.NO_EFFECT)
     else
         local Tamed = false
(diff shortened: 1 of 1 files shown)


ibm2431:
> The following job abilities are altered: 
> [Charm, Gauge, and Tame have no effect whatsoever.](https://www.bg-wiki.com/bg/Category:Ballista)

Do we have a capture to confirm?

KnowOne134:
I know I'll ballista to be 100% true. Atm this is a game crash if you try
it. Makes sense to me that you cant charm a charmed player to you either
will get a cannot do that on target or this

ibm2431:
I will do my best to remember to test retail Charm Ballista behavior this period if someone else doesn't beat me to it.

ibm2431 (changes requested):
We've talked on Discord, but I'm adding a "changes requested" note here so others know that retail does "no effect"~!

ibm2431 (approved):

PR #374, opened 2020-02-22
Bulk remove semicolons from most of the Lua in scripts/globals.

mrhappyasthma:
Command applied:
find . -type f -name "*.lua" | xargs -d '\n' sed -i 's/;//g'

Semicolons are not appropriate in Lua per the style guidelines:
https://github.com/project-topaz/topaz/blob/master/CONTRIBUTING.md

NOTE: I skipped /globals/weaponskills for now, since many files have
a valid use of semicolons (for multiple statements on the same line).
I'll audit those separately.

This change is just a style change, no functionality has been changed.

**_I affirm:_**

### DIFF
--- a/scripts/globals/ability.lua
+++ b/scripts/globals/ability.lua
@@ -3,10 +3,10 @@
 -- ABILITIES
 --
 -----------------------------------
-require("scripts/globals/status");
-require("scripts/globals/msg");
+require("scripts/globals/status")
+require("scripts/globals/msg")
 
-tpz = tpz or {};
+tpz = tpz or {}
 
 tpz.jobAbility =
 {
@@ -428,8 +428,8 @@ tpz.jobAbility =
     CHAOTIC_STRIKE     = 614,
     THUNDERSTORM       = 615,
     JUDGMENT_BOLT      = 616,
-};
-tpz.ja = tpz.jobAbility;
+}
+tpz.ja = tpz.jobAbility
 
 tpz.reaction =
 {
@@ -440,7 +440,7 @@ tpz.reaction =
     HIT      = 0x08,
     EVADE    = 0x09,
     GUARD    = 0x14,
-};
+}
 
 tpz.specEffect =
 {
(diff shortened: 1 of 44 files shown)


mrhappyasthma:
PR is reporting an error on continuous build from Travis. Will investigate.

mrhappyasthma:
The issue has been addressed. Turns out tables need a delimiter.

TeoTwawki:
Generally a comma is used instead of a semicolon for table delimiters

mrhappyasthma:
Is that preferable? I can update it to use a comma instead.

mrhappyasthma:
I updated it to use commas. It's a bit strange, the odd numbered indexes are ever needed. Despite being a combination of inner tables and integers, only the inner table are used -- simulating a 2D array.

But idk how to test Limbus because I'm a FFXI noob, so I didn't wanna mess with it within this PR.

https://github.com/project-topaz/topaz/blob/2943fab983cc1457ce75f15dd9dca9d3bfa21b05/scripts/globals/limbus.lua#L65-L75

```
MIMICPOSITION =
{
    1, {-363, 0, -282};
    2, {-359, 0, -277};
    3, {-326, 0, -301};
    4, {-331, 0, -330};
    5, {-340, 0, -330};
    6, {-345, 0, -311};
    7, {-339, 0, -300};
    8, {-335, 0, -281};
}
```

Could easily be:

```
MIMICPOSITION =
{
    {-363, 0, -282};
    {-359, 0, -277};
    {-326, 0, -301};
    {-331, 0, -330};
    {-340, 0, -330};
    {-345, 0, -311};
    {-339, 0, -300};
    {-335, 0, -281};
}
```

Although then you'd need to update the references here:
https://github.com/project-topaz/topaz/blob/2943fab983cc1457ce75f15dd9dca9d3bfa21b05/scripts/globals/limbus.lua#L603-L616

KnowOne134:
best would be not to touch limbus as its a work in progress, any changes to that area can cause un needed rebasing, merge conflicts

mrhappyasthma:
In that case I'll just revert that file. Its only a few lines and can easily be cleaned up later once the limbus.lua is more finalized.

mrhappyasthma:
Amended the commit to drop the changes to limbus.lua.

Issue #377, opened 2020-02-23
ATTACK_BREATH is unused, possibly by design.


mrhappyasthma:
There's an `ATTACK_BREATH` enum value (and similar one in Lua). But it's not used anywhere.

It's possible this may not be needed, since most breath spells are listed as "magical" by most FFXI websites. e.g. "heat breath" is labeled as "fire" + "magical".

Although if it was intended to be used for something, it's not.

https://github.com/DarkstarProject/darkstar/blob/master/src/map/entities/battleentity.h#L244

https://github.com/DarkstarProject/darkstar/blob/master/scripts/globals/status.lua#L2123

mrhappyasthma:
I'm in the process of updating the `attackType` for all blu spells. I'm going to follow the convention of labeling them as MAGICAL for now.

If we decide this is something we want used, I can go through and change the breath spells to BREATH.

mrhappyasthma:
Per discussion with the appropriate stakeholders, Breath spells will be fixed in the future to have proper damage calculations. I'll update my outstanding PR to label the breath spells as such in preparation for that.

mrhappyasthma:
Issue to fix breath damage calculation appears to be here: https://github.com/project-topaz/topaz/issues/239

PR #378, opened 2020-02-23
Add `attackType` values to all BLU spells.

### LINKED ISSUE #314: BLU Magical Damage Spells Treated as attackType.PHYSICAL
<a href="https://github.com/ibm2431"></img></a> **Issue by [ibm2431](https://github.com/ibm2431)**
_Sunday Sep 22, 2019 at 21:34 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 6225_

----

**_I have:_**

**_Client Version_** (type `/ver` in game) **:** n/a

Before I forget to publicly note this (again): Damage-dealing Blue Magic that goes through `BlueFinalAdjustments` is always treated as `attackType.PHYSICAL`, even if the damage type should be Magical.

Some examples of Blue Spells that should be treated as Magical - but aren't due to `BlueFinalAdjustments` - are:

Death Ray
> github/DarkstarProject/darkstar/blob/master/scripts/globals/spells/bluemagic/death_ray.luaDarkstar Issue L42

Ice Break
> github/DarkstarProject/darkstar/blob/master/scripts/globals/spells/bluemagic/ice_break.luaDarkstar Issue L50

Mind Blast
> github/DarkstarProject/darkstar/blob/master/scripts/globals/spells/bluemagic/mind_blast.luaDarkstar Issue L44

When they get into `BlueFinalAdjustments`, after calculating reductions due to Phalanx and Stoneskin, `takeDamage` is always called with `attackType.PHYSICAL`:

> github/DarkstarProject/darkstar/blob/master/scripts/globals/bluemagic.luaDarkstar Issue L232-L247

mrhappyasthma:
Also updated bluemagic.lua to properly default to NONE for
any non-damage dealing spells.

This commit follows updates breath spells to use the BREATH enum value, per:
https://github.com/project-topaz/topaz/issues/377

Fixes https://github.com/project-topaz/topaz/issues/314

**_I affirm:_**


KnowOne134:
In the future the breath spells will need the breath enum or another call to bluemagic lua to get its correct formula as it doesnt follow the magical formula at all, not even its magic burst formula

ibm2431:
If we're going to redo BLU soon-ish, I'd say keep breath spells labeled as breath spells, because whether a spell is a breath attack _will_ be accounted for.

mrhappyasthma:
Sounds good. I'll update the breath spells after a rebase. Will also add context to the issue.

mrhappyasthma:
Rebased and updated breath spells to use `attachType.BREATH`.

scripts/globals/spells/bluemagic/feather_storm.lua
```diff
@@ -24,22 +24,23 @@ end
...
-        params.duppercap = 17
-        params.str_wsc = 0.0
-        params.dex_wsc = 0.0
-        params.vit_wsc = 0.0
-        params.agi_wsc = 0.30
-        params.int_wsc = 0.0
-        params.mnd_wsc = 0.0
-        params.chr_wsc = 0.0
+    params.tpmod = TPMOD_CRITICAL
+    params.attackType = tpz.attackType.PHYSICAL
```
ibm2431: Do we want a different identifier for ranged physical spells?

scripts/globals/spells/bluemagic/asuran_claws.lua
```diff
@@ -25,6 +25,7 @@ function onSpellCast(caster,target,spell)
     local params = {}
     -- This data should match information on http://wiki.ffxiclopedia.org/wiki/Calculating_Blue_Magic_Damage
         params.tpmod = TPMOD_ACC
+        params.attackType = tpz.attackType.PHYSICAL
         params.damageType = tpz.damageType.BLUNT
```
ibm2431: Not necessarily a question directed at you, mrhappyasthma (unless you _want_ to take a guess):

Bgwiki has two different classifications for spells whose damage type _I_ would normally consider "blunt" - they have identifiers for both "Blunt" and "Hand-to-hand". In this instance, they identify Asuran Claws as "hand to hand".

Are there theories on why they might do this, and is it something we should be mindful of?

scripts/globals/spells/bluemagic/hysteric_barrage.lua
```diff
@@ -24,24 +24,25 @@ end
...
-        params.str_wsc = 0.0
-        params.dex_wsc = 0.3
-        params.vit_wsc = 0.0
-        params.agi_wsc = 0.0
-        params.int_wsc = 0.0
-        params.mnd_wsc = 0.0
-        params.chr_wsc = 0.0
+    params.tpmod = TPMOD_DAMAGE
+    params.attackType = tpz.attackType.PHYSICAL
+    params.damageType = tpz.damageType.HTH
```
ibm2431: Because down here, we _are_ using the "hth" damage type, when for all the "hand-to-hand" (as identified on bgwiki) before this were simply "blunt".

scripts/globals/spells/bluemagic/queasyshroom.lua
```diff
@@ -24,22 +24,23 @@ end
...
-        params.duppercap = 15
-        params.str_wsc = 0.0
-        params.dex_wsc = 0.0
-        params.vit_wsc = 0.0
-        params.agi_wsc = 0.0
-        params.int_wsc = 0.20
-        params.mnd_wsc = 0.0
-        params.chr_wsc = 0.0
+    params.tpmod = TPMOD_CRITICAL
+    params.attackType = tpz.attackType.PHYSICAL
```
ibm2431: Marking queasyshroom here as another "ranged" spell.

scripts/globals/spells/bluemagic/uppercut.lua
```diff
@@ -24,25 +24,26 @@ end
...
-        params.str_wsc = 0.35
-        params.dex_wsc = 0.0
-        params.vit_wsc = 0.0
-        params.agi_wsc = 0.0
-        params.int_wsc = 0.0
-        params.mnd_wsc = 0.0
-        params.chr_wsc = 0.0
+    params.tpmod = TPMOD_ATTACK
+    params.attackType = tpz.attackType.PHYSICAL
+    params.damageType = tpz.damageType.HTH
```
ibm2431: And for example, here we have it as "hand-to-hand" when bgwiki lists it as "blunt"

(I know you weren't responsible for this, mrhappyasthma)

ibm2431 (commented):
Verified that all damage types are correct.

Although I do have a few questions, directed at everyone interested in BLU.

(If it weren't for the "what do we do about ranged spells" question, I would have put an approve on this.)

scripts/globals/spells/bluemagic/asuran_claws.lua
TeoTwawki: @ibm2431 there actually are different dmg resistances between the 2 on mobs. there are only a few mobs where that difference matters however, and I don't know that anyone confirmed it before tossing it onto the wiki.

mrhappyasthma: Would we prefer to update this to H2H then?

scripts/globals/spells/bluemagic/hysteric_barrage.lua
mrhappyasthma: I grabbed the H2H classification from here: https://ffxiclopedia.fandom.com/wiki/Hysteric_Barrage

scripts/globals/spells/bluemagic/uppercut.lua
mrhappyasthma: This should be blunt, it seems. From the various websites that have it documented as such.

I can't find any reference to H2H

scripts/globals/spells/bluemagic/feather_storm.lua
mrhappyasthma: These aren't labeled as range in a very noticeable spot. But seems like the following 3 are classified as ranged:

Queasyshroom
Feather Storm
Pinecone Bomb

scripts/globals/spells/bluemagic/asuran_claws.lua
mrhappyasthma: Interestingly this is `Blunt` here: https://ffxiclopedia.fandom.com/wiki/Asuran_Claws

But in the comments they say:

> Hand-to-Hand-based damage.

scripts/globals/spells/bluemagic/feather_storm.lua
KnowOne134: should be changed to ranged yes

scripts/globals/spells/bluemagic/hysteric_barrage.lua
mrhappyasthma: Seems to be H2H here also: https://www.bg-wiki.com/bg/Hysteric_Barrage

scripts/globals/spells/bluemagic/asuran_claws.lua
mrhappyasthma: It is marked H2H here: https://www.bg-wiki.com/bg/Asuran_Claws

So I'll go with that since it seems to be the correct one.

scripts/globals/spells/bluemagic/feather_storm.lua
mrhappyasthma: Done.

scripts/globals/spells/bluemagic/asuran_claws.lua
mrhappyasthma: Done.

scripts/globals/spells/bluemagic/uppercut.lua
mrhappyasthma: Done.

scripts/globals/spells/bluemagic/queasyshroom.lua
mrhappyasthma: Done.

mrhappyasthma:
Consensus seemed to be that the 3 RANGED spells should be annotated as such. I applied those as well as fixed the two mistakes in the attackTypes.

ibm2431 (approved):
`attackTypes` look good to me, thanks~!

I wasn't fully examining `damageType`s, but that wasn't in the original scope of the PR, so I'm putting an Approval on this before more PR creep sets in!

Going to tag @project-topaz/researcher in case they feel like vetting all the blunt/hth distinctions for the BLU project focus.

PR #379, opened 2020-02-23
Add a simple flow to allow changing user passwords.

mrhappyasthma:
This is a prerequisite for DarkstarProject/xiloader#7.

This change adds support for a new command in the login code to support
changing a users password. It simply verifies the users login similarly
to the existing flow, and then makes another request over the socket for
the new password. Then it updates the SQL.

Tested locally. Have a set of changes ready for xiloader as well, but
this work is required first.

NOTE: There is definitely some refactoring that could be done in this
file to reduce the amount of nested braces and to reduce duplicated code.
I tried to touch as little existing code as possible in this change to
reduce the risk of introducing a regression.

**_I affirm:_**

### DIFF
--- a/src/login/login_auth.cpp
+++ b/src/login/login_auth.cpp
@@ -282,7 +282,7 @@ int32 login_parse(int32 fd)
             {
                 session[fd]->wdata.resize(1);
                 ref<uint8>(session[fd]->wdata.data(), 0) = LOGIN_ERROR;
-                ShowWarning("login_parse: unexisting user" CL_WHITE"<%s>" CL_RESET" tried to connect\n", escaped_name);
+                ShowWarning("login_parse: user" CL_WHITE"<%s>" CL_RESET" could not be found using the provided information. Aborting.\n", escaped_name);
                 do_close_login(sd, fd);
                 return 0;
             }
--- a/src/login/login_auth.cpp
+++ b/src/login/login_auth.cpp
@@ -100,6 +100,9 @@ int32 login_parse(int32 fd)
             return -1;
         }
 
+        Sql_EscapeString(SqlHandle, escaped_name, name.c_str());
+        Sql_EscapeString(SqlHandle, escaped_pass, password.c_str());
+
         switch (code)
         {
         case LOGIN_ATTEMPT:
@@ -107,8 +110,6 @@ int32 login_parse(int32 fd)
             const char* fmtQuery = "SELECT accounts.id,accounts.status \
                                     FROM accounts \
                                     WHERE accounts.login = '%s' AND accounts.password = PASSWORD('%s')";
-            Sql_EscapeString(SqlHandle, escaped_name, name.c_str());
-            Sql_EscapeString(SqlHandle, escaped_pass, password.c_str());
             int32 ret = Sql_Query(SqlHandle, fmtQuery, escaped_name, escaped_pass);
             if (ret != SQL_ERROR  && Sql_NumRows(SqlHandle) != 0)
             {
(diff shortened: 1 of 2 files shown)


src/login/login_auth.cpp
```diff
@@ -273,6 +272,87 @@ int32 login_parse(int32 fd)
...
+        {
+            const char* fmtQuery = "SELECT accounts.id,accounts.status \
+                                    FROM accounts \
+                                    WHERE accounts.login = '%s' AND accounts.password = PASSWORD('%s')";
+            int32 ret = Sql_Query(SqlHandle, fmtQuery, escaped_name, escaped_pass);
+            if (ret == SQL_ERROR || Sql_NumRows(SqlHandle) == 0)
+            {
+                session[fd]->wdata.resize(1);
+                ref<uint8>(session[fd]->wdata.data(), 0) = LOGIN_ERROR;
+                ShowWarning("login_parse: unexisting user" CL_WHITE"<%s>" CL_RESET" tried to connect\n", escaped_name);
```
ibm2431: Technically this warning can be triggered on either the account not existing _or_ a password mismatch

src/login/login_auth.cpp
```diff
@@ -273,6 +272,87 @@ int32 login_parse(int32 fd)
...
+
+            sd->accid = (uint32)Sql_GetUIntData(SqlHandle, 0);
+            uint8 status = (uint8)Sql_GetUIntData(SqlHandle, 1);
+
+            if (status & ACCST_BANNED)
+            {
+                session[fd]->wdata.resize(1);
+                ref<uint8>(session[fd]->wdata.data(), 0) = LOGIN_ERROR_CHANGE_PASSWORD;
+                ShowInfo("login_parse: banned user" CL_WHITE"<%s>" CL_RESET" detected. Aborting.\n", escaped_name);
+                do_close_login(sd, fd);
```
ibm2431: Is there no need for a memset like with [during the login flow](https://github.com/project-topaz/topaz/blob/821bcfe6b8db045a03694d032bde530e8152585f/src/login/login_auth.cpp#L165-L170)?

mrhappyasthma: It shouldn't be needed since I'm resizing the `wdata` (which is a `std::string`). Per the documentation:

http://www.cplusplus.com/reference/string/string/resize/

> If n is smaller than the current string length, the current value is shortened to its first n character, removing the characters beyond the nth.

So the code resizes the string to a single byte, then we set that byte value to `LOGIN_ERROR_CHANGE_PASSWORD`.

So there's no bytes that contain garbage or need to be copied over with memset.

mrhappyasthma: This pattern is used elsewhere in the code as well: https://github.com/project-topaz/topaz/blob/821bcfe6b8db045a03694d032bde530e8152585f/src/login/login_auth.cpp#L97-L98

src/login/login_auth.cpp
mrhappyasthma: Fair point. Let me adjust the wording of this error to be more clear.

mrhappyasthma: For context, this was copied from this line: 
https://github.com/project-topaz/topaz/blob/821bcfe6b8db045a03694d032bde530e8152585f/src/login/login_auth.cpp#L205

It suffers from the same ambiguity problem.

mrhappyasthma: A lot of code in this file can be cleaned up and optimized. I tried to make as minimal changes as possible to reduce the risk of introducing a regression bug.

When I get some time I can try to clean up this file and provide a bit more structure and better organization for maintainability.

src/login/login_auth.cpp
ibm2431: Alright, just curious since I saw a memset was being used for bans later on, and I'm not gonna pretend I'm very familiar with the login server~! 😅

src/login/login_auth.cpp
mrhappyasthma: Updated the wording here. Please take another look and let me know what you think.

ibm2431: New wording looks good~! 👍

src/login/login_auth.cpp
mrhappyasthma: As I mentioned in the other thread. A lot of this code can be pulled into helper functions and consolidated for better readability and maintainability. However there's a lot of copy/paste style code blocks throughout here that I didn't wanna mess with for the sake of this change.

A more solid rewrite of this file should be done with care and tested thoroughly.

PR #390, opened 2020-02-25
Docking Message when traveling to Nashmau and Al Zahbi

kaincenteno:
Added ID for message (as it's not part of CS) for boat silver sea routes
added onspawn on boat captains that handle triggers based on vana epoch
create transport global (so it can be used for other boats and airships)
also removed semicolons for those files

**_I affirm:_**


ibm2431 (approved):

KnowOne134 (approved):
tested and working

PR #394, opened 2020-02-26
Default Confs now in subfolder. No more accidentally sending your password

TeoTwawki:
**_I affirm:_**

Moved them to a new subdirectory and added the original location to ignore list. Put a read me file in the original location to tell new people how to use the confs. Perhaps a afutture update can auto-copy any "missing" file from the default directory.

Also: adjusted some settings to be modern retail in the defaults. Existing servers will be retaining modified confs, this won't affect them. 

**Left speed settings alone** because while modern retails base has moved from 40 to 50, the database has a mix of the 2 for mobs and NPCs. And I think for players the base for what "0" is should be modified before changing the default setting here. Right now 0 = 40 and we I feel should either make that 50 or make 0 actually be 0 and set the default setting to 50.


ibm2431 (approved):

conf/default/map.conf
```diff
@@ -100,8 +100,8 @@ speed_mod: 0
 mob_speed_mod: 0
 
 #Allows you to manipulate the constant multiplier in the skill-up rate formulas, having a potent effect on skill-up rates.
-skillup_chance_multiplier: 2.5
-craft_chance_multiplier: 2.6
+skillup_chance_multiplier: 1.0
```
kaincenteno:
```suggestion
skillup_chance_multiplier: 1.0
craft_chance_multiplier: 1.0
```
Is this the current chance on retail?

TeoTwawki: See the commit comment. Settings should be for adjusting _from_ retail, but I have my personal doubts as to the retail accuracy of what should be the baseline stock retail level. As far as I know every server has this above 1.0

TeoTwawki: See the commit comment. Settings should be for adjusting _from_ retail, but I have my personal doubts as to the retail accuracy of what should be the baseline stock retail level. As far as I know every server has this above 1.0

TeoTwawki: See the commit comment. Settings should be for adjusting _from_ retail, but I have my personal doubts as to the retail accuracy of what should be the baseline stock retail level. As far as I know every server has this above 1.0

TeoTwawki: See the commit comment. Settings should be for adjusting _from_ retail, but I have my personal doubts as to the retail accuracy of what should be the baseline stock retail level. As far as I know every server has this above 1.0

TeoTwawki: See the commit comment. Settings should be for adjusting _from_ retail, but I have my personal doubts as to the retail accuracy of what should be the baseline stock retail level. As far as I know every server has this above 1.0

TeoTwawki: Thanks shithub

PR #395, opened 2020-02-26
Updated levels to match current retail

kaincenteno:
**_I affirm:_**


TeoTwawki:
https://github.com/project-topaz/topaz/blob/master/conf/map.conf#L170
didn't need updated, tried to tell you in discord

kaincenteno:
changed everything that has ismission = 1 to cap at 75 (as some of them were higher than that level)
reverted all changes to anything level 75 and below

PR #398, opened 2020-02-27
only players currently on Dark Spark quest will get credit for the kill

### LINKED ISSUE #371: Credit for helping to defeat Dark Spark persists, even when not on AF hand quest
**_I have:_**

In the script for Dark Spark, [no check is done in the `onMobDeath` before setting a `charVar` giving the player credit for killing it](https://github.com/project-topaz/topaz/blob/master/scripts/zones/Castle_Zvahl_Baileys/mobs/Dark_Spark.lua#L10-L12). This includes when you are assisting someone else with the quest and are not on the quest yourself.

If you later try to complete the quest yourself, and [manage to select the same torch as the last one which Dark Spark was spawned from (barring anything which would wipe the mob's local variable), you will instantly be given the key item without spawning it, due to already possessing credit](https://github.com/project-topaz/topaz/blob/master/scripts/zones/Castle_Zvahl_Baileys/npcs/Torch.lua#L19-L20).

wrenffxi:
**_I affirm:_**

Fixes #371

### DIFF
--- a/scripts/zones/Castle_Zvahl_Baileys/mobs/Dark_Spark.lua
+++ b/scripts/zones/Castle_Zvahl_Baileys/mobs/Dark_Spark.lua
@@ -4,9 +4,16 @@
 -- Involved in Quests: Borghertz's Hands (AF Hands, Many job)
 -- !pos 63 -24 21 161
 -----------------------------------
-require("scripts/globals/settings");
+require("scripts/globals/keyitems")
 -----------------------------------
 
 function onMobDeath(mob, player, isKiller)
-    player:setCharVar("BorghertzSparkKilled",1);
-end;
+    if
+        player:getCharVar("BorghertzSparkKilled") == 0 and
+        player:hasKeyItem(tpz.ki.OLD_GAUNTLETS) and
+        not player:hasKeyItem(tpz.ki.SHADOW_FLAMES) and
+        player:getCharVar("BorghertzCS") >= 2
+    then
+        player:setCharVar("BorghertzSparkKilled", 1)
+    end
+end
--- a/scripts/zones/Castle_Zvahl_Baileys/mobs/Dark_Spark.lua
+++ b/scripts/zones/Castle_Zvahl_Baileys/mobs/Dark_Spark.lua
@@ -14,6 +14,6 @@ function onMobDeath(mob, player, isKiller)
         not player:hasKeyItem(tpz.ki.SHADOW_FLAMES) and
         player:getCharVar("BorghertzCS") >= 2
     then
-        player:setCharVar("BorghertzSparkKilled", mob:getLocalVar("fromTorch"))
+        player:setCharVar("BorghertzSparkKilled", 1)
     end
 end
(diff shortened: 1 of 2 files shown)


scripts/zones/Castle_Zvahl_Baileys/mobs/Dark_Spark.lua
```diff
@@ -4,9 +4,16 @@
...
-require("scripts/globals/settings");
+require("scripts/globals/keyitems")
 -----------------------------------
 
 function onMobDeath(mob, player, isKiller)
-    player:setCharVar("BorghertzSparkKilled",1);
-end;
\ No newline at end of file
+    if
+        player:getCharVar("BorghertzSparkKilled") == 0 and
```
KnowOne134: mind if we change this to a local var same as the torch ID, if you logout with out touchng you would lose the torch refrence anyway. this particular var check on this line really isnt needed though since we have the other 2 checks.

KnowOne134 (approved):
using player:setLocalVar("BorghertzSparkKilled", 1) on mob death, removing if player:getCharVar("BorghertzSparkKilled") == 0 on the death and changing the var on the torch trigger for after killing DS to player:setLocalVar("BorghertzSparkKilled", 0)

sugestions are welcome from the others and works as is

scripts/zones/Castle_Zvahl_Baileys/mobs/Dark_Spark.lua
TeoTwawki: if you get DC'd do you have to repeat the fight though?

KnowOne134: Would have to anyway becuase it's also checking a local var vet on the torch on popping it and retrieving

TeoTwawki: I mean: whats retail do.

just..we're here anyway. ¯\\\_(ツ)\_/¯

megberry: Agree that we should mimic retail.  If someone with access can tell me how it works there, I can code to match.

Until then, how about this new commit?  This saves the torch used to spawn the crediting Spark to each player on the quest who helps kill it.  This way, the particular torch will persist per-player across logouts, or if someone else comes and kills a different Dark Spark before original player has a chance to touch the torch.

wrenffxi: Agree that we should mimic retail.  If someone with access can tell me how it works there, I can code to match.

Until then, how about this new commit?  This saves the torch used to spawn the crediting Spark to each player on the quest who helps kill it.  This way, the particular torch will persist per-player across logouts, or if someone else comes and kills a different Dark Spark before original player has a chance to touch the torch.

KnowOne134 (approved):
i can agree on that, anything can be changed later

wrenffxi:
Nyu tested this on retail.  Turns out it doesn't matter which torch is clicked: any of them will give the reward after Dark Spark is killed.  Also, if you kill Dark Spark, then log out and back in, then click a torch, you will be rewarded.

I have updated the PR to match retail testing.

ibm2431 (approved):

PR #406, opened 2020-03-07
script was checking for attack down instead of defense down

59blargedy:
**_I affirm:_**

Did not test, but minor fix so felt safe.

### DIFF
--- a/scripts/globals/weaponskills/metatron_torment.lua
+++ b/scripts/globals/weaponskills/metatron_torment.lua
@@ -39,11 +39,9 @@ function onUseWeaponSkill(player, target, wsID, tp, primary, action, taChar)
 
     local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, wsID, params, tp, action, primary, taChar)
     if damage > 0 then
-        if not target:hasStatusEffect(tpz.effect.DEFENSE_DOWN) then
-            local duration = tp / 1000 * 20 * applyResistanceAddEffect(player, target, tpz.magic.ele.WIND, 0)
-            target:addStatusEffect(tpz.effect.DEFENSE_DOWN, 19, 0, duration)
-        end
-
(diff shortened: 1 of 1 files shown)


scripts/globals/weaponskills/metatron_torment.lua
```diff
@@ -39,7 +39,7 @@ function onUseWeaponSkill(player, target, wsID, tp, primary, action, taChar)
 
     local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, wsID, params, tp, action, primary, taChar)
     if damage > 0 then
-        if not target:hasStatusEffect(tpz.effect.ATTACK_DOWN) then
+        if not target:hasStatusEffect(tpz.effect.DEFENSE_DOWN) then
```
TeoTwawki: it doesn't even need to check - defense down naturally doesn't reapply unless the new effects power is greater than the existing effect.

59blargedy: Thank you - I'm ridiculously new at this so forgive the question - does that mean i should just delete the if not statement entirely as it is checking that elsewhere? REmoving it is of some benefit since it was checking attack_down, right?

TeoTwawki: yes you want that block to become:
```lua
    if damage > 0 then
        local duration = tp / 1000 * 20 * applyResistanceAddEffect(player, target, tpz.magic.ele.WIND, 0)
        target:addStatusEffect(tpz.effect.DEFENSE_DOWN, 19, 0, duration)

        -- Apply aftermath
        tpz.aftermath.addStatusEffect(player, tp, tpz.slot.MAIN, tpz.aftermath.type.RELIC)
    end
```
note the adjusted indent on the part that was inside the if-then-end pairing

59blargedy: cool, thanks i changed and committed again so hopefully that is right.

TeoTwawki: looks good :+1:

PR #411, opened 2020-03-12
Fixing Ferries in Mhaura

kaincenteno:
* Offset was wrong and time at dock was also wrong. This was fixed.
* Issue with docking position for Ferries is still wrong, but pos in table is now retail accurate.
* NPC at dock in Mhaura now moves when ferry arrives and departs, also tells you about it,
not 100% retail accurate as it needs to rotate when it reaches final location. Leaving it as a TODO
for someone else to complete it.

**_I affirm:_**


scripts/globals/transport.lua
```diff
@@ -24,6 +24,33 @@ tpz.transport.messageTime =
     SILVER_SEA = 480
 }
 
+tpz.transport.dockPorterMhauraTrigger =
```
ibm2431: Can probably shorten these; they're already in the `tpz.transport` namespace, so `Porter` is already implied. 😉 

```
tpz.transport.trigger.mhaura
tpz.transport.offset.mhaura
tpz.transport.interval.mhaura
tpz.transport.pos.mhaura
```

scripts/globals/transport.lua
```diff
@@ -24,6 +24,33 @@ tpz.transport.messageTime =
...
+{
+    FERRY_ARRIVING_FROM_ALZAHBI  = 0,
+    FERRY_DEPARTING_TO_ALZAHBI = 1,
+    FERRY_ARRIVING_FROM_SELBINA  = 2,
+    FERRY_DEPARTING_TO_SELBINA = 3
```
ibm2431: Picky ibm: Could you make sure the `=` line up all pretty like~?

scripts/globals/transport.lua
```diff
@@ -33,4 +60,15 @@ tpz.transport.captainMessage = function(npc, triggerID, messages)
     for _, player in pairs(playersInZone) do
         player:showText(player, messages[triggerID])
     end
+end
+
+tpz.transport.dockPorterMhauraMessages = function(npc, triggerID, messages)
+    npc:showText(npc, messages[triggerID])
+    if triggerID == tpz.transport.dockPorterMhauraTrigger.FERRY_ARRIVING_FROM_ALZAHBI or
+        triggerID == tpz.transport.dockPorterMhauraTrigger.FERRY_ARRIVING_FROM_SELBINA then
```
ibm2431: Here's a trick you could do to save some lines and space, since your "arriving" and "departing" are all even/odd:

```lua
if (triggerID % 2) == 0 then
    npc:pathThrough(-- arriving)
else
    npc:pathThrough(-- departing)
end
```

scripts/globals/transport.lua
```diff
@@ -33,4 +60,15 @@ tpz.transport.captainMessage = function(npc, triggerID, messages)
     for _, player in pairs(playersInZone) do
         player:showText(player, messages[triggerID])
     end
+end
+
+tpz.transport.dockPorterMhauraMessages = function(npc, triggerID, messages)
```
ibm2431: Not going to force you to globalize this _juuusstt_ yet, but you can save yourself a little effort later by:
```lua
tpz.transport.messages = {}
tpz.transport.messages.mhaura = function(npc, triggerID, messages)
```

(Later we can extend this to instead be just one function which takes a `dock`, and then pulls the appropriate tables based on the dock, but I realize right now that the focus is to fix the ferry)

scripts/zones/Mhaura/npcs/Dieh_Yamilsiah.lua
```diff
@@ -4,6 +4,38 @@
...
+    npc:addPeriodicTrigger(tpz.transport.dockPorterMhauraTrigger.FERRY_ARRIVING_FROM_SELBINA,
+        tpz.transport.dockPorterMhauraInterval.FROM_TO_SELBINA,
+        tpz.transport.dockPorterMhauraOffset.FERRY_ARRIVING_FROM_SELBINA)
+    npc:addPeriodicTrigger(tpz.transport.dockPorterMhauraTrigger.FERRY_DEPARTING_TO_SELBINA,
+        tpz.transport.dockPorterMhauraInterval.FROM_TO_SELBINA,
+        tpz.transport.dockPorterMhauraOffset.FERRY_DEPARTING_TO_SELBINA)
+end
+
+function onTimeTrigger(npc, triggerID)
+    tpz.transport.dockPorterMhauraMessages(npc, triggerID, messages)
```
ibm2431: And then your function call down here becomes: `tpz.transport.messages.mhaura(npc, triggerID, messages)`

ibm2431 (changes requested):
Thanks, kain~!

Mainly just some variable shortening and prettification:

ibm2431:
~~Developer: I _was_ going to have this target the `transport` branch, but the commit history did _not_ like that, so I punted back to `release`. I'll have to see what's going on with the `transport` branch later.~~

`transport` is just quite a ways behind `release`, which is what this was based on.

ibm2431 (approved):
Pushed review changes~!

(And she seems to be working!)

PR #419, opened 2020-03-13
make all NMs not charmable

KnowOne134:
**_I affirm:_**

not all NMs were being tagged on initialize for being uncharmable. this will correct that.
if you need a NM to be charmable will have to tag it with mobMod Charmable, 1 anywhere in script beside onMobInitialize

### DIFF
--- a/src/map/instance_loader.cpp
+++ b/src/map/instance_loader.cpp
@@ -230,6 +230,14 @@ CInstance* CInstanceLoader::LoadInstance(CInstance* instance)
 
             PMob->setMobMod(MOBMOD_CHARMABLE, Sql_GetUIntData(SqlInstanceHandle, 65));
 
+            // Overwrite base family charmables depending on mob type. Disallowed mobs which should be charmable
+            // can be set in mob_spawn_mods or in their onInitialize
+            if (PMob->m_Type & MOBTYPE_EVENT || PMob->m_Type & MOBTYPE_FISHED || PMob->m_Type & MOBTYPE_BATTLEFIELD ||
+                PMob->m_Type & MOBTYPE_NOTORIOUS)
+            {
+                PMob->setMobMod(MOBMOD_CHARMABLE, 0);
+            }
+
             // must be here first to define mobmods
             mobutils::InitializeMob(PMob, zone);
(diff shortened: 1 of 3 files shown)


TeoTwawki (commented):
Would be better to have the nm check in the other place you removed it from before so that the mobmod can work even in onmobinit. It just needed correct syntax to exempt nms that had the mobmod 
https://github.com/project-topaz/topaz/pull/332/files#diff-1edaebbb5ff57fedc3a066c6ae7c62fdL3829

ibm2431 (commented):
I know you're appending NMs to where the charmable mod already is, but should this assignment _be_ in `calculateStats`?

I could be wrong, but [`LoadMOBList` might be a better candidate, as it's already attempting to set the charmable mod, and it occurs before the call to InitializeMob, so I want to say onMobInitialize should be able to overwrite the default assignment](https://github.com/project-topaz/topaz/blob/release/src/map/utils/zoneutils.cpp#L494-L509).

If I'm reading this right, [`instance_loader` might need to be modified too.](https://github.com/project-topaz/topaz/blob/release/src/map/instance_loader.cpp#L231-L234)

KnowOne134:
Should I delete all this logic and move to instance loader and zone loader?
https://github.com/project-topaz/topaz/blob/8985d2af68876367592222836db1f8e5b15c8175/src/map/utils/mobutils.cpp#L564

KnowOne134:
As it is I'm just adding to an already in place system

ibm2431:
I think that setting charm mod should never have been in `calculateStats`. The function is used to [recalculate mob stats depending on the level chosen when a mob spawns](https://github.com/project-topaz/topaz/blob/8985d2af68876367592222836db1f8e5b15c8175/src/map/entities/mobentity.cpp#L503-L528). Whether or not a mob is charmable isn't really a "stat" - it never changes from spawn-to-spawn. And as you pointed out, it occurs _after_ mobInitialize, and I feel we shouldn't need to deliberately use a different lua function to overwrite the charm mod simply because core's assignment is in the wrong place.

KnowOne134 (commented):
Looks good

ibm2431 (approved):

PR #425, opened 2020-03-14
Remove all trailing whitespace from src/ files.

mrhappyasthma:
This is a preemptive step for converting all tabs to spaces.

Commands used (on Mac):
```
  export LC_CTYPE=C
  export LANG=C
  find . -not \( -name .svn -prune -o -name .git -prune \) -type f -print0 | xargs -0 sed -i '' -E "s/[[:space:]]*$//"
```


ibm2431 (approved):
Thanks!

PR #426, opened 2020-03-14
Remove trailing whitespace from script/ files.

mrhappyasthma:
Commands used (on Mac):
```
  export LC_CTYPE=C
  export LANG=C
  find . -not \( -name .svn -prune -o -name .git -prune \) -type f -print0 | xargs -0 sed -i '' -E "s/[[:space:]]*$//"
```
The script  also adds a newline to the end of the files where one is missing. This isn't strictly needed, but it seems reasonable since it's the POSIX standard:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

> 3.206 Line
> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

This will also keep future diffs cleaner since some tools add the newline automatically.


ibm2431 (approved):
The line endings _was_ something that needed to be done!

GitHub limits display of files to the first 3,000 (per commit), so in the future, please try to break the number of files changed into multiple, smaller commits~! I pulled the commit locally and verified all the file changes were whitespace only, so this PR is fine, but review would have been a touch more difficult if these _weren't_ whitespace changes! 😅

Approved, thank you!

mrhappyasthma:
Good to know. This one was pretty straightforward so fortunately not too bad.

But I'll keep that in mind going forward for any bulk changes.

PR #427, opened 2020-03-15
Update russian comments to english for some src/ sub-directories.

mrhappyasthma:
Translated comments in:

- documentation/*
- src/common/*
- src/login/*
- src/search/*

NOTE: I also did some whitespace cleanups and removed empty comment blocks to fix up some of these files. However I will be doing separate pull requests to replace pre-existing tabs w/ spaces.

The only directories that still need translations are:

- src/map/*

**_I affirm:_**


src/search/search.cpp
```diff
@@ -971,8 +927,10 @@ search_req _HandleSearchRequest(CTCPRequestPacket& PTCPRequest)
     }
 
     return sr;
-    // не обрабатываем последние биты, что мешает в одну кучу например "/blacklist delete Name" и "/sea all Name"
+    // не обрабатываем последние биты, что мешает в одну кучу
```
mrhappyasthma, line 930: This was the only comment in these files that I couldn't quite understand enough to translate.

It's something about skipping the final bits, but by reading through the code it's not obvious why this is done or what's happening.

The phrase `кучу` here could possibly mean "heap", "struct", or "packet". Or perhaps something else...

kaincenteno:
saw several comment blocks that were empty, wouldn't it be better to fill them with information related to what is below? Maybe whomever wrote that meant to do that?

mrhappyasthma:
I could add them back, in the meantime, if we wanted to keep this PR better scoped, but I felt that they made the files easier to work with for the short term.

I agree that ideally we should comment all of the methods (even those without a placeholder comment block), but I felt it was outside the scope of this immediate change -- which was largely to change existing comments from Russian to English.

I'm also not the original author of the code, so it took me a while between Google Translate and tracing the code to make sure I ported the comments as accurately as possible. So it may take some time to add proper comments to all undocumented methods as well.

So my opinion is that adding documentation to undocumented code would be better scoped in a separate pull request.

src/search/search.cpp
ibm2431: They might be referring to extra information that's usually supposed to be in the same packet, but is being skipped because it's not always there.

I've probably booted the search server _once_, and I'm not finding anything useful in Windower's packet library, so I couldn't guess what that information might potentially be.

ibm2431 (approved):
Thanks! I appreciate the effort you're putting into trying to ensure the translations make sense with the surrounding code!

src/search/search.cpp
mrhappyasthma: What's your preference for this one? I was tempted to just leave it as is.

I could drop in a more literal translation of the text, but I worry if it's not accurate then it won't be meaningful.

mrhappyasthma:
No problem. I'll slowly go through the remaining files as well. :)

PR #430, opened 2020-03-15
[version update] 30200304_3

wrenffxi:
**_I affirm:_**


ibm2431 (approved):
Thanks as always, Wren! 👍 

Feel like updating `zone_settings` while you're here?

wrenffxi:
I don't know where the values come from, for entries into zone_settings and zone_weather.

TeoTwawki:
> Thanks as always, Wren! 
> 
> Feel like updating `zone_settings` while you're here?

I didn't add the flags or music values (presumably same as other WoE zones) but I did change the names and give them the default port in another PR just now

ibm2431:
That works for me, thanks guys!

PR #431, opened 2020-03-17
Claim updates

cocosolos:
This attempts to bring claim mechanics closer to retail. Claim is currently applied in several different places, so I moved all of that to one function in battleutils, I also added a separate function to dirty exp based on party size and player level, and another for handling new unclaiming/claim passing mechanics. I tried to get this as close to retail as I could, but there's probably still things to iron out, like all claimed mobs still having the same color name. Some things this will have an effect on is anything that uses the player on mob death. There will be situations where the mob dies while unclaimed and there is no owner, such as a DoT tick.

**_I affirm:_**

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -81,6 +81,7 @@ CMobEntity::CMobEntity()
     m_EcoSystem = SYSTEM_UNCLASSIFIED;
     m_Element = 0;
     m_HiPCLvl = 0;
+    m_HiPartySize = 0;
     m_THLvl = 0;
     m_ItemStolen = false;
 
@@ -505,6 +506,7 @@ void CMobEntity::Spawn()
     CBattleEntity::Spawn();
     m_giveExp = true;
     m_HiPCLvl = 0;
+    m_HiPartySize = 0;
     m_THLvl = 0;
     m_ItemStolen = false;
     m_DropItemTime = 1000;
@@ -666,18 +668,11 @@ void CMobEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
             this->PAI->EventHandler.triggerListener("WEAPONSKILL_USE", this, PTarget, PSkill->getID(), state.GetSpentTP(), &action);
             PTarget->PAI->EventHandler.triggerListener("WEAPONSKILL_TAKE", PTarget, this, PSkill->getID(), state.GetSpentTP(), &action);
         }
-
-        if (objtype == TYPE_PET && PMaster && PMaster->objtype == TYPE_PC )
+        if (PTarget->isDead())
         {
-            auto mob = dynamic_cast<CMobEntity *>(PTarget);
-            if (mob && !mob->CalledForHelp())
-            {
-                mob->m_OwnerID.id = PMaster->id;
-                mob->m_OwnerID.targid = PMaster->targid;
-                mob->updatemask |= UPDATE_STATUS; //This can go here because we only wanna call the updatemask if this happens
-            }
+            battleutils::ClaimMob(PTarget, this);
         }
-
+        battleutils::DirtyExp(PTarget, this);
         if (msg == 0)
         {
             msg = PSkill->getMsg();
@@ -732,6 +727,12 @@ void CMobEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
         }
         PTarget->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DETECTABLE);
     }
+    PTarget = static_cast<CBattleEntity*>(state.GetTarget());
+    if (PTarget->isDead())
+    {
+        battleutils::ClaimMob(PTarget, this);
+    }
+    battleutils::DirtyExp(PTarget, this);
 }
 
 void CMobEntity::DistributeRewards()
@@ -1025,6 +1026,7 @@ void CMobEntity::Die()
                 loc.zone->PushPacket(this, CHAR_INRANGE, new CMessageBasicPacket(this, this, 0, 0, MSGBASIC_FALLS_TO_GROUND));
 
             DistributeRewards();
+            m_OwnerID.clean();
         }
     }));
     if (PMaster && PMaster->PPet == this && PMaster->objtype == TYPE_PC)
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -704,9 +704,10 @@ namespace battleutils
             // Check for status effect proc. Todo: move to scripts soon™ after item additionalEffect refactor Teo is working on
             HandleSpikesStatusEffect(PAttacker, PDefender, Action);
 
-            if (PAttacker->objtype == TYPE_MOB && ((CMobEntity*)PAttacker)->m_HiPCLvl < PDefender->GetMLevel())
+            battleutils::DirtyExp(PAttacker, PDefender);
+            if (PAttacker->isDead())
             {
-                ((CMobEntity*)PAttacker)->m_HiPCLvl = PDefender->GetMLevel();
+                battleutils::ClaimMob(PAttacker, PDefender);
             }
             return true;
         }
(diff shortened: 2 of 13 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -4098,38 +4042,118 @@ namespace battleutils
...
+            {
+                uint8 pcinzone = 0;
+                PAttacker->ForAlliance([&pcinzone, &mob](CBattleEntity* PMember) {
+                    if (PMember->getZone() == mob->getZone() && distance(PMember->loc.p, mob->loc.p) < 100)
+                    {
+                        pcinzone++;
+                    }
+                });
+                mob->m_HiPartySize = std::max(pcinzone, mob->m_HiPartySize);
+                mob->m_HiPCLvl = std::max(PAttacker->GetMLevel(), mob->m_HiPCLvl);
```
ibm2431: Should the highest player level be updated for each member within range? From my reading, the highest level is determined by a call to `dirtyExp`, which seems to only happens on claiming actions - what about a high level WHM that's only throwing cures?

src/map/entities/mobentity.cpp
```diff
@@ -732,6 +727,12 @@ void CMobEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
         }
         PTarget->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DETECTABLE);
     }
+    PTarget = static_cast<CBattleEntity*>(state.GetTarget());
+    if (PTarget->isDead())
+    {
+        battleutils::ClaimMob(PTarget, this);
```
ibm2431, line 733: I realize the secondary targets shouldn't, but shouldn't the primary target of the AoE always get claimed?

(Same for magic too)

ibm2431 (commented):
Thanks cocosolos! A few questions:

How does/should this interact with switching targets in a party situation? From my reading (and please correct me if this is wrong), at the moment if a party member is the mob's owner, but then attacks another mob (spell, WS, etc), that original mob's owner (`attacker->PClaimedMob`) is wiped in `claimMob`, but `relinquishClaim` isn't called to attempt to pass on the claim to another party member.

src/map/utils/battleutils.cpp
TeoTwawki: exp changes should be decided by highest player character to take action on or otherwise effect a mob directly - debuffs, damage (inc damage from spikes), target of JA - but not from actions that have no effect on the mob itself such as curing people from outside their party.

uncertain about damaging undead via cure but I presume it is treated like a nuke on undead and as no effect when on non undead monsters

src/map/entities/mobentity.cpp
cocosolos: I believe I read that mobskills don't actually claim, only normal attacks/magic, but I may be wrong about that. Magic should be claiming the primary target.

cocosolos:
Switching targets doesn't call relinquishClaim, only actually disengaging from a mob. So if you're the owner and switch targets the mob will go unclaimed unless another party member hits it first.

src/map/utils/battleutils.cpp
ibm2431: Sorry, I should have been more clear; a high level WHM _in_ the party.

cocosolos: The highest level party member is checked in dirtyExp, and also in charutils::DistributeExperiencePoints. So If a high level WHM doesn't physically interact with a mob, they should still be considered when exp is actually split. [Starting around here.](https://github.com/project-topaz/topaz/blob/432058d0ae232b2f0341b1e379e0f9b4a9515eb8/src/map/utils/charutils.cpp#L3265)

Edit: [Sorry it's actually the block above that.](https://github.com/project-topaz/topaz/blob/432058d0ae232b2f0341b1e379e0f9b4a9515eb8/src/map/utils/charutils.cpp#L3245)

ibm2431: Right; but the highest player level is a comparison against the attacker and the currently highest set for the mob:
`mob->m_HiPCLvl = std::max(PAttacker->GetMLevel(), mob->m_HiPCLvl);`
`dirtyExp` does loop through the entire party when a claiming action has been taken, but only to count the number of players in the party; as written it's not keeping track of the highest level player:
```
PAttacker->ForAlliance([&pcinzone, &mob](CBattleEntity* PMember) {
    if (PMember->getZone() == mob->getZone() && distance(PMember->loc.p, mob->loc.p) < 100)
    {
        pcinzone++;
    }
});
```
`DistributeExperiencePoints` does do so, but that only factors the WHM in upon mob death if the WHM or buffer remains in the party.

ibm2431:
Re: Switching targets, a player who has claim attacking another mob calls a clean for the owner ID of the mob they originally had claim on:
```
if (PDefender->isAlive() && attacker->PClaimedMob && attacker->PClaimedMob != PDefender
&& attacker->PClaimedMob->isAlive() && attacker->PClaimedMob->m_OwnerID.id == attacker->id)
    { // unclaim any other living mobs owned by attacker
        attacker->PClaimedMob->m_OwnerID.clean();
        attacker->PClaimedMob->updatemask |= UPDATE_STATUS;
        attacker->PClaimedMob = nullptr;
```
Which, unlike in `RelinquishClaim`, doesn't attempt to find another party member to pass claim onto.

Scenario:
1. Members A and B have taken aggressive action against a mob
2. Member A was the last one to do so, their ID is currently the one set in m_OwnerID
3. Member A, without taking an action which would trigger a `ReliniquishClaim` call, attacks another mob

What happens to the claim which should pass onto Member B?

src/map/utils/battleutils.cpp
cocosolos: Is the WHM supposed to factor into the party level if they don't act on the mob directly and they're not in the party when the mob dies? It's easy to change but I was under the impression that a high level WHM could be in the party during a fight, drop party, and not affect exp other than party size, given they don't interact with the mob.

From wiki:
`Party level is the level of the highest level player or pet to have acted on the monster since it was at 100% health with no debilitating status effects.`

cocosolos:
In that scenario, the mob is supposed to go unclaimed. Unless Member B does an action to become the owner before Member A performs a claiming action on another mob, claim will drop. Member B will still have that mob as their ClaimedMob, but the mob loses its owner. This is supposed to happen.

src/map/utils/battleutils.cpp
ibm2431: I _assumed_ it's supposed to affect exp but should probably have someone confirm~

@project-topaz/researcher

59blargedy: just to clarify, the scenario you wish to test is:
player a lvl 50
player b lvl 75
in party together, engage mob, player b takes no action
player b drops party and either:
does nothing (what is xp?)
PLs (what is xp?)

cocosolos: Player A can be level 1, Player B can be 99, as long as there's a sufficient difference in level to impact exp. Then party up, have only Player A attack the mob, and have Player B drop party before the mob dies. What does the exp look like when the mob dies? Does anything change if Player B cures/buffs Player A before dropping party?

TeoTwawki: > but that only factors the WHM in upon mob death if the WHM or buffer remains in the party.

thats expected: on retail I've jumped in party power buffed ppl then left without effecting the exp. (haste n refresh) tho SE nerfed the buff durations)

59blargedy:
https://youtu.be/BWZDHvzbns4
perhaps changed at some point. tested: 
lvl 32 rng, lvl 99 rdm. 
lvl 32 engaged sole member of party, received outside cures, xp.
32 engaged, 99 buffed, 99 dropped, no xp.
32 engaged 99 buffed with flurry (in case multiple buffs/regen messing things up), 99 dropped, no xp
32 engaged 99 dropped, cured outside, no xp
32 engaged 99 dropped, no 99 action, no xp
(all on xp bearing mobs

KnowOne134 (approved):
been testing and seems to be good

ibm2431:
> In that scenario, the mob is supposed to go unclaimed.

Even if Member B is still engaged?

cocosolos:
> > In that scenario, the mob is supposed to go unclaimed.
> 
> Even if Member B is still engaged?

Yep, unless they're the owner they basically have no claim over the mob. This is also why Member A could drop party instead of attacking another mob and the mob would become exclusively theirs, Member B would auto disengage.

ibm2431:
What about a scenario of two BLMs who have both (single-target) nuked a mob, but then the one currently listed as the mob's owner decides to nuke a second?

I'm worried about people outside a party trying to yoink an NM if the one single party member who is marked as the claimant tries to deal with an add.

cocosolos:
If the owner nukes another mob, the first mob will go unclaimed. I believe this is intentional. However, anyone not in this party will have to place themselves at the top of the hate list to steal the claim. Any party member of the original owner will be able to reclaim instantly.

PR #433, opened 2020-03-20
Fix gil split with treasure chests

cocosolos:
table.remove was messing up index causing out of range errors,
also split gil with alliance and not just party.

**_I affirm:_**

### DIFF
--- a/scripts/globals/treasure.lua
+++ b/scripts/globals/treasure.lua
@@ -1475,10 +1475,8 @@ tpz.treasure.onTrade = function(player, npc, trade, chestType)
         local gilAmount = math.random(info.gil[2], info.gil[3])
         local gil = gilAmount/#membersInZone
         for i = 1, #membersInZone do
-            if player:getZoneID() == membersInZone[i]:getZoneID() then
-                membersInZone[i]:addGil(gil)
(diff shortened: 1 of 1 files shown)


scripts/globals/treasure.lua
```diff
@@ -1465,18 +1465,19 @@ tpz.treasure.onTrade = function(player, npc, trade, chestType)
...
         end
         local gilAmount = math.random(info.gil[2], info.gil[3])
-        local gil = gilAmount/#players
-        for i = 1, #players, 1 do
-            if player:getZoneID() == players[i]:getZoneID() then
-                players[i]:addGil(gil)
-                players[i]:messageSpecial(ID.text.GIL_OBTAINED, gil)
+        local gil = gilAmount/#membersInZone
+        for i = 1, #membersInZone do
+            if player:getZoneID() == membersInZone[i]:getZoneID() then
```
ibm2431: You've already verified these members were in the same zone, hence the name of the table. 😉

ibm2431 (changes requested):

scripts/globals/treasure.lua
cocosolos: True! Not sure why I left that there, fixed.

ibm2431 (approved):

PR #434, opened 2020-03-22
Mystery box additions

cocosolos:
Adds a function to load vectors full of obtainable item ids from each mystery box dial on map init. Adds a function to select a random item id from one of those vectors, using the dial number as an argument. Updated Arbitrix to use this function and give out items, and hold onto an item until collected by the player when inventory is full.

**_I affirm:_**


scripts/zones/Windurst_Walls/npcs/Arbitrix.lua
```diff
@@ -7,47 +7,210 @@
...
-        player:updateEvent(1, 1, (dailyTallyPoints >= 10) or 1)
-    elseif csid == 520 and option == 10 then
-        -- Deliver Dial 1
+    if csid == 520 then
+        switch (option): caseof
+        {
+            [4] = function() -- Peek items
+                player:updateEvent(player:selectDailyItem(6), player:selectDailyItem(6), player:selectDailyItem(6))
+            end,
+            [9] = function() -- Dial 1
```
ibm2431: Can shrink this option logic with the following:
```lua
if option == 4 then
    -- peek
else
    local dial = math.floor(option / 8)
    local option_type = option % 8
    local dial_used = 0
    local dial_cost = costs[dial]
    local dial_mask = false
    if dial >= 6 then
        dial_mask = dial - 6
        dial_used = player:getMaskBit(gobbieBoxUsed, dial_mask)
    end
    switch (option_type): caseof
        [1] = function()
            if dial_used then
                player:updateEvent(1, dial, 2) -- already used this dial
            elseif dailyTallyPoints >= dial_cost then
                itemid = player:selectDailyItem(dial)
                player:setCharVar("gobbieBoxHoldingItem", itemid)
                player:setCurrency("daily_tally", dailyTallyPoints - dial_cost)
                if dial_mask then
                    player:setMaskBit(gobbieBoxUsed, "gobbieBoxUsed", dial_mask, true)
                end
                player:updateEvent(itemid, dial, 0)
            else
                player:updateEvent(1, dial, 1) -- not enough points
            end
        end,
        [2] = function()
            if player:getFreeSlotsCount() == 0 then
                player:updateEvent(holdingItem, 0, 0, 1) -- inventory full, exit event
                player:messageSpecial(ID.text.ITEM_CANNOT_BE_OBTAINED + 2) -- generic "Cannot obtain the item."
            end
        end,
        [5] = function()
            if holdingItem > 0 and npcUtil.giveItem(player, holdingItem) then
                player:setCharVar("gobbieBoxHoldingItem", 0)
            end
            player:updateEvent(specialDialUsed, adoulinDialUsed, pictlogicaDialUsed, wantedDialUsed, 0, 0, hideOptionFlags, dailyTallyPoints)
        end,
    end
end
```

Will need a `costs` table somewhere. But in Arbitrix is okay - you don't need to globalize this just yet!

scripts/zones/Windurst_Walls/npcs/Arbitrix.lua
```diff
@@ -7,47 +7,210 @@
...
 end
 
 function onEventFinish(player, csid, option)
     if csid == 519 then
         player:setCurrency("daily_tally", 50)
+    elseif csid == 521 then
+        if player:getFreeSlotsCount() == 0 then
+            player:messageSpecial(ID.text.ITEM_CANNOT_BE_OBTAINED + 2) -- generic "Cannot obtain the item."
+        elseif npcUtil.giveItem(player, player:getCharVar("gobbieBoxHoldingItem")) then
+            player:setCharVar("gobbieBoxHoldingItem", 0)
```
ibm2431, line 112: `npcUtil.giveItem` will handle the slot count and displaying unable to obtain for you:
```lua
if not npcUtil.giveItem(player, player:getCharVar("gobbieBoxHoldingItem") then
    player:setCharVar("gobbieBoxHoldingItem", 0)
end
```

src/map/lua/lua_baseentity.cpp
```diff
@@ -3870,6 +3871,28 @@ inline int32 CLuaBaseEntity::breakLinkshell(lua_State* L)
...
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+    CCharEntity* PChar = ((CCharEntity*)m_PBaseEntity);
+    int selection = daily::SelectItem((uint8)lua_tointeger(L, 1));
+    while ((itemutils::GetItem(selection)->getFlag() & 0x8000) > 0 && charutils::HasItem(PChar, selection))
+    {
+        selection = daily::SelectItem((uint8)lua_tointeger(L, 1));
+    }
```
ibm2431: I'm leaning towards keeping all the logic for selection in a single place, so the binding just passes the Char and dial to `daily::` and returns the final result.

_Ideally_ we'd have the select daily util function be exposed to Lua instead of adding yet another binding to baseentity, but I don't feel strongly enough to _press_ that particular route. 😅

ibm2431 (changes requested):
Thanks for taking this on, cocosolos!

But you know me; always looking for how we can make stuff more concise~! 😉

src/map/lua/lua_baseentity.cpp
cocosolos: I actually started by putting the rerolling directly in daily::SelectItem, but it involved adding a loop for each dial case, which is why I moved it here.

scripts/zones/Windurst_Walls/npcs/Arbitrix.lua
cocosolos: npcUtil.giveItem actually gives a different message than this. "Unable to obtain \<item\>." vs "Unable to obtain the item."

scripts/zones/Windurst_Walls/npcs/Arbitrix.lua
cocosolos: This looks a lot better. I kinda just threw this together to demonstrate the item selection, but I'll make these edits and make sure everything works right.

scripts/zones/Windurst_Walls/npcs/Arbitrix.lua
ibm2431: Oh right, the mystery box doesn't say what the item is, does it? Nevermind then!

ibm2431 (approved):

PR #438, opened 2020-03-24
This *should* fix Guivre's ability to walk around

TeoTwawki:
**_I affirm:_**
- [x] that I've _tested my code_ since the last commit in the PR, and will test after any later commits ~~(need a volunteer to test, as I lack time - but I have other mobs using this same code with diff pos's working fine!)~~ tested by Zach

thought: should onPath be a local function?


kaincenteno:
ouch that's a huge array, i take npc path is still borken

TeoTwawki:
> ouch that's a huge array, i take npc path is still borken

yeah I seriously dislike the way these are handled but doubly so when there are so freakin many points in the array. I probably should have rounded these to nearest rather than truncating as well, but I want to know if he even moves now before messing with that more.

zach2good:
Confirmed, he wanders around as he should 👍

ibm2431:
Culled some intermediate points where only one axis was being changed
[culled.txt](https://github.com/project-topaz/topaz/files/4383940/culled.txt)

ibm2431 (approved):

PR #444, opened 2020-03-27
Fix new GCC warning in Trust branch

zach2good:
**_I affirm:_**

**Work:**
In hindsight, I shouldn't have been using `CCharEntity` stuff inside `CBattleEntity`. This moves `ForPartyWithTrusts()` into `CCharEntity` and blocks incorrect usage from the lua side with debug break.

**Tested:**
Summoning trusts, the only current use of this function.
Warning is gone 👍

### DIFF
--- a/src/map/entities/battleentity.h
+++ b/src/map/entities/battleentity.h
@@ -582,22 +582,6 @@ public:
     }
 
     template        <typename F, typename... Args>
-    void            ForPartyWithTrusts(F func, Args&&... args)
-    {
-        if (PParty) {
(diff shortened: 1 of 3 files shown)


ibm2431:
Instead of debug break, maybe push a nil? Then you can have liberal usage of `for pairs` wherever you want without fear of crashing the server~!

ibm2431:
Nevermind me, this is a template and isn't returning a table.

ibm2431 (approved):

Wiggo32 (approved):

PR #445, opened 2020-03-27
Move version.info

zach2good:
**_I affirm:_**

Purely selfishly, I'm fighting with this file all the time while I'm switching branches, rebasing, merging etc. So moving the default into `conf/default`, updating core to look for it in `conf` and adding an entry for it in the `.gitignore` for `conf`

Also, historically a file that people new to the project, or git, try to submit with their PRs


ibm2431 (approved):
I'd be lying if I said I hadn't been considering this for years.

Just need to make sure it's still updated!

TeoTwawki:
@ibm2431 I _want_ to say whatever reason I had for putting it where I did is no longer valid. But I don't remember.

TeoTwawki:
only thing that will be annoying about this is we'll have to manually update our file to reflect version updates.

TeoTwawki:
request: lets rename it to a proper conf since its going subfolder there now, and make sure the core defaults to allow all versions in the event the file isn't found. (I do not know the current default behavior, it was changed more than once since created)

ibm2431:
Good point on the rename.

Do we want to change the default ver_lock to 2? (In the file itself, also agreed with Teo for 2 default for instances where file isn't found)

TeoTwawki:
I only mean if the core doesn't see a file -at all- and not the default value in the file itself. A standard behavior for it like the other confs have.  The in file value (or even the default behavior for no file found) being 2 would make sense - already have warnings for players with a client newer than their server.

```
#0 - disabled (every version allowed)
#1 - enabled - strict (only exact CLIENT_VER allowed, default)
#2 - enabled - greater than or equal  (matching or greater than CLIENT_VER allowed)
VER_LOCK: 1
```
I just don't want file missing to be "1" because that will be annoying when I forget to copy it for quick testing a pr

ibm2431 (changes requested):
Change the default `VER_LOCK` in `version.conf` to 2 (matching the default option when the file isn't found), and I'll put a green `merge ready` back on this!

ibm2431 (approved):
Soon an era will end! 👍

PR #447, opened 2020-03-28
Fix pet claiming on initial hit

cocosolos:
Fixes various issues with pets and claiming. Applies initial enmity when pets attack so physical attacks now claim on the first hit, pet spells no longer claim, SMN bloodpacts now claim, other pet abilities no longer claim.

**_I affirm:_**


kaincenteno:
does this fix that weird claim issue with treasure caskets crashing servers when pets kill the mob?

cocosolos:
> does this fix that weird claim issue with treasure caskets crashing servers when pets kill the mob?

Is this a new issue? I thought that was resolved months ago?

KnowOne134 (approved):

ibm2431 (approved):

PR #449, opened 2020-03-29
Fix Trust party size counting

zach2good:
**_I affirm:_**

Just before I submitted my original Trust PR I jiggled the party and trust counting logic a bit and this got broken. Was picked up by @cocosolos.

Tested all my original scenarios, works as intended.


ibm2431 (approved):
I consider this an "unarguable" fix~

PR #450, opened 2020-03-29
Trust Movement, Positioning, Enmity Tracking & Combat Engagement

zach2good:
**_I affirm:_**

Thanks again to @Omnione for having a more reasonable solution for having trusts follow each other.

**Combat and Roaming Movement**
- Trusts spawn in a line and run in just a little bit
- Trusts roam as "follow the leader"
- Trusts cleanly form up in combat
- No more randomness
- No more strange lerping and jittering
- No more trusts fighting for position
- No more Trusts getting trapped behind one another
- No more Trusts getting trapped on top of each other and running round in circles... (even though that one was funny)

**Enmity**
There is a new binding for hasEnmity(), which checks all nearby mobs if an entity has ANY enmity. Used during trust summoning, probably shouldn't be used in battle (or often).

**Resting**
Trusts heal 3% HP and MP per tick after waiting 10 seconds after combat finishes. If your 3% doesn't go above 1hp/mp it will try 5%, after that you don't get anything.

**Engagement**
There was a PLastAttacker all along. I've done a good bit of testing and everything works. No strange edge cases, I'm sure those will come out in time.

**Job Latents and Party Jobs**
Summoned trusts are taken into account for job-related latents. Tested with Nanaa Mihgo with Forte Earring (Latent: +1 Dex with THF in party).
They are also now counted in hasPartyJob(), used in COR rolls.

Still not perfect, but at least 90% better than before.


src/map/latent_effect_container.cpp
```diff
@@ -813,6 +814,14 @@ bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect)
                         break;
                     }
                 }
+                for (auto trust : static_cast<CCharEntity*>(member)->PTrusts)
+                {
+                    if (trust->GetMJob() == latentEffect.GetConditionsValue())
```
ibm2431, line 819: I know you said you tested trusts activating latent effects, but was the test on Topaz or retail?

src/map/lua/lua_baseentity.cpp
```diff
@@ -10269,6 +10277,54 @@ inline int32 CLuaBaseEntity::updateClaim(lua_State *L)
     return 0;
 }
 
+/************************************************************************
+*  Function: hasEnmity()
```
ibm2431, line 10327: I'm a little surprised that there isn't already a struct keeping track of all the different hate lists a player is on (player->onEnmityLists = { *container a, *container b}).

"Surely we have to have something - cure would use it, right?"
[No, every cure checks the entire mob list too.](https://github.com/project-topaz/topaz/blob/release/src/map/utils/battleutils.cpp#L3461-L3466)

(A struct like that isn't a request, just me musing something we might consider in the future.)

src/map/lua/lua_baseentity.cpp
```diff
@@ -10269,6 +10277,54 @@ inline int32 CLuaBaseEntity::updateClaim(lua_State *L)
...
+        auto* mob = static_cast<CMobEntity*>(it->second);
+        auto* enmityList = mob->PEnmityContainer->GetEnmityList();
+        auto foundEnmityEntityResult = std::find_if(enmityList->begin(), enmityList->end(), [&](auto& obj)
+        {
+            return obj.second.PEnmityOwner->id == m_PBaseEntity->id;
+        });
+        if (foundEnmityEntityResult != enmityList->end())
+        {
+            hasEmity = true;
+            break;
```
ibm2431: `EnmityContainer` does have `HasID` though, so you can save yourself a little effort.
```cpp
CMobEntity* mob = static_cast<CMobEntity*>(it->second);
if (mob->PEnmityContainer->HasID(m_PBaseEntity->id))
{
    hasEnmity = true;
    break;
}
```

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -76,40 +76,64 @@ void CTrustController::DoCombatTick(time_point tick)
...
     if (POwner->PMaster->GetBattleTargetID() != POwner->GetBattleTargetID())
     {
         POwner->PAI->Internal_ChangeTarget(POwner->PMaster->GetBattleTargetID());
-        m_LastTopEntity = nullptr;
+        m_LastTopEnmity = nullptr;
     }
 
     float currentDistance = distance(POwner->loc.p, POwner->PMaster->loc.p);
     PTarget = POwner->GetBattleTarget();
+    uint8 currentPartyPos = GetPartyPosition();
```
ibm2431, line 91: Would it be better for a trust's party position to be set on spawn, and only recalculated when another trust is despawned?

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -118,23 +142,61 @@ void CTrustController::DoCombatTick(time_point tick)
...
     }
 
-    float currentDistance = distance(POwner->loc.p, POwner->PMaster->loc.p);
+    uint8 currentPartyPos = GetPartyPosition();
+    CBattleEntity* PFollowTarget = (GetPartyPosition() > 0) ? (CBattleEntity*)PMaster->PTrusts.at(currentPartyPos - 1) : POwner->PMaster;
+    float currentDistance = distance(POwner->loc.p, PFollowTarget->loc.p);
+
+    for (auto POtherTrust : PMaster->PTrusts)
+    {
+        if (POtherTrust != POwner && distance(POtherTrust->loc.p, POwner->loc.p) < 1.0f)
```
ibm2431: I'm wondering if it might be better to have declumping occur after distance checks so trusts aren't too preoccupied declumping from each other when they should pathing to the master.

(Or if kept here, add in a check here to make sure the _current_ trust isn't already following a path)

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -118,23 +142,61 @@ void CTrustController::DoCombatTick(time_point tick)
...
         }
         else if (POwner->GetSpeed() > 0)
         {
-            POwner->PAI->PathFind->WarpTo(POwner->PMaster->loc.p, RoamDistance);
+            POwner->PAI->PathFind->WarpTo(PFollowTarget->loc.p, RoamDistance);
+        }
+    }
+
+    if (POwner->CanRest() &&
+        m_Tick - m_CombatEndTime > 10s &&
```
ibm2431, line 187: A thought occurred that we might be able to cheat by setting a special regen status effect on disengage - which is cleared on engage - and then letting the status effect container handle it.

I don't know if it'd be _better_ than doing it yourself here; just bringing it up as an option in case someone else might say, "Yes, that'd be better because X and Y".

ibm2431 (commented):
Not really requesting changes (maybe the enmity container HasID if possible, so we're consistent) - just general thoughts and questions. Nothing here looks "off" to me~

src/map/latent_effect_container.cpp
zach2good: It was on Topaz, because I `!additem`'d myself a `Forte Earring`. I don't have confirmation that Trusts trigger Latents, but (the wikis say) that they affect COR Rolls, so that was my logical jump

src/map/lua/lua_baseentity.cpp
zach2good: Yeah, out of scope of this PR, but would be super useful

src/map/ai/controllers/trust_controller.cpp
zach2good: I shall test and see how it looks 👍

zach2good: All the different combinations look about the same, so I changed to use your suggestions

src/map/ai/controllers/trust_controller.cpp
zach2good: Probably, the trust could keep their own position, and the char could re-calculate when things change and tell each trust.

src/map/lua/lua_baseentity.cpp
TeoTwawki: @ibm2431 suggestion : should track the little weird things like that on a spreadsheet in a gist for the future so they don't get lost.

src/map/ai/controllers/trust_controller.cpp
zach2good: I think it would probably be better to do the sneaky regen, this system I'm using has the healing "ticks" not in line with regular healing ticks

zach2good: Hmmm, I tried it out with a regen/refresh status and I couldn't get it to work as cleanly as I'd like, I'll come back to this later if needed

ibm2431 (approved):
Looks good for now!

PR #451, opened 2020-03-29
mob UDMG fixes

wrenffxi:
**_I affirm:_**

Tinnin: should be immune to breath damage.

Dynamis Lord: should be immune to damage before initial Ying/Yang die, not take double damage. (Because Dynamis Lord is now force-popped, this code probably doesn't matter anymore)


ibm2431 (approved):
Thanks, Wren!

PR #452, opened 2020-03-30
Fixed Evisceration WS Quest

Solarsurge:
*Fixed server crash on NM spawn due to undefined text.
*Rearranged onTrigger order to prevent blocking Zilart/Windurst Mission progress until Evisceration quest is completed.

**_I affirm:_**


ibm2431 (approved):
Thanks, Solarsurge!

PR #453, opened 2020-03-31
Claim drop grace period

cocosolos:
This adds a short grace period before claim drops and also fixes an issue were rewards are not distributed when proper claim isn't achieved before death. This still isn't 100% as there are some situations where claim does drop instantly, needs more retail testing. Also moved the call to RelinquishClaim from OnDisengage to OnChangeTarget.

**_I affirm:_**


ibm2431 (approved):
Looks legit!

PR #458, opened 2020-03-31
FixedPromyvions

Solarsurge:
*ALL Promyvion cutscenes now properly working according to mission progress.
*Fixed player dropoff after interacting with Shattered Telepoints during 2nd Promyvion
*Map acquisition now works properly in Holla and Mea

**_I affirm:_**


scripts/zones/Promyvion-Mea/npcs/qm1.lua
```diff
@@ -0,0 +1,27 @@
...
+-----------------------------------
+
+function onTrigger(player, npc)
+    player:messageSpecial(ID.text.NOTHING_OUT_OF_ORDINARY)
+end
+
+function onTrade(player, npc, trade)
+    if npcUtil.tradeHas(trade, 1722) and not player:hasKeyItem(tpz.ki.MAP_OF_PROMYVION_MEA) then
+        npcUtil.giveKeyItem(player, tpz.ki.MAP_OF_PROMYVION_MEA)
+        player:confirmTrade()
```
ibm2431: I know you're just using what's currently in these files, but while we're here, we might as well fix a retail inaccuracy~!

These should be starting an event where the screen flashes and there's text about blinding white lights and images being burned into your mind. This is event 49 for all three promyvions.

(And then give the KI and confirm the trade in onEventFinish for those events)

scripts/zones/Promyvion-Mea/npcs/qm1.lua
```diff
@@ -0,0 +1,27 @@
...
+    else
+        player:messageSpecial(ID.text.NOTHING_HAPPENS)
```
ibm2431, line 19: Does this text display when you trade a bad item? Bad trades are usually just ignored.

ibm2431 (changes requested):
You got close to a no-comment approval, but were sabotaged by previous scripters! 😉

scripts/zones/Promyvion-Mea/npcs/qm1.lua
Solarsurge: Fixed

scripts/zones/Promyvion-Mea/npcs/qm1.lua
Solarsurge: "Nothing happens..." is displayed on a bad trade. To my knowledge, that's correct.

ibm2431 (approved):
Thanks Solar!

scripts/zones/Promyvion-Dem/npcs/qm1.lua
```diff
@@ -24,4 +25,7 @@ function onEventUpdate(player, csid, option)
 end
 
 function onEventFinish(player, csid, option)
+    if csid == 49 then
+        npcUtil.giveKeyItem(player, tpz.ki.MAP_OF_PROMYVION_DEM)
```
TeoTwawki, line 29: Confirm trade should also be at event finish, not before it starts

PR #463, opened 2020-04-03
Make caskets stop trying to disappear players

ibm2431:
`checkItemChestIsEmpty` [only takes an argument of `npc`; after dealing with some not-lua-local vars, it then calls `removeChest` on this first argument](https://github.com/project-topaz/topaz/blob/release/scripts/globals/caskets.lua#L421-L428).

[At which point the chest tries to disappear the player.](https://github.com/project-topaz/topaz/blob/release/scripts/globals/caskets.lua#L338)

To be fair, it does technically succeed in this endeavor. Unfortunately, it disappears the rest of the server in addition to the intended victim.

**_I affirm:_**

### DIFF
--- a/scripts/globals/caskets.lua
+++ b/scripts/globals/caskets.lua
@@ -723,7 +723,7 @@ local function giveItem(player, npc, itemNum)
                     if player:addItem(itemID, 33) then
                         messageChest(player, "PLAYER_OBTAINS_ITEM", itemID, 0, 0, 0)
                         npc:setLocalVar(itemQuery, 0)
-                        checkItemChestIsEmpty(player, npc)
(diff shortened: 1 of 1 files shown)


Wiggo32 (approved):

ibm2431:
I've kicked Travis twice now and it's still acting up. AppVeyor checks out, so I am utilizing my almighty write-access to ignore Travis this time.

TeoTwawki:
> Unfortunately, it disappears the rest of the server in addition to the intended victim.

Casket: and for my next trick, ima pull a hat out of this rabbit!

Rabbit: _readies whirls claws_

PR #464, opened 2020-04-03
Add Elemental Sforzo placeholder, to make the ability usable.

### LINKED ISSUE #429: Adding Elemental Sforzo to ability list
**_I have:_**

Elemental Sforzo is already coded in the status effect list and is available when you use godmode, but the ability itself is not coded as an ability for runefencer to use natively. 

-- Ability: Elemental Sforzo
-- Grants immunity to all magical attacks.
-- Obtained: Runefencer Level 1
-- Recast Time: 1:00:00
-- Duration: 0:00:30
require("scripts/globals/status")

function onAbilityCheck(player,target,ability)
    return 0,0
end

function onUseAbility(player,target,ability)
    player:addStatusEffect(tpz.effect.ELEMENTAL_SFORZO,1,0,30)
end

mrhappyasthma:
A few things are still needed:

1. Add enmity when cast.
2. Update the status_effect to also protect against status effect
spells that cause debuffs.

**_I affirm:_**


TeoTwawki:
>spell castable

*ability usable

:)

scripts/globals/abilities/elemental_sforzo.lua
```diff
@@ -0,0 +1,20 @@
...
+-----------------------------------
+require("scripts/globals/status")
+-----------------------------------
+
+function onAbilityCheck(player,target,ability)
+  return 0,0
+end
+
+function onUseAbility(player,target,ability)
+  -- TODO: Add enmity when cast.
```
TeoTwawki: there are fields in the db table for the CE and VE, if the amounts are known

mrhappyasthma: Could you clarify what CE and VE stand for?

TeoTwawki: Cumulative Enmity and Volatile Enmity. One of them decays over time, the other does not.

https://www.bg-wiki.com/index.php?curid=11855#Cumulative_Enmity_Loss

https://github.com/project-topaz/topaz/blob/release/sql/abilities.sql#L28

mrhappyasthma: Looks like those values aren't populated: https://github.com/project-topaz/topaz/blob/release/sql/abilities.sql#L293

It's listed as `1800` CE and `7200` VE here: 
https://www.bg-wiki.com/bg/Elemental_Sforzo

TeoTwawki: That's why I was telling you where to populate them.

ibm2431 (changes requested):
Thanks, mrhappyasthma!

Could you add the CE and VE to the abilities.sql?

Indenting should be 4 spaces; the indents here are two~

mrhappyasthma:
Whoops. Force of habit. Will update the spacing.

Is it sufficient to just add the CE/VE to the sql file? Or do I need to apply those enmity values somewhere in code?

It seems like they will get picked up in the core and added to the `spell`. But I couldn't seem to find where it's applied to the player (if that's done automatically).

Any pointers would be helpful :) thanks!

TeoTwawki:
> Is it sufficient to just add the CE/VE to the sql file? Or do I need to apply those enmity values somewhere in code?

Yep just plug into the table, core does the rest on [ability ](https://github.com/project-topaz/topaz/blob/8985d2af68876367592222836db1f8e5b15c8175/src/map/ai/states/ability_state.cpp#L76)use.

scripts/globals/abilities/elemental_sforzo.lua
mrhappyasthma: Done. Please take another look.

mrhappyasthma:
Updated indentation and added the CE/VE values.

Thanks for the code pointers.

ibm2431 (approved):
Thanks, mrhappyasthma!

PR #465, opened 2020-04-03
Upgrade !hp, !mp and !tp commands

zach2good:
**_I affirm:_**

I've been using this modification to these commands for a couple of weeks now and I couldn't do without them. Maybe they could be useful to others too.

### DIFF
--- a/scripts/commands/hp.lua
+++ b/scripts/commands/hp.lua
@@ -27,16 +27,16 @@ function onTrigger(player, hp, target)
     -- validate target
     local targ
     local cursor_target = player:getCursorTarget()
-    if cursor_target ~= nil then
-        targ = cursor_target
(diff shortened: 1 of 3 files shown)


ibm2431 (commented):
Should the entered name take priority over the cursor target? I can envision someone becoming frustrated trying to get a player's stats but it failing because they were targeting something else.

TeoTwawki:
> Should the entered name take priority over the cursor target? I can envision someone becoming frustrated trying to get a player's stats but it failing because they were targeting something else.

I also expect I'll accidentally heal mobs I didn't mean to.

ibm2431 (approved):
Reordered the if a little so that the typed-in name takes priority over the current cursor target. Tested locally.

Approved. Thanks, zach!

PR #466, opened 2020-04-04
Trust Spell and Skill List SQL Reservation (and Mob Pools)

zach2good:
**_I affirm:_**

As mentioned in discord, trying to dump all the sql straight into release so it doesn't get broken up later


ibm2431 (approved):
I started on a review, but then realized it'd be easier on both of us if I just pushed this commit myself~

Approved!

Wiggo32 (approved):

PR #470, opened 2020-04-05
Client version update 30200327_2

wrenffxi:
**_I affirm:_**

Nothing big in this one.  Just a couple small ID shifts.


ibm2431 (approved):
Thanks, Wren!

PR #471, opened 2020-04-05
Fix issues with multi phase bcnms

cocosolos:
Fixes issue with battlefield being marked as won even when timing out
or wiping. Also fixed some issues in airship fight.

**_I affirm:_**

### DIFF
--- a/scripts/zones/QuBia_Arena/mobs/Warlord_Rojgnoj.lua
+++ b/scripts/zones/QuBia_Arena/mobs/Warlord_Rojgnoj.lua
@@ -7,8 +7,7 @@ mixins = {require("scripts/mixins/job_special")}
 require("scripts/globals/status")
 local ID = require("scripts/zones/QuBia_Arena/IDs")
 
-function allHeirMobsDead(mob)
-    local battlefield = mob:getBattlefield()
+function phaseChangeReady(battlefield)
     local inst = battlefield:getArea()
     local instOffset = ID.mob.HEIR_TO_THE_LIGHT_OFFSET + (14 * (inst-1))
     for i = instOffset + 3, instOffset + 13 do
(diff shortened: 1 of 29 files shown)


scripts/zones/QuBia_Arena/mobs/Warlord_Rojgnoj.lua
```diff
@@ -7,16 +7,28 @@ mixins = {require("scripts/mixins/job_special")}
...
-function onMobDeath(mob, player, isKiller)
-    local inst = mob:getBattlefield():getArea()
-    local instOffset = ID.mob.HEIR_TO_THE_LIGHT_OFFSET + (14 * (inst-1))
-    if GetMobByID(instOffset + 0):isDead() and GetMobByID(instOffset + 1):isDead()
-        and GetMobByID(instOffset + 2):isDead()
-    then
-        mob:getBattlefield():setLocalVar("loot", 0)
+function onMobSpawn(mob)
+    local battlefield = mob:getBattlefield()
+    if allHeirMobsDead(mob) then
```
ibm2431: 1) Maybe rename this function to "phaseOneDead" or "phaseOneComplete"

2) Can pass in the battlefield directly into the function instead of the mob; the original mob isn't used once inside it

ibm2431 (changes requested):

scripts/zones/QuBia_Arena/mobs/Warlord_Rojgnoj.lua
TeoTwawki: `phaseChangeReady()` ?

cocosolos: True to both points. I just copied this function from the phase 1 mobs, should I change it there too?

ibm2431: Yeah, "allHeirMobsDead" is even more confusing inside the phase 1 mob files~! 😅

ibm2431 (approved):
Thanks, cocosolos!

PR #474, opened 2020-04-08
Hopper

hooksta4:
**_I affirm:_**


scripts/zones/Mamook/npcs/Toad_s_footprint1.lua
```diff
@@ -0,0 +1,38 @@
...
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player, npc)
+    local princeandhopper = player:getCharVar("princeandhopper")
+    if  princeandhopper == 3 then
+        player:startEvent(223)
+    elseif  princeandhopper == 4 then
+        player:messageSpecial(ID.text.DRAWS_NEAR)
+        npcUtil.popFromQM(player, npc, ID.mob.POROGGO_CASANOVA, {hide = 30})
```
cocosolos: You might want this one to spawn last. I think with the recent claim changes only the last one spawned will stay claimed after about 3 seconds unless engaged.

hooksta4:
I just realized I will need to update the footprint to only give the CS and param for those current on that stage of the fight.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Maudaal.lua
```diff
@@ -3,16 +3,33 @@
...
 function onEventFinish(player,csid,option)
-end
+    if (csid == 889 and option == 1) then
+        return
+    elseif (csid ==889 and option ==2) then
+        player:addQuest(AHT_URHGAN,tpz.quest.id.ahtUrhgan.THE_PRINCE_AND_THE_HOPPER)
+        player:setCharVar("princeandhopper",1)
+    elseif (csid ==890) then
+        player:addItem(16270)
+        player:messageSpecial(ID.text.ITEM_OBTAINED,16270)
```
cocosolos: Probably wanna use `npcUtil.giveItem(player, 16270)` here as a condition for finishing the quest, does the same thing and checks inventory space, returns true/false.

cocosolos:
> I just realized I will need to update the footprint to only give the CS and param for those current on that stage of the fight.

It looks like you're already doing that with the charvar

scripts/zones/Mamook/npcs/Toad_s_footprint1.lua
hooksta4: So just switch casa to the bottom of the list eh?

cocosolos: Probably, that one looks like the main target and if any of them should have their claim stick it'd be that one.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Maudaal.lua
```diff
@@ -3,16 +3,33 @@
...
+    if (csid == 889 and option == 1) then
+        return
+    elseif (csid ==889 and option ==2) then
+        player:addQuest(AHT_URHGAN,tpz.quest.id.ahtUrhgan.THE_PRINCE_AND_THE_HOPPER)
+        player:setCharVar("princeandhopper",1)
+    elseif (csid ==890) then
+        npcUtil.giveItem(player, 16270)
+        player:setCharVar("princeandhopper",0)
+        player:completeQuest(AHT_URHGAN,tpz.quest.id.ahtUrhgan.THE_PRINCE_AND_THE_HOPPER)
+    end
```
cocosolos:
```lua
    elseif (csid ==890) then
        if npcUtil.giveItem(player, 16270) then
            player:setCharVar("princeandhopper",0)
            player:completeQuest(AHT_URHGAN,tpz.quest.id.ahtUrhgan.THE_PRINCE_AND_THE_HOPPER)
        end
    end
```
Doing it like this will ensure they get their item before marking as complete. If it fails to give them the item they have to make room and watch the cs again.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Maudaal.lua
hooksta4: I somehow fucked this all up I need a masterclass on github for real.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Maudaal.lua
wrenffxi: Or:
```
elseif csid == 890 then
    npcUtil.completeQuest(
        player,
        AHT_URHGAN,
        tpz.quest.id.ahtUrhgan.THE_PRINCE_AND_THE_HOPPER,
        {item = 16270, var = "princeandhopper"}
    )
end
```

npcUtil.completeQuest handles:
awarding items, keyitems, gil, XP, title, and fame
setting quest-related variables to 0
completing the quest

hooksta4: I guess the best question is what is going to be the standard moving forward?

cocosolos: That's much better, I didn't even realize we had that.

ibm2431 (approved):
After talking with hooksta4, I did a little bit of PR work to fix the history, and then did a follow-up commit with changes I wanted to see. Nothing else **hop**s out at me~

Approved!

PR #477, opened 2020-04-08
QoL for Moongate

kaincenteno:
Moongate now open from 18:00 to 6:00 on full moon.
Removed semicolons at the end, and removed parenthesis on if statements.

**_I affirm:_**

### DIFF
--- a/scripts/zones/RoMaeve/Zone.lua
+++ b/scripts/zones/RoMaeve/Zone.lua
@@ -3,60 +3,55 @@
 -- Zone: RoMaeve (122)
 --
 -----------------------------------
-local ID = require("scripts/zones/RoMaeve/IDs");
-require("scripts/globals/conquest");
-require("scripts/globals/missions");
-require("scripts/globals/npc_util");
-require("scripts/globals/weather");
-require("scripts/globals/status");
+local ID = require("scripts/zones/RoMaeve/IDs")
+require("scripts/globals/conquest")
+require("scripts/globals/missions")
+require("scripts/globals/npc_util")
+require("scripts/globals/weather")
+require("scripts/globals/status")
 -----------------------------------
 
 function onInitialize(zone)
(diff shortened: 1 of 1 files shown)


scripts/zones/RoMaeve/Zone.lua
```diff
@@ -3,60 +3,55 @@
...
-    if (weather ~= tpz.weather.SUNSHINE and GetNPCByID(ID.npc.MOONGATE_OFFSET):getAnimation() ~= tpz.anim.CLOSE_DOOR) then
+    if IsMoonFull() and hour == 18 then
         for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET+7 do
-            GetNPCByID(i):setAnimation(tpz.anim.CLOSE_DOOR);
+            GetNPCByID(i):openDoor(1728)
+            GetNPCByID(i):untargetable(true)
         end
-    elseif (weather == tpz.weather.SUNSHINE and IsMoonFull() == true and VanadielHour() < 3) then -- reactivate things for the remainder of the time until 3AM
-        local moonMinRemaining = math.floor(432 * (180 - VanadielHour() * 60 + VanadielMinute())/180) -- 180 minutes (ie 3AM) subtract the time that has passed since midnight
+    elseif IsMoonFull() and hour == 6 then
```
ibm2431: Will the moon still be "full" after the final night? We wouldn't want to accidentally leave the doors untargetable. 😉

kaincenteno: change to
    elseif IsMoonFull() then
as
    elseif hour == 6 and GetNPCByID(ID.npc.MOONGATE_OFFSET):untargetable() then
didnt work for some reason even tho it looks legit (i just didnt want to be valid on every single vanaday).

ibm2431: We might be able to set up a timer when we _open_ the door, but it's probably not worth it - setting it to false once a real-life hour isn't expensive (and might actually be cheaper than adding an additional timer).

ibm2431 (approved):
Thanks, kain~

ghost (approved):

PR #482, opened 2020-04-09
Add family column to spell_list

zach2good:
**_I affirm:_**

I'm going to need this for the extensions to the MobSpellContainer I'm doing for Trust spellcasting, allows specifying a family of spells so you can pick the best out of them (Fire I vs Fire II) etc.

**Useful commands:**
```
SELECT spellid, name, family FROM tpzdb.spell_list;
```
```
SELECT name, family FROM tpzdb.spell_list Where family > 0;
```
```
"C:\Program Files\MySQL\MySQL Server 5.7\bin\mysqldump" --hex-blob --extended-insert=FALSE tpzdb -h localhost -u root -proot spell_list > spell_list.sql
```


ibm2431 (approved):
Yep, looks like SQL!

Only thoughts kicking around: Should currently un-family'd things of the same type (teleports, bar- spells, -na spells) be given the same family? They're technically not tiers of each other, so it may confuse the purpose of the column - just thinking of how I used to define gearswap set tables (ie: there are times when I want to quickly check if a spell is a -na spell). I can't say off the top of my head of how that would be a use for us, but there _might_ be something?

ibm2431:
Others: Results of a regex on `spell_list.sql` to put spell ID, name, and family into a table, and then running a lua script that spits out a reverse-lookup of family names.

As of ad80d14:
[family_results.txt](https://github.com/project-topaz/topaz/files/4482140/family_results.txt)
(Empty lines are from me breaking up groups while reviewing this output.)

ibm2431 (approved):
Approved!

Spoke with zach about this: spells which are currently commented out don't have families. They'll need to be added when they're uncommented; assuming I fail to get around to a follow-up commit to this branch after this PR is merged.

PR #483, opened 2020-04-09
Check if enmity list is empty for unclaimed mobs

cocosolos:
Prior to the recent changes the attack round would simply set the owner instead of checking enmity since without unclaiming mechanics it didn't really matter. Attack enmity is set in takePhysicalDamage, so missed attacks and 0 damage attacks didn't generate any enmity, BUT since the owner was getting set either way, during the mob roam tick it checks if it has anyone on their enmity list, and if it doesn't but it does have an owner, it adds base enmity there. This fixes an oversight in the claim code for when claim is attempted with an empty enmity list.

**_I affirm:_**

### DIFF
--- a/src/map/ai/states/mobskill_state.cpp
+++ b/src/map/ai/states/mobskill_state.cpp
@@ -104,7 +104,7 @@ bool CMobSkillState::Update(time_point tick)
     if (IsCompleted() && tick > m_finishTime)
     {
         auto PTarget = GetTarget();
-        if (PTarget && PTarget->objtype == TYPE_MOB)
+        if (PTarget && PTarget->objtype == TYPE_MOB && PTarget != m_PEntity && m_PEntity->allegiance == ALLEGIANCE_PLAYER)
         {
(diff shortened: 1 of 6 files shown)


src/map/ai/states/mobskill_state.cpp
```diff
@@ -102,6 +103,11 @@ bool CMobSkillState::Update(time_point tick)
     }
     if (IsCompleted() && tick > m_finishTime)
     {
+        auto PTarget = GetTarget();
+        if (PTarget && PTarget->objtype == TYPE_MOB)
+        {
+            static_cast<CMobEntity*>(PTarget)->PEnmityContainer->UpdateEnmity(m_PEntity, 0, 0, true);
```
ibm2431: Will we run into any surprises if mobs of the same allegiance use buffing mobskills on each other (or a goblin drops his bomb near his buds)?

ibm2431 (commented):
Helps if I press the "finish your review" button!

ibm2431 (approved):
Thanks as always, cocosolos!

ibm2431 (approved):

PR #486, opened 2020-04-11
Add geomancer unlock quest 'Dances with Luopans'.

mrhappyasthma:
The only TODO left here is to handle the cutscene 37
which happens after you have geomancer unlocked. It
allows the user to re-buy the Matre Bell. It needs
more research to determine how to handle the
`player:updateEvent(....)` during `onEventUpdate`.

Also I only implemented one of the /heal areas. I
found conflicting documentation, but seems like
you can rest at any of the Ergon Locus locations.
However I only implemented the one commonly used
in the unlock quest and mentioned in most of the
online wiki guides.

**_I affirm:_**


mrhappyasthma:
I wasn't sure the best way to handle the resting. I feel a bit icky about pulling quest info into `scripts/globals/effects/healing.lua`, but it seemed to be the only place I could find to track the user while they are using `/heal`.

I'm open to other suggestions.

mrhappyasthma:
Made a few tweaks and re-tested everything.

scripts/globals/effects/healing.lua
```diff
@@ -4,18 +4,32 @@
...
+local CeizakBattlegroundsID = require("scripts/zones/Ceizak_Battlegrounds/IDs")
 -----------------------------------
 
 function onEffectGain(target,effect)
     target:setAnimation(33)
+
+    -- Dances with Luopans
+    if target:getQuestStatus(ADOULIN, tpz.quest.id.adoulin.DANCES_WITH_LUOPANS) == QUEST_ACCEPTED and target:hasKeyItem(tpz.ki.LUOPAN) then
+        -- Ceizak Battlegrounds near Ergon Locus at K-10
+        if target:getZoneID() == 261 and target:getPlayerRegionInZone() == 1 then
```
ibm2431: Can use `tpz.zone.CEIZAK_BATTLEGROUNDS` instead of 261, prevents more "magic numbers" in the codebase~

In my capture of the quest I specifically aborted this one early to go test the other one in the zone. You can see my capture for confirmation that the other locus works and maybe get range/position data.

scripts/globals/effects/healing.lua
```diff
@@ -4,18 +4,32 @@
 -- Activated through the /heal command
 --
 -----------------------------------
-require("scripts/globals/status")
+require("scripts/globals/keyitems")
+require("scripts/globals/quests")
 require("scripts/globals/settings")
+require("scripts/globals/status")
+local CeizakBattlegroundsID = require("scripts/zones/Ceizak_Battlegrounds/IDs")
```
ibm2431: We don't need to require the IDs up here on load: instead down below, inside the successful quest status check:

```lua
if target:getCharVar("GEO_DWL_Luopan") == 0 then
    local ID = zones[target:getZoneID()]
    target:messageSpecial(ID.text.GEO_DWL_ARCANE_ENERGY)
end
```

scripts/globals/effects/healing.lua
```diff
@@ -4,18 +4,32 @@
...
     target:setAnimation(33)
+
+    -- Dances with Luopans
+    if target:getQuestStatus(ADOULIN, tpz.quest.id.adoulin.DANCES_WITH_LUOPANS) == QUEST_ACCEPTED and target:hasKeyItem(tpz.ki.LUOPAN) then
+        -- Ceizak Battlegrounds near Ergon Locus at K-10
+        if target:getZoneID() == 261 and target:getPlayerRegionInZone() == 1 then
+           if target:getCharVar("GEO_DWL_Luopan") == 0 then
+                target:messageSpecial(CeizakBattlegroundsID.text.GEO_DWL_ARCANE_ENERGY)  -- The arcane energies begin to course within your veins.
+                ticksRequired = math.random(2, 30)
+                target:setCharVar("GEO_DWL_Resting", ticksRequired)
```
ibm2431: `setCharVar` saves to SQL so the variable is kept between zones and logging out. Primary purpose is for things like quest status.

For volatile things we don't need to be able to save forever, there's `localVar` (not to be confused with with a lua `local` var - it's confusing I know, and we need to come up with a better name for these type of vars).
`player:setLocalVar("some_name", int_value)`
`player:getLocalVar("some_name", int_value)`
These are stored in memory, and let us be more gentle with the SQL server.

scripts/globals/effects/healing.lua
```diff
@@ -32,9 +46,17 @@ function onEffectTick(target,effect)
             target:addMP(12 + ((healtime - 2) * (1 + target:getMod(tpz.mod.CLEAR_MIND))) + target:getMod(tpz.mod.MPHEAL))
         end
     end
+
+    -- Dances with Luopans
+    if target:getCharVar("GEO_DWL_Resting") > 0 and healtime > target:getCharVar("GEO_DWL_Resting") then
+        target:messageSpecial(CeizakBattlegroundsID.text.GEO_DWL_WARMTH)  -- You feel a mystical warmth welling up inside you!
+        target:setCharVar("GEO_DWL_Resting", 0)
+        target:setCharVar("GEO_DWL_Luopan", 1)
+    end
```
ibm2431: We can get around doing any checks in the ticks with a timer:

(Disclaimer: I haven't done a timer recently, so my syntax may not be exact; I also don't remember the rules on what args need to be passed into the lambda)

```lua
-- onEffectGain:
if target:getCharVar("GEO_DWL_Luopan") == 0 then
    local timeRequired = math.ceil(some_num_ticks * 3)
    target:setLocalVar("GEO_DWL_Resting", os.time() + timeRequired)
    target:timer(timeRequired * 1000, function()
        local finishTime = target:getLocalVar("GEO_DWL_Resting")
        if finishTime > 0 and os.time() >= finishTime then
            target:messageSpecial(ID.text.GEO_DWL_WARMTH)
            target:setCharVar("GEO_DWL_Luopan", 1)
        end
    end)
end

-- onEffectLose
target:setLocalVar("GEO_DWL_Resting", 0)

File: scripts/zones/Western_Adoulin/npcs/Sylvie.lua [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+require("scripts/globals/keyitems")
+require("scripts/globals/npc_util")
+require("scripts/globals/quests")
+require("scripts/globals/status")
+local ID = require("scripts/zones/Western_Adoulin/IDs")
+
+function onTrade(player,npc,trade)
+    -- DANCES WITH LUOPANS
+    if player:getQuestStatus(ADOULIN, tpz.quest.id.adoulin.DANCES_WITH_LUOPANS) == QUEST_ACCEPTED then
+        if player:hasKeyItem(tpz.ki.FISTFUL_OF_HOMELAND_SOIL) and npcUtil.tradeHas(trade, 703, true) then -- Petrified Log
```
ibm2431: `true` triggers `hasExactly` - Sylvie will accept the trade even when other junk items are included (see my capture), so you _don't_ want to use `hasExactly` in this case.

(But `npcUtil.tradeHasExactly` will be what we want to use for AF armor forging later, if I recall correctly)

scripts/zones/Western_Adoulin/npcs/Sylvie.lua
```diff
@@ -0,0 +1,73 @@
...
+      player:confirmTrade()
+      player:delKeyItem(tpz.ki.FISTFUL_OF_HOMELAND_SOIL)
+      npcUtil.giveKeyItem(player, tpz.ki.LUOPAN)
+  elseif csid == 36 then
+      if npcUtil.giveItem(player, { 6074, 21460 }) then -- 'plate of Indi-Poison' and 'Matre Bell'
+        player:unlockJob(tpz.job.GEO)
+        player:messageSpecial(ID.text.YOU_CAN_NOW_BECOME, 0)  -- You can now become a geomancer!
+        npcUtil.giveKeyItem(player, tpz.ki.JOB_GESTURE_GEOMANCER)
+        player:setCharVar("GEO_DWL_Luopan", 0)
+        player:completeQuest(ADOULIN, tpz.quest.id.adoulin.DANCES_WITH_LUOPANS)
```
ibm2431: Indenting a couple spaces off in this block (but the logic looks great)

ibm2431 (changes requested):
Thanks mrhappyasthma! I'm happy to see you're making good use of npcUtil! 😃 

Couple requests below, including what I think _might_ be a solution for solving the healing message problem without needing to check all players' healing ticks!

scripts/globals/effects/healing.lua
TeoTwawki: https://github.com/project-topaz/topaz/wiki/Understanding-variables:--a-brief-guide

/shamelessplug

scripts/globals/effects/healing.lua
TeoTwawki: @Omnione didn't you say you tried this with a registered region, how'd that go?

Omnione: Yep i got it working perfectly, ill DM over the files

scripts/zones/Western_Adoulin/npcs/Sylvie.lua
mrhappyasthma: Ah, good to know. The capture I found on discord didn't demonstrate this.

scripts/zones/Western_Adoulin/npcs/Sylvie.lua
mrhappyasthma: Whoops. Fixed.

scripts/globals/effects/healing.lua
mrhappyasthma: Oh, good to know. Never saw this. 

Definitely will use it from now on, when appropriate.

scripts/globals/effects/healing.lua
mrhappyasthma: I'm always favorable of removing hardcoded values. Thanks for the tip!

Apparently 3 luopans work, according to: https://www.bg-wiki.com/bg/Dances_with_Luopans

But I couldn't find video capture of them.

Can you message me your capture on discord? I found one capture on there, but it didn't cover everything I would have preferred. Compared to the RUN capture, it was very minimalist.

scripts/globals/effects/healing.lua
mrhappyasthma: Nice trick. Will use this from now on. Really useful.

mrhappyasthma:
Resolved a handful of comments. Will look into applying the changes sent over by @Omnione next.

mrhappyasthma:
Applied the timer idea and added the other 2 areas that you can rest within.

Did some thorough testing and the timer idea works great.

The only thing I wish we had was the ability to keep a reference to the timer and cancel it immediately if the rest state ends. As of now it continues and just acts like a no-op on execution in that case.

But the core doesn't seem to support this and I didn't have time to look into adding that for a small optimization.

scripts/globals/effects/healing.lua
mrhappyasthma: Applied the timer and rewrote the logic based off of some of files.

scripts/globals/effects/healing.lua
mrhappyasthma: The capture had all 3 areas. Added and tested them all.

scripts/globals/effects/healing.lua
```diff
@@ -4,12 +4,36 @@
...
 require("scripts/globals/settings")
+require("scripts/globals/status")
+require("scripts/globals/zone")
 -----------------------------------
 
 function onEffectGain(target,effect)
     target:setAnimation(33)
+
+    -- Dances with Luopans
+    if target:getLocalVar("GEO_DWL_Locus_Area") == 1 and not target:setCharVar("GEO_DWL_Luopan", 1) then
```
ibm2431: Probably meant a `not (target:getCharVar("GEO_DWL_Luopan") == 1)` here instead of `not target:setCharVar("GEO_DWL_Luopan", 1)` 😉

mrhappyasthma: Oh my gosh. Good catch!

I'm surprised this didn't introduce a bug when I ran through the quest.

I'll correct it and re-run through all the quest steps again to ensure it's working as expected. I'll also test this edge case again.

59blargedy:
https://youtu.be/3NL-h8TGReM
https://www.dropbox.com/s/4nqthprth9wdaxq/matre%20bell.rar?dl=0
for matre bell dropping and reacquisition

mrhappyasthma:
Thank you for this! Very helpful. I'll update the pull request to handle this case as well.

mrhappyasthma:
Adjusted the final comment. And added logic for the matre bell repurchase.

I thoroughly tested out every edge case I could think of. So hopefully it's pretty bug-free now.

Please take another look.

scripts/globals/effects/healing.lua
mrhappyasthma: Done.

scripts/zones/Western_Adoulin/npcs/Sylvie.lua
```diff
@@ -0,0 +1,102 @@
...
+    if csid == 37 then
+        if option == 1 then  -- gil
+            if player:getGil() >= 300000 then
+                player:setLocalVar("Sylvie_Matre_Bell", 1)
+                player:updateEvent(7, 1, 2964, 0, 0, 0, 4480, 1)
+            end
+        elseif option == 2 then  -- bayld
+            if player:getCurrency('bayld') >= 150000 then
+                player:setLocalVar("Sylvie_Matre_Bell", 2)
+                player:updateEvent(256, 0, 127, 0, 0, 0, 4095, 1)
```
ibm2431: For updateEvent params, retail will often send "junk data" as params which aren't required for the event (usually random memory values). We usually try to keep params limited to what we know is the cause of different behavior (so it's easier to tell which params are "important" when debugging in the future).

I toyed with the EventUpdate locally, and param 8 is what determines Sylvie's next line for a chosen option:
```
0 - You can't afford it
1 - Okay here you go
2 - You don't have the inventory space
```

I was a little surprised to find the special inventory space line. While doing a `npcUtil.giveItem` check is usually all one needs to do, Sylvie is special.

Thinking something like:
```lua
local updateParam = 0
if player:getFreeSlotsCount() >= 1 then
    if option == 1 and player:getGil() >= 300000 then
        player:setLocalVar("Sylvie_Matre_Bell", 1)
        updateParam = 1
    elseif option == 2 and player:getCurrency('bayld') >= 150000 then
        player:setLocalVar("Sylvie_Matre_Bell", 2)
        updateParam = 1
    end
else
    updateParam = 2
end

player:updateEvent(0, 0, 0, 0, 0, 0, 0, updateParam)
```

You can still keep `giveItem` inside the `if` for the `onFinish`; you'll need it to give the item anyway!

mrhappyasthma: Ah,  I see. I just mirrored the exact params for updateEvent from the captures, to be safe.

I'm happy to clean that up, as I think it simplifies the code too by removing the rest of the seemingly magic numbers.

After work I can try to apply this change and play around with it to verify that it works for all edge cases.

mrhappyasthma: Rewrote this. Those params worked great. Please take a look.

mrhappyasthma:
Updated the params and rewrote the logic to be more straightforward for `onEventUpdate`.

ibm2431 (approved):
Thanks, mrhappyasthma! Glad to see our favorite Geomancer have something to say!

PR #487, opened 2020-04-11
Make ENABLE_TRUST_QUESTS use obvious

zach2good:
**_I affirm:_**

Make the checks obvious/correct


ibm2431 (approved):

ibm2431:
I really don't need to wait on builds for this.

PR #488, opened 2020-04-11
Limbus SQL updates

cocosolos:
**_I affirm:_**


ibm2431:
@wrenffxi : I'm going over the level ranges for this and comparing them to some of @59blargedy 's captures. Some of these groups (from your draft PR I'm assuming) are a bit higher than what aether picked up: level 100+ Aerns, Wynavs, Euvhis, and Ghrahs (aether's were in the 80-82 range).

I'm guessing the level 100+ versions are from Central Temenos/Apollyon II, right?

cocosolos: Do you think we should append the version of Limbus to group names so the level 80 Temenos Aerns (group 95) don't get confused for the level 100 Temenos Aerns (group 101)?

ibm2431:
Oh, checking your draft PR again, you added a comment to the group saying where they were from. Sorry for the mention, the comment got lost on its way here!

ibm2431 (approved):

59blargedy (approved):

PR #492, opened 2020-04-13
Remove food from the list of drainable effects for Thinker Tap moves

ibm2431:
Contrary to what wikis claim, food is not a buff that can be stolen by these moves. This is been confirmed on retail.

See the following capture on the FFXI Captures Discord server:
- NM[Brooder] [Ruminator] (Abyssea - La Theine) [+] Food Absorb Test

"Fixes" #149 . The status container still needs work, and checks for cancel-able buffs need to happen (in general, not just for buff theft), but Tap moves are currently the only abilities that caused crashing in this manner (that I'm aware of, at least).

**_I affirm:_**
- [🤞] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

### DIFF
--- a/scripts/globals/mobskills/binary_tap.lua
+++ b/scripts/globals/mobskills/binary_tap.lua
@@ -19,8 +19,8 @@ end
 function onMobWeaponSkill(target, mob, skill)
 
     -- try to drain buff
-    local effectFirst = mob:stealStatusEffect(target, tpz.effectFlag.DISPELABLE+tpz.effectFlag.FOOD)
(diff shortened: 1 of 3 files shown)


TeoTwawki:
Do to the likelihood somebody comes along remembering it happening and seeing “wiki says” it would be great if someone with a wiki account would update those and link the video proof in the edit description there.

It probably was a real thing in the distant past and just never got any mention by SE.

59blargedy (approved):

PR #499, opened 2020-04-15
Fix AH 7-item limit not being enforced correctly

### LINKED ISSUE #86: AH allowing more than 7 listings
_Sunday Jan 24, 2016 at 04:05 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 2673_

----

Logged into whasf's oldschool server to verify. I was able to list far more than 7 slots worth.

Please note that if you at all check your sales status, the client will impose its own restriction for the duration of the session. If you never check sales history, client relies on the server to restrict it. The server restriction is somehow broke but used to work.

Kreidos:
AH listing count is now checked at every sale,
and not just the cached value from the last time
player viewed their Sale History.

Fixes #86

As previously discussed, the limit is enforced on the client so there's no reason to make it a configurable value.

**_I affirm:_**

### DIFF
--- a/conf/default/map.conf
+++ b/conf/default/map.conf
@@ -66,6 +66,10 @@ ah_tax_rate_single: 1.0
 ah_tax_rate_stacks: 0.5
 ah_max_fee: 10000
 
+# Max open listings per player, 0 = no limit. (Default 7)
+# Note: Settings over 7 may need client-side plugin to work under all circumstances.
+ah_list_limit: 7
+
 #Misc EXP related settings
 exp_rate: 1.0
 exp_loss_rate: 1.0
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -951,6 +951,7 @@ int32 map_config_default()
     map_config.ah_tax_rate_single = 1.0;
     map_config.ah_tax_rate_stacks = 0.5;
     map_config.ah_max_fee = 10000;
+    map_config.ah_list_limit = 7;
     map_config.exp_rate = 1.0f;
     map_config.exp_loss_rate = 1.0f;
     map_config.exp_retain = 0.0f;
@@ -1102,6 +1103,10 @@ int32 map_config_read(const int8* cfgName)
         else if (strcmp(w1, "ah_max_fee") == 0)
         {
             map_config.ah_max_fee = atoi(w2);
+        }
+        else if (strcmp(w1, "ah_list_limit") == 0)
+        {
+            map_config.ah_list_limit = atoi(w2);
         }
         else if (strcmp(w1, "exp_rate") == 0)
         {
--- a/src/map/map.h
+++ b/src/map/map.h
@@ -79,6 +79,7 @@ struct map_config_t
     float  ah_tax_rate_single;        // Percent of listing price to tax single items
     float  ah_tax_rate_stacks;        // Percent of listing price to tax stacks
     uint32 ah_max_fee;                // Maximum total AH fees/taxes
+    uint32 ah_list_limit;             // Maximum open AH listings per player
 
     float  exp_rate;                  // множитель получаемого опыта
     float  exp_loss_rate;             // same as exp rate but applies when player dies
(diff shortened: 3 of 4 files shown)


Kreidos:
This could also alleviate a common cause of #140, where players can list way too many items and overflow their mailboxes.

Wiggo32:
Not sure what the general consensus would be but, I know there are a good number of private servers that actually like the fact players can list more than 7 items (especially when the population is lower). Though this fix is definitely needed, I'm wondering if making a setting that would allow unlimited items to be listed would make things easier on a lot of the private servers who wish to run it that way.

zach2good:
+1 for wrapping this in a setting, where the default is retail behaviour. Otherwise looks good!

TeoTwawki:
Unless its as simple as commenting something out, server owners would probably love a setting. I'm glad to see it finally getting fixed though. 

Seeing it required an additional query kind of tells me the original code structure could use some refactoring later.

Kreidos:
> +1 for wrapping this in a setting, where the default is retail behaviour. Otherwise looks good!

I agree, it would be handy to configure it and I had considered making it a config setting, the issue is that the client enforces the 7-item limit even if the server isn't. If the client ever learns there are 7 items or more in the history list, it won't allow any more to be posted.  Even previously, only the first 7 items of the "effectively unlimited" listings would be shown in the Sales History. From my understanding, the only proper way to do this would be to use a modified client which doesn't enforce the limit. 

Alternatively we could only show the last 6 listings to the client so it believes there's always room for one more. The issue with that is the listings become a 6-length FIFO window and any additional listings aren't shown and players can't see or cancel them. 

Third option we could remove the 7-item cap on the sales listing and just show them all. This should work, however the behaviour would still mirror the way it was; where if the player leaves the AH interface and comes back, or if they view sales history then the client forbids additional listings. This just feels buggy.

TeoTwawki:
@Kreidos the client doesn't enforce its limit until the player views the history. this can then be reset by relogging or rezoning. If you don't look, the client will let you keep listing till you break AH (exceeding max number of a specific item, or max returnable items for the character).

There is also an addon for ahsita that bypasses this entirely.

I'm in favor of just getting the fix merged and worrying about how to make a setting after  if doing so is problematic in any way :+1: (as a server owner, I would just temporary roll back the change if I want infinity listings again, though an easy toggle is nice to have)

Kreidos:
> There is also an addon for ahsita that bypasses this entirely.

@TeoTwawki I wasn't aware such an easy client-side fix was available, any chance you could link which plugin that is?

I have made the value configurable but will need to test it when I get home. I'll push the new commit once I'm confident nothing blows up.

TeoTwawki:
I shouldn't have said "bypasses". I just checked and all its doing is stopping the player from looking and tripping the client side check. the client doesn't apply its limit until we check the status of our current on sale items for some reason. I also thought this was on ashita's list but its not (though ahgo and instantah are - they allow using ah from anywhere and remove the delay in opening ah menus, respectively), which means I dunno where I got this from.

Kreidos:
That's unfortunate. Still, having it as-configurable-as-is-possible on the server end is probably good; someone may write a better client-side fix later.

Kreidos:
I've tested and pushed the new commit which makes it configurable. 

The existing client-side issues (blocking beyond 7 after viewing history, and history only showing the first 7 even when more are in the system.) will still be present if values above 7 are used. In testing, it wasn't possible to show more than 7 items in the AH Sale Status screen even if they exist in the DB, the client just ignores them.

TeoTwawki:
Random trivia: client limit of 7 is due to retails AH server crashing and them giving up on getting it working. They started at 10, went to 8, then 7 and called it quits and in more than 14 years never revisited the issue other than to make excuses pretending this did not happen when people ask for a slot increase. I was in game when they pulled the emergency maintenance for it both times.
https://giphy.com/gifs/reactionseditor-xUPGcl3ijl0vAEyIDK

ibm2431 (approved):
Thanks, Kreidos! We're happy to have you here with us at Project Topaz!

PR #500, opened 2020-04-15
Add config flag to dis/allow new accounts

Kreidos:
There are surely people hosting private servers for themselves or their friends who would prefer not have open registrations. True, the chance of someone port scanning you, finding your server, and logging in is minimal but most server emulators offer an option to disable registration entirely. This PR adds that capability.

Adds new flag "account_creation" (true/false) in login.conf file which permits new accounts to be created. Defaults to true. A one-line warning will also show at topaz_connect startup if this flag is set False, as this could be a cause of issues or confusion if someone had done so inadvertently or had forgotten they'd done so.

Considerations:
If account creation is disabled, topaz_connect will always return 0x04 ERROR_CREATE when receiving a new account request, which is visibly shown as "Username is taken" in xiloader. This is of course not **entirely** accurate, but as xiloader only recognizes a few return codes, this was the most suitable available.

**_I affirm:_**


src/login/login_auth.cpp
```diff
@@ -208,7 +208,20 @@ int32 login_parse(int32 fd)
         }
         break;
         case LOGIN_CREATE:
-            //looking for same login
+            Sql_EscapeString(SqlHandle, escaped_name, name.c_str());
+            Sql_EscapeString(SqlHandle, escaped_pass, password.c_str());
```
ibm2431: [Looks like strings are already being escaped before the switch starts up at L103 and L104](https://github.com/project-topaz/topaz/blob/7e139a4645911acfca0712bd43a2a7a4efbe3041/src/login/login_auth.cpp#L102-L106).

Kreidos: Totally was. Can be removed.

Fixed.

ibm2431 (approved):
Thanks, Kreidos!

conf/default/login.conf
```diff
@@ -65,3 +65,6 @@ msg_server_ip: 127.0.0.1
 
 #Logging of user IP address to database (true/false)
 log_user_ip: false
+
+#New accounts may be created? (true/false)
```
TeoTwawki: maybe rephase to specify that creation via the loader/lobby is what is disabled.

Kreidos: Is there another method for account creation? I know in theory they can be added to the database directly, linking into some web front-end but that wouldn't involve the server and I assume anyone doing that level of integration wouldn't be confused by the comment here.

I was aiming for it to be simple for newer users who may not know whether they're missing another setting somewhere to disable any theoretical "other types" of account creation methods. Do you have a specific phrase you'd prefer?

TeoTwawki: Answered your own question: I was thinking people might mistake it for preventing any new accounts, when in reality it will be used mainly to steer people toward one of those webpage signups you just referenced.

> I assume anyone doing that level of integration wouldn't be

haha oh man...If you only knew the things I've had to explain :(

TeoTwawki: I guess I'll suggest:
`#Allow account creation via the loader (true/false)`

Kreidos: Sounds clear enough, I like it. 

Changed.

PR #502, opened 2020-04-17
Add noKiller to onMobDeath

cocosolos:
Usage of noKiller is similar to isKiller, but is only true when a mob dies
unclaimed. Useful for executing lua on mob death when something needs to
happen only once and a player object isn't required.

Example usage:
```lua
function onMobDeath(mob, player, isKiller, noKiller)
    if isKiller or noKiller then
        -- open a door, spawn a chest, show some death dialogue, etc
        -- do not use player with noKiller
    end
end
```

**_I affirm:_**


TeoTwawki:
I think this is more intuitive than checking for nil player/killer - isKiller pushes nil instead of bool when there isn't one, but someone is bound to be confused if they see us check for it being true or being nil on the same line.

PR #505, opened 2020-04-18
[#490] Fix crash on zone with party

### LINKED ISSUE #490: topaz_game crashing when 2+ characters on zone teleport
**_I have:_**

- x[] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened

This is happening on the Canary branch.
When 2+ characters are zoning at the same time from one zone to another it will make the map server crash.

```[12/Apr] [16:07:10][Info] parse: 016 | 16AD 169B 04 from user: Mio
[12/Apr] [16:07:11][Info] parse: 05E | 176A 1758 0C from user: Hiro
[12/Apr] [16:07:11][Info] Zoning from zone 100 to zone 102: Hiro
[12/Apr] [16:07:11][Info] parse: 00D | 176B 1759 04 from user: Hiro
[12/Apr] [16:07:11][Debug] CZone:: West_Ronfaure DecreaseZoneCounter <2> Hiro
[12/Apr] [16:07:11][Debug] Message: Received message 8 from message server
[12/Apr] [16:07:12][Info] parse: 016 | 17A8 1797 04 from user: Mei
/bin/sh: 1: Syntax error: Bad fd number
[12/Apr] [16:07:14][Fatal Error] --- gdb backtrace ---
[12/Apr] [16:07:14][Status] luautils::free:lua free...         - [OK]
topaz_game: /usr/include/zmq.hpp:406: zmq::context_t::~context_t(): Assertion `rc == 0' failed.
[12/Apr] [16:07:14][Fatal Error] --- gdb backtrace ---
double free or corruption (!prev)```

### PR DESCRIPTION
**Pull request by zach2good**
**_I affirm:_**

Fixes #490 

Not yet tested, but given @Kreidos did some digging and it looks like this check will guard against looking at `loc.zone` while it's `nullptr`. Will try and get set up to do some testing later, unless someone else gets to it first.

The existing guarding getter:
```
uint16 CBaseEntity::getZone()
{
    return loc.zone != nullptr ? loc.zone->GetID() : loc.destination;
}
```

### DIFF
--- a/scripts/globals/spells/recall-jugner.lua
+++ b/scripts/globals/spells/recall-jugner.lua
@@ -12,11 +12,13 @@ function onMagicCastingCheck(caster,target,spell)
 end
 
 function onSpellCast(caster,target,spell)
-    if (target:hasKeyItem(tpz.ki.JUGNER_GATE_CRYSTAL) == true) then
+    if target:getObjType() == tpz.objType.PC then
+        if target:hasKeyItem(tpz.ki.JUGNER_GATE_CRYSTAL) then
             target:addStatusEffectEx(tpz.effect.TELEPORT,0,tpz.teleport.id.JUGNER,0,4.7)
-        spell:setMsg(tpz.msg.basic.MAGIC_TELEPORT)
-    else
-        spell:setMsg(tpz.msg.basic.NO_EFFECT)
+            spell:setMsg(tpz.msg.basic.MAGIC_TELEPORT)
+        else
+            spell:setMsg(tpz.msg.basic.NO_EFFECT)
+        end
     end
     return 0
 end
--- a/scripts/globals/spells/recall-meriph.lua
+++ b/scripts/globals/spells/recall-meriph.lua
@@ -12,11 +12,13 @@ function onMagicCastingCheck(caster,target,spell)
 end
 
 function onSpellCast(caster,target,spell)
-    if (target:hasKeyItem(tpz.ki.MERIPHATAUD_GATE_CRYSTAL) == true) then
+    if target:getObjType() == tpz.objType.PC then
+        if target:hasKeyItem(tpz.ki.MERIPHATAUD_GATE_CRYSTAL) then
             target:addStatusEffectEx(tpz.effect.TELEPORT,0,tpz.teleport.id.MERIPH,0,4.7)
-        spell:setMsg(tpz.msg.basic.MAGIC_TELEPORT)
-    else
-        spell:setMsg(tpz.msg.basic.NO_EFFECT)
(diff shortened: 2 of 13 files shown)


Kreidos:
I did some testing and unfortunately I was still able to reproduce the crash. I have updated #490 with details.

zach2good:
So, there are now two crashes in here that are fixed:
If you were running a debug build and teleported with a Trust in your party, it would try and teleport the trust and check to see if the trust had the right key item. This triggered a debug break. Letting the trust through is fine, because it will try and fail to cast the entity into a char, and return false anyway. It's a little hacky, but is probably fine as a quick fix.

The second bug was trying to access the PLeader in a party while zoning and they weren't available. This check was needed to see if the party definition packets that have information about trusts needed to be sent. I did a little cleanup of that interface and the whole flow is a little better now. 

Hopefully those crashes are squashed now.

ibm2431:
I rebased from canary back onto trust~

I wonder if we're falling into a trap re: the key item checking. While `hasKeyItem` returning false for trusts _is_ a valid answer to the explicit question the programmer is asking, letting trusts through might obscure causes of problems later - for example, what will happen if a player with trusts tries to enter content where everyone requires a KI?

Potentially outside of the scope of this PR, but the subject of teleports has me thinking: What happens if a player casts Warp II on a trust? Retrace? Escape?

Wiggo32 (approved):

Wiggo32 (approved):

ibm2431 (approved):

TeoTwawki:
> I rebased from canary back onto trust~
> 
> I wonder if we're falling into a trap re: the key item checking. While `hasKeyItem` returning false for trusts _is_ a valid answer to the explicit question the programmer is asking, letting trusts through might obscure causes of problems later - for example, what will happen if a player with trusts tries to enter content where everyone requires a KI?
> 
> Potentially outside of the scope of this PR, but the subject of teleports has me thinking: What happens if a player casts Warp II on a trust? Retrace? Escape?

when they first came out, just to see wtf happened I tried it. with most spells I could not target them. AoEs like teleport-dem -did- hit them tho, and after a few tries saw my trusts despawn before I myself got zoned. SO I think they answer is they despawn.

As far as non spell teleport pads, if there is a zone change they get left behind and then despawn. if there is not a zone change, they get re-positioned around you as if you ran really far at high speed. occasionally one gets left behind and lost (stop being a ditz kupipi, get up here and heal me!) which I think was because my new position wasn't far enough from her old position or another position update hit her just prior and the AI got confused. I had to go use the despawn all /command and recast to get my ducklings in a row again.

PR #515, opened 2020-04-22
Casket additions and adjustments

Omnione:
• Moved loot tables to casket_loot.lua as it was a bloating caskets.lua.
• Added all remaining zone mixins, treasure_casket.lua's and checked zone ID text tables.
• Added all remaining loot tables for zones added.
• Fixed an issue where a hint would return an incorrect number.
• Fixed an issue where you cannot enter the number 10 as an input.

**_I affirm:_**


scripts/globals/caskets.lua
```diff
@@ -243,8 +98,10 @@ local function timeElapsedCheck(npc)
     local spawnTime   = os.time() + 360000 -- defualt time in case no var set.
     local timeTable   = {0, 0, 0}          -- HOURS,MINUTES,SECONDS.
 
-    spawnTime = npc:getLocalVar("[caskets]SPAWNTIME")
-
+    if npc ~= nil then
+        spawnTime = npc:getLocalVar("[caskets]SPAWNTIME")
+    end
+
```
ibm2431: Thoughts on doing:
```lua
if npc == nil then
    return false
end
```
To immediately abort the rest of the function?

scripts/globals/casket_loot.lua
```diff
@@ -0,0 +1,4462 @@
...
+            { 690, 12706}, -- Iron Mittens
+            {1100, 12731}, -- Velvet Cuffs
+            { 860, 12827}, -- Cuir Trousers
+            { 720, 12836}, -- Iron Subligar
+            { 490, 12859}, -- Velvet Slops
+            {1120, 12955}, -- Cuir Highboots
+            { 390, 12987}, -- Ebony Sabots
+        }
+    },
+        [tpz.zone.QUICKSAND_CAVES] =
```
ibm2431: Got an extra indent on this key

Omnione: stupid copy pasta

scripts/globals/caskets.lua
Omnione: thats a good call i'll do that

Omnione: im gonna add a check the see if the casket has a local var set for its spawn time too.

scripts/globals/casket_loot.lua
Omnione: Fixed

scripts/globals/caskets.lua
Omnione: Added the the suggested change plus a check for the local var.

ibm2431 (approved):
Thanks, Omni!

Issue #401, opened 2020-03-04
Crafting Skill up not working when set to 1.0


kaincenteno:
**_I have:_**

- [x] searched existing issues (http://github.com/project-topaz/topaz/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Get crafting to level 25.
Get a level 30 kit.

After 8 stacks of level 30 kit. No skill ups will happen. Formula seems to be bad when setting chance of crafting skill up to 1.0

TeoTwawki:
Just for clarification: you were able to reach 25 at that same setting right? I mean this is not turn it to 1.0 and skillups break, this is with it at 1.0 thats as far as you could go and ***then*** skillups just stopped right?

kaincenteno:
@TeoTwawki  to clarify:
Server was on 2x settings, then reverted to 1.0 for retail accuracy.

I was already level 25 crafting when starting to test 1.0

I did try skilling a craft from 0 to 1 and it increased normally. Did not test when it stops skilling up

TeoTwawki:
K so going from 0 to 1 is possible at 1.0 rate then? That would support the theory I was floating in discord about bad scaling.

kaincenteno:
@kaincenteno  yes, going from 0 to 1 at 1.0 rate is possible.

kaincenteno:
crafting at level 40 stops when the rate is set to 2.6

TeoTwawki:
I've made it above 40 with it at 2.6 it just turns agonizingly slow. made it to 100 with it at 3.0

kaincenteno:
disregard my last message, it was a dumb moment i forgot i was capped lol. thanks @TeoTwawki

ibm2431:
#779

PR #517, opened 2020-04-23
Some crafting fixes

Xaver-DaRed:
- Translated comments
Also added some more

- Made HQ calculations take into account the level of all skills involved in a recipe, not just highest.
Retail sets HQ tiers, depending on the levels of all skills involved. Topaz and DSP on the other hand just does it based on the highest lvl skill.
- Removed unused function

- Made desynths work
  If a recipe is a desynth, Success rate is lowered and HQ rate is raised.

- Removed bias against lighting crystals.
  It was being used a substitute for desynth.

Thanks to Teo for the help. ^^

**_I affirm:_**


TeoTwawki:
> Removed bias against lighting crystals.
Crafting (right now) doesnt check if a recipe is a synth or a desynth.
Desynths are suposed to have a lower success rate and a higher HQ rate.
Currently, it just lowered success rate whenever a Lightning crystal is used, wich isn't correct, since not all recipes that use a lightning crystal are desynths and not all desynths use lightning crystals.

[There is actually a database field for this](https://github.com/project-topaz/topaz/blob/release/sql/synth_recipes.sql#L28) that for completely dumb reasons was never used in the core.

The existing rows in the table will need checked to ensure they have a consistent use of it (some rows may have the incorrect value in that column at present, suggest running a select query to see which ones have it set to `1` and which don't after decide if it will be 1 or 0 to mean "yes its a desynth")

kaincenteno:
Does this PR fix https://github.com/project-topaz/topaz/issues/401 ? :cat:

Xaver-DaRed:
nope, this doesnt fix #401, although...

double skillUpChance = ((double)baseDiff*(map_config.craft_chance_multiplier - (log(1.2 + charSkill/100))))/10;

this line might have something to do with it

Xaver-DaRed:
@TeoTwawki suposedly, that value is stored.

line 69 and 70
SELECT ID, KeyItem, Wood, Smith, Gold, Cloth, Leather, Bone, Alchemy, Cook, Result, ResultHQ1, ResultHQ2, ResultHQ3, ResultQty, ResultHQ1Qty, ResultHQ2Qty, ResultHQ3Qty, Desynth

FROM synth_recipes

this takes the values from synth recipe list and this....
line 115

PChar->CraftContainer->setCraftType((uint8)Sql_GetUIntData(SqlHandle, 18)); // Store if it's a desynth

so it should be as easy as pulling the value out of the CraftContainer and doing a few "if" checks here and there, but i suck at programing, so i dont know how to properly write....

if CraftContainer cell18 =1 //if desynth
{
    success  = success/2
}

TeoTwawki:
If someone actually added it I dunno why they left the crystal check in instead of 
```cpp
if (PChar->CraftContainer->getCraftType() ==  1)
{
    doTheThing;
}
```
According to the snippet you quoted its being fed into `setCraftType()` already, so it should already be there for the matching "get"

Xaver-DaRed:
gonna test that and see if it works.

TeoTwawki:
Does any wiki mention any diff in HQ rates on desynths? I kinda feel like only mt "break" rate was worse on desynths, but I don't have data so you can't trust me on that.

Xaver-DaRed:
https://ffxiclopedia.fandom.com/wiki/Desynthesis
"Also, unlike normal crafting, the chances of getting a High Quality result from Desynthesis is greatly increased"

I also read it from various old forums

In the end, this numbers are guess-work and need to be tested and tweaked.

This PR aims to make crafting less cheaty than what we currently have, wich are easy HQs and desynths not being desynths

TeoTwawki:
awesome, so actually goes up, not down

src/map/utils/synthutils.cpp
```diff
@@ -492,32 +476,31 @@ uint8 calcSynthResult(CCharEntity* PChar)
...
-*  Ломать не строить. Ломание столов и стульев не сделает из нас        *
-*  плотника, значит умение в данном случае не повышается.               *
-*                                                                       *
-************************************************************************/
+/********************************************************************
+*                                                                   *
+* We are trying to increase the skill of the character.             *
+* Desynthesis does not increase skill. NOT TRUE, RATE IS JUST LOWER *
+* Breaking does not build. Breaking tables and chairs won't make us *
+* carpenter, then skill in this case does not increase.             *
```
ibm2431: Is this currently being reflected in the code? If not, the comment can be removed - we don't need to keep the inaccurate comments 😉

ibm2431 (commented):
Hello, Xaver-DaRed! Happy to have you here with us at Project Topaz!

I'm sorry this took me a bit to get around to - I saw the line diffs and initially thought it'd be more mentally taxing to review than it turned out to be! 😅

Xaver-DaRed:
I've updated the comment block and added some to-do things left.
Even then, i'd rather this be reviewed and tested by more people other than me before correcting anything else, mainly to ease any future edits and reviews

ibm2431 (approved):

PR #518, opened 2020-04-23
Misc. updates and additions.

ghost:
**_I affirm:_**

Hello.

- Fixed and added some BaseSell prices (looks like even older items which had a price in the table needs to be updated).
- Fixed a model ID (Adapa Shield). Intentionally left out other things Lelia added (item_basic, item_latents, item_mods, item_weapon and .lua scripts). I tried to include them but was having errors I can't figure out.
- Fixed Weapon Bash message ID not being displayed.
- Renamed OPENING_CUTSCENE_ENABLE to NEW_CHARACTER_CUTSCENE in the topaz\scripts\globals\settings.lua file (with multiple occurrences in starting zones) then turned it back on by default. 

Thanks!


kaincenteno:
great job @UynGH  :cat:

59blargedy (approved):

ibm2431 (approved):

PR #525, opened 2020-04-25
Fix Tractor removing status effects on zone

cocosolos:
For effects with EFFECTFLAG_ON_ZONE.

**_I affirm:_**


cocosolos:
Never mind, guess it's supposed to do this. Not sure about Dynamis.

Issue #512, opened 2020-04-22
Closed Position trait is given freely at 75 without putting merits in it.


tankfest:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

```INSERT INTO `traits` VALUES (96,'closed position',19,75,1,0,0,'WOTG',0);```

It is supposed to be obtained from merit group 2, but any 75 dnc gets it.

ibm2431:
Fixed by #526

PR #526, opened 2020-04-25
Fix for DNC closed position

Omnione:
ability to be added only after a merit has been added to the trait.

This fixes the following issue:

Closed Position trait is given freely at 75 without putting merits in it. #512

**_I affirm:_**


ibm2431 (approved):
Thanks, Omnione!

Others: [ID confirmation](https://github.com/Windower/Resources/blob/master/resources_data/merit_points.lua#L285)

ghost (approved):

59blargedy (dismissed):

PR #527, opened 2020-04-25
Fix COP 6-4 Airship Fight

RunAwayDSP:
Add missing var player:setCharVar("bcnm_instanceid", inst) to Airship Door so you can progress through COP 6-4

**_I affirm:_**


cocosolos:
This was fixed in #471

PR #529, opened 2020-04-26
Update CoP mission IDs to retail and expand mission log

cocosolos:
Adds 32 bits of extra data to the mission log for tracking progress in CoP and possibly other missions, is sent as part of the mission update packet for CoP. This will allow the mission log to display branching missions correctly. Has migrations to first expand the mission logs of any existing characters, then to convert their CoP mission ID to the correct retail value. It is very important to run these migrations.

**_I affirm:_**


zach2good:
LGTM 👍

ibm2431 (approved):
Looks good to me!

PR #530, opened 2020-04-26
Remove FR server message

zach2good:
**_I affirm:_**

This has always seemed a little vestigial and I'd be interested to hear if anyone actually uses it


TeoTwawki:
Trivia:
Retail stopped supporting it ages ago but the bits are still in client and just so happen to work if we send the byte for the no longer used languages. 


No real reason to keep it, and we never did get around to supporting the Japanese one. All non-Asian languages can be put into the English server message without issue, but the Asian character set will come out trashed every time (SE uses a non-standard Shift JIS variant).

ibm2431:
Slapping a WIP label on this until you edit the title and/or say otherwise~!

zach2good:
Tested with `!updateservermessage`, all seems fine. Even with error (when I swap around to other branches), the error is just a single line log saying it couldn't find the file, so very safe to merge.

ibm2431 (approved):

ghost (approved):

PR #531, opened 2020-04-26
Globalized Gobbiebox Script and added scripts for all goblins

Noikar:
**_I affirm:_**


kaincenteno:
Awesome @Xephera  :D 
Those goblins are going to make everyone happy
Dunno if it might need to be rebased cuz there's way too many changes

PR #533, opened 2020-04-27
Globalized Gobbiebox Script and added scripts for all goblins 

Noikar:
**_I affirm:_**
- [X] that I agree to Project Topaz's [Limited Contributor License Agreement](https://github.com/project-topaz/topaz/blob/master/CONTRIBUTOR_AGREEMENT.md), as written on this date

Didn't do adoulin gobs cuz they're not spawned. onTrade code needs to be added by someone else. Also added the "you must be this old to ride" messages in all zones with goblins

### DIFF
--- a/scripts/zones/Aht_Urhgan_Whitegate/npcs/Wondrix.lua
+++ b/scripts/zones/Aht_Urhgan_Whitegate/npcs/Wondrix.lua
@@ -4,27 +4,27 @@
 -- Gobbie Mystery Box
 -- !pos -215.5 0.0 -147.3
 -----------------------------------
-local ID = require("scripts/zones/Windurst_Walls/IDs")
+local ID = require("scripts/zones/Aht_Urhgan_Whitegate/IDs")
 require("scripts/globals/settings")
 require("scripts/globals/gobbiemysterybox")
 -----------------------------------
 
 local events =
 {
-    INTRO                   = 519,
-    DEFAULT                 = 520,
-    HOLDING_ITEM            = 521,
-    TRADE                   = 522,
-    BAD_TRADE               = 523,
-    DAILY_COOLDOWN          = 524,
-    HIT_MAX                 = 525,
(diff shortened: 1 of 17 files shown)


kaincenteno:
awesome rebaze @Xephera  :cat: thank you for your contribution

scripts/zones/Aht_Urhgan_Whitegate/npcs/Wondrix.lua
```diff
@@ -0,0 +1,40 @@
+-----------------------------------
+-- Area: Windurst Woods
+--  NPC: Arbitrix
```
ibm2431, line 3: Wondrix is trying to steal Arbitrix's identity! Even falsely proclaiming their residence as Windurst Woods instead of Aht Urhgan Whitegate!

scripts/zones/Aht_Urhgan_Whitegate/npcs/Wondrix.lua
```diff
@@ -0,0 +1,40 @@
...
+    BAD_TRADE               = 983,
+    DAILY_COOLDOWN          = 984,
+    HIT_MAX                 = 985,
+    RESULT                  = 988,
+    KEY_TRADE               = 989,
+    NO_THANKS               = 990,
+    FULL_INV                = 991,
+    OTHER_BAD_TRADE         = 992,
+    ITEM_CANNOT_BE_OBTAINED = 219,
+    LIL_BABY                = 833
```
ibm2431, line 27: I think TOO_YOUNG might be a more obvious descriptor~ 😉

ibm2431 (commented):
Thanks Lynnea~

Noikar:
Yeah I was hoping someone would do the headers for me cuz I was swamped
that day lol. Also I figured LIL_BABY would be changed but I cracked myself
up so I left it

PR #534, opened 2020-04-27
Update default settings

cocosolos:
- Increase base player speed to 50.
- Enable homepoint teleport.
- Fix disabling of homepoint teleport.
- Set max level to 99.
- Use force pop land king system.
- Fix order of IDs for ranged relics.

**_I affirm:_**


scripts/globals/homepoint.lua
```diff
@@ -205,46 +205,51 @@ tpz.homepoint.onEventUpdate = function(player, csid, option)
...
-                temp = favs[x]
-                favs[x] = index
-                index = temp
-            end
-        elseif choice == selection.REM_FAVORITE then
-            for x = 1, 9 do
-                if favs[x] == index then
-                    for x = x, 8 do
-                        favs[x] = favs[x+1]
+            if HOMEPOINT_TELEPORT == 1 then
```
ibm2431: Got an extra `if HOMEPOINT_TELEPORT == 1` here

cocosolos: Whoops, fixed.

ibm2431 (approved):

PR #536, opened 2020-04-27
Fix CS priority + remove "set homepoint?" events (to match retail)

ibm2431:
I _assume_ this will also let GMs escape their moghouses. But for all I know @zach2good might just have an overly attached Moogle.

**_I affirm:_**

### DIFF
--- a/scripts/zones/Bastok_Markets/Zone.lua
+++ b/scripts/zones/Bastok_Markets/Zone.lua
@@ -17,8 +17,15 @@ end
 function onZoneIn(player, prevZone)
     local cs = -1
 
+    -- FIRST LOGIN (START CS)
+    if player:getPlaytime(false) == 0 then
+        if NEW_CHARACTER_CUTSCENE == 1 then
+            cs = 0
+        end
+        player:setPos(-280, -12, -91, 15)
+        player:setHomePoint()
     -- SOA 1-1 Optional CS
-    if
+    elseif
         ENABLE_SOA and
         player:getCurrentMission(SOA) == tpz.mission.id.soa.RUMORS_FROM_THE_WEST and
         player:getCharVar("SOA_1_CS2") == 0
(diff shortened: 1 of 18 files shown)


zach2good:
Will test this evening 👍 
EDIT: Looks good

ibm2431:
Will need another commit to clean up the eventFinishes, which I spaced.

Thanks for pointing them out, Dera!

ibm2431:
Removed the no longer accessible eventFinishes which I forgot.

I did not need to do so for Adoulin because the nagging Moogles never made it there.

Nor the Shadowreign cities because although the events were being started, they had no finishes and _never once set a homepoint_.

scripts/zones/Bastok_Mines/Zone.lua
```diff
@@ -29,21 +29,15 @@ function onZoneIn(player, prevZone)
...
         player:setPos(-45, -0, 26, 213)
         player:setHomePoint()
+    -- ENTER THE TALEKEEPER
+    elseif prevZone == tpz.zone.ZERUHN_MINES and player:getCurrentMission(BASTOK) == tpz.mission.id.bastok.ENTER_THE_TALEKEEPER and player:getCharVar("MissionStatus") == 5 then
+        cs = 176
     end
 
     -- MOG HOUSE EXIT
     if player:getXPos() == 0 and player:getYPos() == 0 and player:getZPos() == 0 then
         position = math.random(1, 5) - 75
```
TeoTwawki: _spots a global variable_

scripts/zones/Bastok_Markets/Zone.lua
```diff
@@ -17,32 +17,26 @@ end
...
     if player:getPlaytime(false) == 0 then
         if NEW_CHARACTER_CUTSCENE == 1 then
-            --cs = 0
+            cs = 0
         end
         player:setPos(-280, -12, -91, 15)
         player:setHomePoint()
+    -- SOA 1-1 Optional CS
+    elseif
+        ENABLE_SOA and
```
TeoTwawki: I coulda sword someone pr'd a fix to that check, since its 0 or 1 instead of nil/false/true

scripts/zones/Port_Jeuno/Zone.lua
```diff
@@ -26,6 +26,13 @@ function onZoneIn(player,prevZone)
         -- No need for an 'else' to change it back outside these dates as a re-zone will handle that.
     end
 
+    if
+        ENABLE_ABYSSEA == 1 and player:getMainLvl() >= 30
```
TeoTwawki, line 30: this shouldn't be its own if, its should be in the main collection of if/else that determine cut scene events - its not like you can get 2 of them on the same zone-in.

TeoTwawki: never mind I see the rest of the below block is full a rats nest - will restructure clean up another day

PR #542, opened 2020-04-28
mob group QC: zones 1 - 50

wrenffxi:
**_I affirm:_**

* Adjust level range when retail capture revealed a level out of range
* Add group when retail capture included a mob we didn't have a group for
* Remove some no-longer-needed "after the break / high groupid" records


zach2good:
Thanks as always Wren!

ibm2431 (approved):
Thanks, Wren!

ghost (approved):

PR #545, opened 2020-04-29
Blue Mage Transformations quest + other 3 armor pieces

kaincenteno:
Currently a WIP. Only thing ready is Lathuya's logic. I need to verify that the order of the quest is correct, also one thing i am having issues is the zone:registerRegion and onZoneIn for it as i just added a testing print to player to make sure that the area is correct and i do not get any feedback.

**_I affirm:_**


kaincenteno:
I'm done with the quest :cat: 

Only thing missing are:
* 2 mob attacks for the NM that are missing, there were 3 missing attacks before but added one weird one that removes all buffs lol.
* correct location params for onRegionEnter for both optional cutscenes. Since they were options i didn't think it was really a stopped.

Now I'm at the mercy of a review. Please @UynGH @59blargedy remove the WIP tag and change it to ready for review :cat2:

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -195,5 +213,14 @@ function onEventFinish(player, csid, option)
     elseif csid == 713 and option == 1 and omensProgress == 3 and player:getGil() >= 1000 then
         player:delGil(1000)
         player:messageSpecial(ID.text.PAY_DIVINATION) -- You pay 1000 gil for the divination.
+
+    -- TRANSFORMATIONS
+    elseif csid == 721 and option == 1 and player:getGil() >= 1000 and omens == QUEST_COMPLETED then
```
ibm2431: We already know Omens is completed here in the eventFinish because we checked for it during the onTrigger 😉

(Will need to break 720 and 721 into different checks to account for one "starting" the quest and the other being the reminder)

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -92,6 +94,20 @@ function onTrigger(player, npc)
...
         end
 
+    -- TRANSFORMATIONS
+    elseif omens == QUEST_COMPLETED and transformations == QUEST_AVAILABLE and currentJob == tpz.job.BLU then
+        if not divinationReady then
+            player:startEvent(63)
+        elseif waoudNeedToZone == 1 then
+            player:startEvent(78, player:getGil()) -- dummy questions, costs you 1000 gil
+        else
+            player:startEvent(721, player:getGil()) -- starts Transformations
```
ibm2431: The very _first_ divination for Waoud during Transformations is 720. The hint during the first stage (ie: progress == 1) is event 721.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
```diff
@@ -92,6 +94,20 @@ function onTrigger(player, npc)
...
+        end
+    elseif transformations == QUEST_ACCEPTED then
+        if player:getCharVar("TransformationsProgress") == 2 then
```
ibm2431: Will it be possible to speak to Waoud after Transformations was accepted but _without_ being on stage 2? If not, can eliminate the stage check and only see if it's been accepted.

ibm2431:
You'll have a dedicated skill list ID to use after #584

sql/mob_pools.sql
```diff
@@ -2891,7 +2891,7 @@ INSERT INTO `mob_pools` VALUES (2830,'Nephiyl_Pinnacletosser','Nephiyl_Pinnaclet
 INSERT INTO `mob_pools` VALUES (2831,'Nephiyl_Rampartbreacher','Nephiyl_Rampartbreacher',126,0x0000BF0200000000000000000000000000000000,1,1,5,240,100,0,1,1,1,18,0,0,4216,131,0,0,0,1,0,126);
 INSERT INTO `mob_pools` VALUES (2832,'Nepionic_Bladmall','Nepionic_Bladmall',251,0x0000510500000000000000000000000000000000,1,1,7,290,100,0,1,1,1,0,0,0,3253,129,0,0,0,1,0,251);
 INSERT INTO `mob_pools` VALUES (2833,'Nepionic_Parata','Nepionic_Parata',251,0x0000510500000000000000000000000000000000,1,1,7,290,100,0,1,1,1,0,0,0,1247,129,0,0,0,1,0,251);
-INSERT INTO `mob_pools` VALUES (2834,'Nepionic_Soulflayer','Nepionic_Soulflayer',233,0x0000EF0600000000000000000000000000000000,4,4,5,240,100,0,1,1,1,2,0,0,0,131,0,0,28,1,0,233);
+INSERT INTO `mob_pools` VALUES (2834,'Nepionic_Soulflayer','Nepionic_Soulflayer',233,0x0000EF0600000000000000000000000000000000,4,4,5,240,100,0,1,1,1,2,0,0,0,131,0,0,28,1,0,1011);
```
ibm2431: When the Soulflayer_NM skill list is merged, I'll give `blue-mage` a merge from `release`, and then you can use that skill list ID (234) instead

sql/mob_skill_lists.sql
```diff
@@ -3461,3 +3461,10 @@ INSERT INTO `mob_skill_lists` VALUES ('Siren',1010,3512); -- Sonic Buffet
 INSERT INTO `mob_skill_lists` VALUES ('Siren',1010,3513); -- Entice
 INSERT INTO `mob_skill_lists` VALUES ('Siren',1010,3514); -- Hysteric Assault
 INSERT INTO `mob_skill_lists` VALUES ('Siren',1010,3515); -- Clarsach Call
+INSERT INTO `mob_skill_lists` VALUES ('Nepionic_Soulflayer',1011,1963);
+-- INSERT INTO `mob_skill_lists` VALUES ('Nepionic_Soulflayer',1011,1964);
+-- INSERT INTO `mob_skill_lists` VALUES ('Nepionic_Soulflayer',1011,1965);
+INSERT INTO `mob_skill_lists` VALUES ('Nepionic_Soulflayer',1011,1966);
+INSERT INTO `mob_skill_lists` VALUES ('Nepionic_Soulflayer',1011,1967);
+INSERT INTO `mob_skill_lists` VALUES ('Nepionic_Soulflayer',1011,1968);
+INSERT INTO `mob_skill_lists` VALUES ('Nepionic_Soulflayer',1011,1969); -- Reprobation
```
ibm2431: And with skill list ID 234, you wouldn't need to define a new list here~!

scripts/zones/Alzadaal_Undersea_Ruins/mobs/Nepionic_Soulflayer.lua
```diff
@@ -2,7 +2,22 @@
 -- Area: Alzadaal Undersea Ruins (72)
 --  Mob: Nepionic Soulflayer
 -----------------------------------
-mixins = {require("scripts/mixins/job_special")};
+mixins = {require("scripts/mixins/job_special")}
+-----------------------------------
+
+--TODO: Immortal Shield - Magic Shield (Only prevents direct damage from spells)
```
ibm2431, line 8: For Immortal Shield, see `effects/magic_shield.lua`. It seems to be used in `abilities/rampart`, `mobskills/mana_screen`, `mobskills/polar_bulwark`, and `mobskills/spectral_barrier`

scripts/zones/Alzadaal_Undersea_Ruins/mobs/Nepionic_Soulflayer.lua
```diff
@@ -2,7 +2,22 @@
...
+--TODO: Immortal Mind - Magic Atk Boost
```
ibm2431, line 9: We might able able to fill Immortal Mind out with a volunteer brave enough to eat a bunch of nukes from some soulflayers

scripts/globals/mobskills/reprobation.lua
```diff
@@ -0,0 +1,33 @@
...
+require("scripts/globals/msg")
+---------------------------------------------
+
+function onMobSkillCheck(target,mob,skill)
+    return 0
+end
+
+function onMobWeaponSkill(target, mob, skill)
+    local dispel =  target:dispelAllStatusEffect(bit.bor(tpz.effectFlag.DISPELABLE, tpz.effectFlag.FOOD))
+    local msg -- to be set later
```
ibm2431: You can save two lines by setting msg to be `SKILL_NO_EFFECT` right here when you define it, and then override it down below if dispel > 0.

scripts/globals/mobskills/reprobation.lua
```diff
@@ -0,0 +1,33 @@
...
+
+    if (dispel == 0) then
```
ibm2431: Don't need these outer parens 😉

scripts/zones/Alzadaal_Undersea_Ruins/npcs/21.lua
```diff
@@ -0,0 +1,37 @@
...
+        player:messageSpecial(ID.text.NOTHING_OUT_OF_ORDINARY)
+    end
+end
+
+function onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+    if csid == 4 then
+        SpawnMob(ID.mob.NEPIONIC_SOULFLAYER)
```
ibm2431: You can use npcUtil.popFromQM to spawn the soulflayer and lock out the spawn point so that other people can't use the QM to spawn a second soulflayer (and forcibly despawning one that might be up).

scripts/zones/Alzadaal_Undersea_Ruins/mobs/Nepionic_Soulflayer.lua
```diff
@@ -2,7 +2,22 @@
...
+
+function onMobInitialize(mob)
+    mob:setMobMod(tpz.mobMod.IDLE_DESPAWN, 180)
+end
+
+function onMobSpawn(mob)
+    DespawnMob(mob:getID(), 180)
```
ibm2431: Without a way to lock out the spawning point, I suspect that malicious players would be able to prevent another player from getting their win 😉

scripts/zones/Alzadaal_Undersea_Ruins/npcs/21.lua
```diff
@@ -0,0 +1,37 @@
...
+require("scripts/globals/npc_util")
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end;
+
+function onTrigger(player,npc)
+    local transformationsProgress = player:getCharVar("TransformationsProgress")
+    -- TRANSFORMATIONS
+    if transformationsProgress >= 2 and transformationsProgress <= 4 and not GetMobByID(ID.mob.NEPIONIC_SOULFLAYER):isSpawned() then
```
ibm2431: If using popFromQM to spawn the soulflayer, you won't need to check to see if the soulflayer is spawned, because the QM won't be available to interact with~!

I'm also not finding a source that the two CSes leading up to the soulflayer are optional~ 😅

scripts/zones/Alzadaal_Undersea_Ruins/npcs/21.lua
```diff
@@ -0,0 +1,37 @@
+-----------------------------------
+-- Area: Alzadaal Undersea Ruins
+--  NPC: 19 (no name)
```
ibm2431: But the script says it's 21~!

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -9,36 +9,129 @@ require("scripts/globals/quests")
...
 
+local craftingItems = {
+-- itemBit = {firstItem, secondItem, thirdItem, fourthItem, coinType, coinAmmount, craftedItem, optionId}
+    [1] = {754,  828,  879, 4158, 2186, 2, 14928, 1}, -- magus bazubands
+    [2] = {761,  828, 2175, 2340, 2186, 2, 15600, 2}, -- magus shalwar
+    [4] = {828, 2229, 2288, 2340, 2186, 4, 14521, 3}  -- magus jubbah
+}
+
 function onTrade(player, npc, trade)
+    local currentTask = player:getCharVar("LathuyaCurrentTask_Mask")
```
ibm2431: You might want an outer-wrapping if check for all this logic to see if the player is on or has completed Transformations 😉 

Or alternatively: Check if a "remaining AF" is >= 1.

Checking that currentTask isn't 0 might be a good idea too~

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -9,36 +9,129 @@ require("scripts/globals/quests")
...
+    local craftingBluArmor = player:getCharVar("CraftingBluArmor")
+    local totalCraftedPieces = player:countMaskBits(player:getCharVar("LathuyaCraftingList_Mask"))
```
ibm2431: And I think total crafted pieces here would be `3 - countMaskBits(BLUAFremaining)`

scripts/zones/Aht_Urhgan_Whitegate/npcs/Imperial_Whitegate.lua
```diff
@@ -85,5 +91,8 @@ function onEventFinish(player,csid,option)
         player:setCharVar("TOAU_RINGRECV", 0)
     elseif csid == 3155 and option == 6 then
         npcUtil.giveItem(player,129)
+    elseif csid == 722 then
+        player:addQuest(AHT_URHGAN, tpz.quest.id.ahtUrhgan.TRANSFORMATIONS)
```
ibm2431, line 94: Let's pop a charVar for `[BLUAF]Remaining` here, with a value of 7. Then we can use it as a "reverse mask" that the player will eventually set to 0 (and clear from the db).

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -9,36 +9,129 @@ require("scripts/globals/quests")
...
 function onTrigger(player, npc)
     local omensProgress = player:getCharVar("OmensProgress")
     local omens = player:getQuestStatus(AHT_URHGAN, tpz.quest.id.ahtUrhgan.OMENS)
+    local transformations = player:getQuestStatus(AHT_URHGAN, tpz.quest.id.ahtUrhgan.TRANSFORMATIONS)
+    local craftingBluArmor = player:getCharVar("CraftingBluArmor")
+    local craftedPieces = player:getCharVar("LathuyaCraftingList_Mask")
+    local totalCraftedPieces = player:countMaskBits(player:getCharVar("LathuyaCraftingList_Mask"))
+    local currentTask = player:getCharVar("LathuyaCurrentTask_Mask")
+    local pickupReady = vanaDay() > player:getCharVar("BluPaymentDay")
+    local craftmanAwake = vanaDay() > player:getCharVar("BluRestingDay")
```
ibm2431: Let's do server admins some favors and standardize the vars to all start with something like `[BLUAF]` so they can be found in one spot

`[BLUAF]Current`, `[BLUAF]Payment`, `[BLUAF]Resting`

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -50,5 +143,30 @@ function onEventFinish(player, csid, option)
...
+                player:setMaskBit(player:getCharVar("LathuyaCraftingList_Mask"), "LathuyaCraftingList_Mask", 1, true)
+            elseif currentTask == 4 then
+                player:setMaskBit(player:getCharVar("LathuyaCraftingList_Mask"), "LathuyaCraftingList_Mask", 2, true)
+            end
+            player:setCharVar("BluPaymentDay", 0)
+            player:setCharVar("CraftingBluArmor", 0)
+            player:setCharVar("LathuyaCurrentTask_Mask", 0)
+            player:setCharVar("BluRestingDay", vanaDay())
+            player:delKeyItem(tpz.ki.MAGUS_ORDER_SLIP)
+        end
```
ibm2431, line 184: Then after you give the item, you can check if `[BLUAF]Remaining` is 0. If it is, you can delete _all_ BLU AF associated variables (like Resting Day), so there are _zero_ vars leftover when a BLU has finished crafting all their items.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -9,36 +9,129 @@ require("scripts/globals/quests")
...
+
+            -- MAGUS JUBBAH
+        elseif option == 8 then
+            player:setCharVar("LathuyaCurrentTask_Mask", 0)
+            player:setMaskBit(player:getCharVar("LathuyaCurrentTask_Mask"), "LathuyaCurrentTask_Mask", 2, true)
+            local currentTask = player:getCharVar("LathuyaCurrentTask_Mask")
+            player:updateEvent(0, craftingItems[currentTask][1], craftingItems[currentTask][2], craftingItems[currentTask][3], craftingItems[currentTask][4])
+        elseif option == 9 then
+            player:updateEvent(0, craftingItems[currentTask][5], craftingItems[currentTask][6])
+        end
```
ibm2431: Shorter alternative:
```lua
if option >= 2 and option <= 9 then
    local updateType = option % 3
    if updateType == 2 then
        -- Choosing a piece
        local piece = 2 ^ math.floor(option / 4)
        -- Make sure the player isn't trying to cheat somehow
        if bit.band(piece, player:getCharVar("[BLUAF]Remaining")) > 0 then
            player:setCharVar("[BLUAF]Current", 2 ^ piece)
            player:updateEvent(0, craftingItems[currentTask][1], craftingItems[currentTask][2], craftingItems[currentTask][3], craftingItems[currentTask][4])
        end
    else
        -- Needs payment
        player:updateEvent(0, craftingItems[currentTask][5], craftingItems[currentTask][6])
    end
end
```

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -9,36 +9,129 @@ require("scripts/globals/quests")
...
+                if totalCraftedPieces == 2 then
+                    currentTask = 7 - craftedPieces
+                    player:startEvent(746, 0, 0, 0, 0, 0, 0, 0, craftingItems[currentTask][8]) -- 8th param is response
+                else
+                    player:startEvent(730 + csOffset, craftedPieces) -- Will prompt for choosing which armor to work on
+                end
+            else
+                player:startEvent(737 + (csOffset - 8)) -- Asleep message, wait until 1 day passes
+            end
+        elseif player:countMaskBits(player:getCharVar("LathuyaCurrentTask_Mask")) == 1 then
```
ibm2431: This is equivalent to checking if  `current task > 1`~

(Ie: No need to count the bits)

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -50,5 +143,30 @@ function onEventFinish(player, csid, option)
...
+        player:confirmTrade()
+    elseif csid == 736 + csOffset then
+        if npcUtil.giveItem(player, craftingItems[currentTask][7]) then
+            if currentTask == 1 then
+                player:setMaskBit(player:getCharVar("LathuyaCraftingList_Mask"), "LathuyaCraftingList_Mask", 0, true)
+            elseif currentTask == 2 then
+                player:setMaskBit(player:getCharVar("LathuyaCraftingList_Mask"), "LathuyaCraftingList_Mask", 1, true)
+            elseif currentTask == 4 then
+                player:setMaskBit(player:getCharVar("LathuyaCraftingList_Mask"), "LathuyaCraftingList_Mask", 2, true)
+            end
```
ibm2431: Then `[BLUAF]Available` could be set to `[BLUAF]Available - (2 ^ (currentTask - 1))`

(Or: You wouldn't need a "positive-oriented" list of things you've crafted, because you instead have a "negative-oriented" list of things remaining to _be_ crafted)

This would mean adjusting the final piece to have a key of 3 in your craftingItems table (so all the pieces / tasks line up). I want to say this shouldn't be a problem, as `LathuyaCraftingList_Mask` seems to only be used as a table key (and mask bit manipulation which we're changing), and isn't used for any CS or param offsets (you already calculate those based on the total number of pieces crafted).

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -9,36 +9,129 @@ require("scripts/globals/quests")
...
-            player:startEvent(771) -- Default dialog
         end
 
+    -- CRAFTING OTHER 3 BLUE MAGE ARMOR PIECES
+    elseif transformations >= QUEST_ACCEPTED then
+        if currentTask == 0 and totalCraftedPieces ~= 3 then
+            if craftmanAwake then
+                if totalCraftedPieces == 2 then
+                    currentTask = 7 - craftedPieces
+                    player:startEvent(746, 0, 0, 0, 0, 0, 0, 0, craftingItems[currentTask][8]) -- 8th param is response
```
ibm2431: Won't this always result in an 8th param of 3 (same as the Jubbah)?

scripts/zones/Alzadaal_Undersea_Ruins/Zone.lua
```diff
@@ -136,6 +140,16 @@ function onRegionEnter(player,region)
                 player:startEvent(1, tpz.besieged.getMercenaryRank(player))
             end
         end,
+        [24] = function (x)
+            if transformationsProgress == 2 then
+                player:startEvent(2)
+            end
+        end,
+        [25] = function (x)
+            if transformationsProgress <= 2 then
```
ibm2431: Should probably be transformationsProgress == 3 😉

ibm2431 (changes requested):
Always remember that I ❤️ you kain~

(I especially want to make sure that Lathuya is _solid_ because I suspect that her logic will be copy-pasted to other craftable AF jobs)

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
kaincenteno: didn't we have some weird discussions of what an AF was beforehand that went into a tangent when trying to explain what AF2 was? lol

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
kaincenteno: no cause they are not completed in order, if you complete the the 1st and 2nd item then it will result in 7- 1 - 2 = 4
if you complete the 1st and 3rd item then it will be 7 - 1 - 4 =  2
if you complete the 2nd and 3rd it will be 7 - 2 - 4 = 1
then depending on that it will pass the 1st 2nd or 3rd options (sorry i don't make sense explaining here but we can get on a video call and i can explain better)

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: i thought the check on eventfinish was needed for prevent cheating (unless i got told wrong before xD)

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
ibm2431: Oh, because `craftedPieces` is the pieces crafted, not `totalCraftedPieces`.

I was reading the second line as `currentTask = 7 - totalCraftedPieces`

So nevermind~

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
ibm2431: If you want to make the tag `[BLUAF1]`, you'd have my _full_ support~! 😉

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
ibm2431: "it depends"

If the check was done to _start_ the event, then a recheck on eventFinish is redundant (ie: quest progress before firing event). A player can't cheat an event start.

But if the initial startEvent _didn't_ require a certain check, and the check only becomes relevant due to a player's choice during it, then that thing should be checked onFinish before messing with it (ie: gil). A player _can_ cheat an option choice.

kaincenteno: regarding 721 and 720, they are the same cutscene so i didnt add 720.

Removing check with questcompleted as requested.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: both the hint and start of transformations are exactly the same one, this is why i didnt include one of the cutscenes.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Waoud.lua
kaincenteno: I'm unsure, that would require a volunteer to talk to waoud after each step is completed

scripts/zones/Alzadaal_Undersea_Ruins/Zone.lua
kaincenteno: this will require further testing as you have mentioned these CS are not optional. For example, wiggo's capture only shows the 2nd CS on that area and not the 1st one.

kaincenteno:
I'm burnt out after writing this quest. Please make the required changes.
Sorry for my candor.
I'll try to focus on small quests from now on.

ibm2431:
Sure thing! Things like this _can_ be pretty grinding.

You're not in this all by yourself~!

ibm2431:
Pushed a commit to flip the "orientation" of the charVars for AF crafting to start at max and go towards zero, making sure all associated vars are gone when the player has crafted all the AF pieces.

And I tested it too! I have made sure that my changes didn't break kain's code and cause this to happen:
discord


As it's my own code, I'll tag @cocosolos in case they want to look at the most recent commit.

I intend to push another commit later for the other small things (hopefully tomorrow) .

scripts/zones/Alzadaal_Undersea_Ruins/Zone.lua
cocosolos: It looks like the order is always Caedarva Mire exit -> Mt Zhayolm exit. Currently both of these can be skipped the way char vars are being checked at the QM and NM.

scripts/zones/Aht_Urhgan_Whitegate/npcs/Lathuya.lua
```diff
@@ -9,37 +9,143 @@ require("scripts/globals/quests")
...
+        currency = 2186,
+        currencyAmt = 4,
+        result = 14521
+    }  
+}
+
 function onTrade(player, npc, trade)
+    local remainingBLUAF = player:getCharVar("[BLUAF]Remaining") -- Bitmask of AF the player has NOT crafted
+    if remainingBLUAF >= 1 then
+        local currentTask = player:getCharVar("LathuyaCurrentTask_Mask")
```
cocosolos: Is this being used anywhere?

ibm2431: Nope, it _was_ but I guess I accidentally left it in there when reworking it; will remove. Nice eye~!

ibm2431:
Pushed a second commit for various fixes and cleanup, including the Undersea regions and CSes. Had to rework Waoud and Whitegate a little after trying them as non-BLU and naked (respectively) caused unexpected behavior.

I set the Soulflayer to use the skill list ID that will be available to it in `release`. This branch is _way_ overdue for a update from `release` (switching to it stills need to deal with non-ignored confs). I'll do that merge after this PR is merged into blue-mage, and I'll uncomment Reprobation then.

I'm okay with merging this without Immortal Shield and Immoral Mind being scripted. We can add those later (after some testing on retail for Immortal Mind). Can potentially hold off on a `blue-mage` -> `canary` merge if those two missing skills are a concern for balance/"earn it" reasons.

ibm2431 (approved):

cocosolos (approved):

PR #546, opened 2020-04-29
Fixes Thuban and Herensegue's WS.

tankfest:
**_I affirm:_**

Thuban and Heren's current WS is set to Behemoth's Thunderbolt, they need to have the raptor version thunderbolt_breath to avoid having a spasm when using their WS.


ghost:
Hello, bruisefest. Thank you!

ghost (approved):

Issue #493, opened 2020-04-13
Novalmauge causes server crash due to pathing.


tankfest:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Novalmauge has crashed our server 3-4 times before I finally caught him doing it with VS's debugger.  It has to do with his pathfinding, when I disabled that, he no longer causes crashes.  Unfortunately the callstack that told me the specifics is long lost.  The crashes occurred when players would stop him to turn in his various quests.

I believe this issue is related to #319

kaincenteno:
npc pathin is just broken :cry:  if it ever gets fixed it would be awesome. Also there are some npc pathing that currently work but that they are all hacky as they have a series of coordinates to make them seem retail accurate so they don't slide.

would be awesome if they npc pathing gets fixed.

TeoTwawki:
`wait()` can trigger crashes, so its any roaming NPC that uses that command to stop moving when a  player clicks on it.

copy pasta from previous discussions:

```cpp
/************************************************************************
*  Function: wait()
*  Purpose : Makes a non-PC inactive for a set amount of time
*  Example : npc:wait(10000) -- wait 10 seconds
*  Notes   : Default is 4 seconds unless specified in ms
************************************************************************/

inline int32 CLuaBaseEntity::wait(lua_State* L)
{
    DSP_DEBUG_BREAK_IF(m_PBaseEntity->objtype == TYPE_PC);

    CBattleEntity* PBattle = (CBattleEntity*)m_PBaseEntity;

    int32 waitTime = 4000;

    if (lua_isnumber(L, 1))
    {
        waitTime = (int32)lua_tonumber(L, 1);
    }
    PBattle->PAI->Inactive(std::chrono::milliseconds(waitTime), true);

    return 0;
}
```
This is the version of wait NPC's use. The mob AI controller has a diff version that involves the server tick. I do not think we should be expecting this to work with `wait(0)` nor expecting a behavior of "keep waiting until event ends" which is how it was being used. 

I think we need to do something different at the core to genuinely do "wait till my event ends".

ghost:
Hello! Seems to be related to this old one too: https://github.com/project-topaz/topaz/issues/186.

TeoTwawki:
[...]

PR #548, opened 2020-04-29
Pathfind fix for NPC/MOB and poor old Novalmauge

Omnione:
• Addition of npcBasicPath to pathfind.lua, this pathfind function requires paths to be set in a linear fashion like this example
local path =
{
    [1] = {1.000, 2.000, 3.000},
    [2] = {2.000, 2.000, 3.000},
    [3] = {3.000, 2.000, 3.000},
    [4] = {2.000, 2.000, 3.000}
}
This way the path returns to the origin from the last path point to the first, i.e a circle.
the npcBasicPath will keep track of what path point the npc is on by the newly added m_pathPoint in npcentity.

• As stated above i have added a m_pathPoint to npcentity and given the the entity a getter and setter to help keep track of the npc's path.

• Adjusted Novalmauge's script to reflect the changes.

He's walk animation still seems a bit buggy but its about as good as i could get it.
But atleast it shouldn't crash the server anymore.
 
This should fix the following bug
Novalmauge causes server crash due to pathing. #493

**_I affirm:_**


scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
```diff
@@ -15,39 +15,38 @@ require("scripts/globals/quests")
...
-    42.256676, -24.000000, 19.885197,
-    41.168694, -24.000000, 19.904638,
-    21.859211, -24.010996, 19.792259,
-    51.917370, -23.924366, 19.970068,
-    74.570229, -24.024828, 20.103880,
-    44.533886, -23.947662, 19.926519
+    [1] = {20.000, -24.000, 20.000, 0},
+    [2] = {50.000, -24.000, 20.000, 0},
+    [3] = {75.000, -24.000, 20.000, 0},
+    [2] = {50.000, -24.000, 20.000, 0}
```
ibm2431: 123..2?

In either case, the table keys aren't needed; Lua will internally give un-keyed elements an increasing integer representation 😉 

(Unless you've changed the key structure with hard-indexed elements, but that's a different can of worms.)

src/map/lua/lua_baseentity.cpp
```diff
@@ -1741,6 +1741,46 @@ inline int32 CLuaBaseEntity::clearTargID(lua_State* L)
...
+*  Purpose : To set the current path point the entity is on
+*  Example : npc:setPathPoint(pathPoint +1)
+*  Notes   : See Novalmauge in Bostaunieux Oubliette
+*  Notes   : currently used for pathfind.lua npcBasicPath
+************************************************************************/
+
+inline int32 CLuaBaseEntity::setPathPoint(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_NPC);
```
ibm2431: If change the break to be only on players, we might be able to use this pathing method for mobs too~!

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
```diff
@@ -15,39 +15,38 @@ require("scripts/globals/quests")
...
 }
 
 local wsQuest = tpz.wsquest.spiral_hell
 
 function onSpawn(npc)
     npc:initNpcAi()
-    npc:setPos(tpz.path.first(path))
+    npc:setPos(path[1][1], path[1][2], path[1][3], 0)
+    npc:speed(12)
+    npc:setPathPoint(1)
```
ibm2431: Thoughts of baking this into `initNpcAi`? Can make it take a table, grab the first point while in core, set the pos, and also potentially take a walking speed as well to set as a local var.

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
```diff
@@ -15,39 +15,38 @@ require("scripts/globals/quests")
...
 end
 
 function onTrade(player, npc, trade)
     local wsQuestEvent = tpz.wsquest.getTradeEvent(wsQuest, player, trade)
-
+    
     if player:getCharVar("troubleAtTheSluiceVar") == 2 and npcUtil.tradeHas(trade, 959) then -- Dahlia
         player:startEvent(17)
-        npc:wait()
+        npc:speed(0)
```
ibm2431: If all paroling NPCs use this new method, will `npc:wait()` still have a use? Should we consider changing the internals of `:wait()` to essentially set speed to 0? (Would pathing need to be stopped?)

If we need to keep `npc:wait()`, should there be a wrapper like `tpz.path.pause()`?

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
```diff
@@ -109,5 +108,6 @@ function onEventFinish(player, csid, option, npc)
         tpz.wsquest.handleEventFinish(wsQuest, player, csid, option, ID.text.SPIRAL_HELL_LEARNED)
     end
 
-    npc:wait(0)
+    npc:speed(12)
```
ibm2431: If the "new" initNpcAi (which has a commented TODO to change the name) ends up setting a local var to the NPC, then could add a `tpz.path.resume()` which sets the speed back and calls `npcBasicPath`

ibm2431 (commented):
General question: I assume that setting speed to 0 actually does cease movement and doesn't cause sliding? I remember that being a question in #319

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
TeoTwawki: opinion: wait() does a lot of things besides stopping movement, it stops all the stuff and things. because of that its tempting to use it for a lot of situations that result in crashes. wait should be removed from its use in pathing scripts, but may have future use for..something.. 

could we possibly cause the event triggering core side to set the entities speed to 0 and then the event finish to restore it to its original value? :thinking: `m_event.Target` appears to be the npc in the core.

ibm2431: > m_event.Target appears to be the npc in the core

I was not aware of this - we can make event starts accurate! (On retail, the packet is the initiating NPC, but we've always invoked it directly as the "player" being the initiator)

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
Omnione: sounds good to me, seems a better way of doing it.

Omnione: Might aswell hold this till that gets implimented if thats what you want to do, then i can change it, might have to give me a nudge to change it incase i forget lol.

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
Omnione: yeah true, i guess i got so used to doing table like this for readability, that i just did it this way.

src/map/lua/lua_baseentity.cpp
Omnione: yeah sounds good to me

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
Omnione: implimented pathStop() and pathResume() to move away from using wait()

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
Omnione: I'm not really sure where to go with this but as we now have pathStop() and pathResume(), we will no longer get crashes when interacting with an npc that uses pathing, i think i found and converted all npc's that use pathing to use the new pathing systems.

src/map/lua/lua_baseentity.cpp
Omnione: This has been done now

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
Omnione: i did change the name of initNpcAI to initNpcPathing and you can now call it as npc:initNpcPathing() or pass an x,y,z to set the start position so npc:initNpcPathing(x,y,z) and also it will set the path point to 1 as default, but you can pass a 4th param to set the point also so npc:initNpcPathing(x, y, z, 5) will set the position and set the starting path point to 5.

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
Omnione: I can look into making event set the speed to 0 at a later date if still wanted.

scripts/zones/Bostaunieux_Oubliette/npcs/Novalmauge.lua
Omnione: all tables that use the new system should now be a simple point i.e. {10, 10, 10} with additional extra params able to be passed if required.

scripts/zones/Davoi/npcs/Quemaricond.lua
```diff
@@ -4,58 +4,57 @@
...
+    {35.431, -0.045, -20.227},
+    {51.795, -1.790, -18.843},
+    {58.037, -0.957, -18.724},
+    {65.919, -1.722, -18.767},
+    {92.143, -0.451, -16.231, {delay = 5}},
+    {65.919, -1.722, -18.767},
+    {58.037, -0.957, -18.724},
+    {51.795, -1.790, -18.843},
+    {35.431, -0.045, -20.227},
+}
```
zach2good, line 25: This section really nicely highlights the benefits of this PR!

scripts/zones/Davoi/npcs/Quemaricond.lua
```diff
@@ -4,58 +4,57 @@
...
-    npc:initNpcAi();
-    npc:setPos(tpz.path.first(path));
-    onPath(npc);
-end;
+    npc:initNpcPathing()
+    onPath(npc)
+end
 
 function onPath(npc)
+    tpz.path.general(npc, path, tpz.path.flag.WALLHACK, false)
```
zach2good, line 33: Looking at this "in the wild", I wonder if `general` is really the right name for the default, regular, general call?

Now that I say the other options, I guess it's as good as all the others 🤷

scripts/zones/Kazham/npcs/Kukupp.lua
```diff
@@ -3,75 +3,30 @@
...
+    {44.000, -11.000, -176.000},
+    {47.000, -11.000, -179.500},
+    {44.000, -11.000, -177.000},
+    {40.500, -11.000, -179.500},
+    {39.500, -11.000, -176.500},
+    {41.300, -11.000, -172.300},
+    {45.000, -11.000, -171.300},
+    {46.000, -11.000, -173.600},
+    {42.700, -11.000, -174.300},
+    {44.000, -11.000, -179.800},
```
zach2good, line 19: I hope you were able to automate this process and didn't have to manage all these tables by hand...

src/map/lua/lua_baseentity.h
```diff
@@ -108,10 +108,17 @@ class CLuaBaseEntity
...
 
+    int32 getPathPoint(lua_State* L);
+    int32 setPathPoint(lua_State* L);
     int32 atPoint(lua_State* L);             // is at given point
     int32 pathTo(lua_State* L);              // set new path to point without changing action
+	int32 stepTo(lua_State* L);
     int32 pathThrough(lua_State* L);         // walk at normal speed through the given points
-    int32 isFollowingPath(lua_State* L);     // checks if the entity is following a path
+    int32 isFollowingPath(lua_State* L);
+	int32 rotateToAngle(lua_State* L);
```
zach2good, line 118: Theres some borked tab indents around here

zach2good (approved):
So, it all looks good. There are some styling things I'd like to see fixed and in an ideal world, this would have been much easier to go through and review if you had split out the commits to be logic changes, whitespace/semicolon changes, tabling changes etc.

Sick work 👍

zach2good:
I think you've taken care of all semicolons and random things in Lua, but if you could take a merge of release just to be sure that would be grand

scripts/zones/Davoi/npcs/Quemaricond.lua
zach2good: Maybe refactor to `tpz.pathing.path()`, I don't know how much work that would be?

src/map/lua/lua_baseentity.cpp
```diff
@@ -1878,6 +1996,100 @@ inline int32 CLuaBaseEntity::isFollowingPath(lua_State* L)
...
+
+    uint8 default_speed = 0;
+
+    if (m_PBaseEntity != nullptr)
+    {
+        if (m_PBaseEntity->objtype == TYPE_NPC)
+        {
+            auto PNpc = static_cast<CNpcEntity*>(m_PBaseEntity);
+            uint32 npcID = PNpc->id;
+            const char* NPCQuery = "SELECT speed FROM npc_list WHERE npcid = %u";
```
TeoTwawki, line 2056: Why not store the original speed in a localVar instead of re-querying the database?
You can actually get and set localVars from the C++ side too, not just in lua.

TeoTwawki: For NPCs we could try having the event code itself stop and restart NPC movement. since `m_event.target` is the NPC entity itself. Mobs we'd still need a script command for stop/start.

src/map/lua/lua_baseentity.cpp
```diff
@@ -1878,6 +1996,100 @@ inline int32 CLuaBaseEntity::isFollowingPath(lua_State* L)
...
+* Example : npc:pathStop()
+* Notes   : used in conjunction with pathResume()
+************************************************************************/
+
+inline int32 CLuaBaseEntity::pathStop(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype == TYPE_PC);
+
+    CBattleEntity* PBattle = (CBattleEntity*)m_PBaseEntity;
```
TeoTwawki, line 2029: `m_PBaseEntity->speed = 0;` actually does work, no need to recast as a battleentity for either mobs or npc's (or players). Which makes this entire function in its current form a duplicate of `speed()`

Lets improve it by having this function also store the entities current speed to a localVar prior to setting it zero. Then resume path can read that value instead of querying the database every time.

```c++
 m_PBaseEntity->SetLocalVar(lua_tostring(L, 1), (uint64_t)lua_tointeger(L, 2));
```

TeoTwawki (changes requested):

src/map/lua/lua_baseentity.cpp
59blargedy: do we want to place all these in an issue or ?

TeoTwawki: not particularly no. I'd rather see it be changed here and its such a small correction to make instead of unnecessarily adding more iops of our inefficient db connection.

zach2good (changes requested):
I think @Wiggo32 had some things to say about this, putting on hold until they can chime in

PR #549, opened 2020-04-29
Change auto add master on UpdateEnmity

cocosolos:
Stop adding master by default when a pet calls `UpdateEnmity`. `ClaimMob` should handle this now when an actual claiming action is performed, and actions that don't claim (for example Automaton mobskills and spells) will only add the pet.

**_I affirm:_**


ibm2431 (approved):
Looks legit!

PR #550, opened 2020-04-30
Adds timer to spawn for Knight Crabs in Jugner.

tankfest:
**_I affirm:_**

Adds timer to spawn for Knight Crabs in Jugner after server reset.

### DIFF
--- a/scripts/zones/Jugner_Forest/Zone.lua
+++ b/scripts/zones/Jugner_Forest/Zone.lua
@@ -24,6 +24,11 @@ function onInitialize(zone)
     tpz.conq.setRegionalConquestOverseers(zone:getRegionID())
 
     tpz.helm.initZone(zone, tpz.helm.type.LOGGING)
+
+    local respawnTime = 900 + math.random(0, 6) * 1800 -- 0:15 to 3:15 spawn timer in 30 minute intervals
+    for offset = 1, 10 do
+        GetMobByID(ID.mob.KING_ARTHRO - offset):setRespawnTime(respawnTime)
+    end
 end
 
 function onZoneIn( player, prevZone)
--- a/scripts/zones/Jugner_Forest/mobs/King_Arthro.lua
+++ b/scripts/zones/Jugner_Forest/mobs/King_Arthro.lua
@@ -55,7 +55,7 @@ function onMobDespawn(mob)
     GetMobByID(KingArthroID):setLocalVar("[POP]King_Arthro", 0)
 
     -- Set respawn of 21:05 to 24:05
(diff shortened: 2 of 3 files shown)


scripts/zones/Jugner_Forest/Zone.lua
```diff
@@ -24,6 +24,11 @@ function onInitialize(zone)
     tpz.conq.setRegionalConquestOverseers(zone:getRegionID())
 
     tpz.helm.initZone(zone, tpz.helm.type.LOGGING)
+
+    local respawnTime = math.random(900, 10800) -- 15 minutes to 3 hours
```
TeoTwawki: If you want to do the spawn windows, they can be done with 
```
local respawnTime(900 * math.random(1, 12)
```
for 12 windows of 15 min each, though that coukd possibly be written better than how I just did to make it obvious whats going on.

Someday maybe we’ll just fee in window size and total max time to a helper function that spit out the right answer.

tankfest: i just followed Roc and Simurgh's lead... are they supposed to have precise windows too?

tankfest: fixed!  thanks teo

cocosolos: Wiki says 30 min windows `If Time of death is 6:00pm, the window will begin at 3:05pm the next day and windows will be at the 05 and 35 minute mark.` King Arthro himself should probably also be changed to reflect this. [See Fafnir.](https://github.com/project-topaz/topaz/blob/b72f5d756047375e9123a7cee0a2cb1783c07180/scripts/zones/Dragons_Aery/mobs/Fafnir.lua#L40)

scripts/zones/Jugner_Forest/mobs/King_Arthro.lua
```diff
@@ -55,7 +55,7 @@ function onMobDespawn(mob)
     GetMobByID(KingArthroID):setLocalVar("[POP]King_Arthro", 0)
 
     -- Set respawn of 21:05 to 24:05
-    local respawnTime = math.random(21, 24) * 3600 + 300
+    local respawnTime = math.random(0,6) * 1800 + 75900
```
TeoTwawki: can we use a consistent order of operations in the 2 scripts (and some clarifying paran's) ?  I know they execute the same, but people new to the language might not understand so extra clarity and consistency is good.

ibm2431 (approved):

ghost (approved):

PR #553, opened 2020-04-30
Replace all Couerl Charged Whiskers with correct mob skill ID.

tankfest:
**_I affirm:_**

3416 is Dragon Kick.  All couerls need 483.

### DIFF
--- a/sql/mob_skill_lists.sql
+++ b/sql/mob_skill_lists.sql
@@ -696,10 +696,10 @@ INSERT INTO `mob_skill_lists` VALUES ('Coeurl_CoP',161,482); -- pounce
 INSERT INTO `mob_skill_lists` VALUES ('Coeurl_CoP',161,652); -- blaster
 INSERT INTO `mob_skill_lists` VALUES ('Coeurl_CoP',161,653); -- chaotic_eye
 INSERT INTO `mob_skill_lists` VALUES ('Coeurl_CoP',161,1336); -- frenzied_rage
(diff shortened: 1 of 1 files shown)


tankfest:
It should be noted somewhere that 3416 is indeed dragon kick, and it needs to be changed in the sql file.  I don't have the capture info to update the parameters for the entire entry.

Also cocosolos theorized that a second charged whisker skill ID may have been shifted by SE, and it had a shorter CD time (i assume for NM use?).

TeoTwawki:
These are all the mob versions of Charged Whisker in the client:
483  (00227+256)	Charged Whisker
2407 (02151+256)	Charged Whisker
3912 (03656+256)	Charged Whisker
(+256 because we offset for formor and other mobs that use player ws)
We may or may not need the other 2 for some mob(s) someday. In a lot of cases we can use scripting to do the special things (like an extra effect or higher dmg) without using a new ID, usually need a new one for animation differences or CD  like you said.

tankfest:
Good to know.  I think in this case it wasn't a difference in cooldown, rather a difference in cast time, which also seems like it's skill ID specific.

TeoTwawki:
I knew what you meant I was just trying to use the same term you did for it.

59blargedy:
so i've fought sekhmet twice now, for 60 mins total with it sitting under 25%, and the mob does not  use any weaponskills at all. while wiki mentions charged whisker can be used, my only guess is it is used during lightningsday or the wiki is just wrong. will continue to test and just submit a new pr for sekhmet and dragon kick changes.

59blargedy (approved):

ibm2431 (approved):

PR #554, opened 2020-04-30
Update COR AF quest timer to current retail.

tankfest:
**_I affirm:_**

Update COR AF quest to current retail timer, 60 seconds.


ibm2431 (approved):
Thanks - easiest review I've had in a while!

59blargedy (approved):

PR #560, opened 2020-05-01
Add range check to wyvern's Healing Breath

### LINKED ISSUE #523: Dragoon's Wyvern will attempt to use Healing Breath (and fail) regardless of position or zone of low health pt member.
**_I have:_**

Be in a party with a low health member >30y away.  Use a spell that would normally trigger a healing breath from your wyvern, he will ready healing breath but not use it.  Need a position and zone check for the low health party member on this breath command.

Kreidos:
Does what it says on the tin. Previously the Wyvern would ignorantly try using healing breath even when members were on another map or way out of range.

Used the Wiki value of 13 for being 'in-range' for healing breath.

Fixes #523

**_I affirm:_**

### DIFF
--- a/scripts/globals/pets/wyvern.lua
+++ b/scripts/globals/pets/wyvern.lua
@@ -34,12 +34,17 @@ local wyvernTypes = {
 }
 
 function doHealingBreath(player, threshold, breath)
-    if player:getHPP() < threshold then
+    local breath_heal_range = 13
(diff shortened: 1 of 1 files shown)


ibm2431 (approved):
Thanks, Kreidos!

KnowOne134 (approved):

PR #561, opened 2020-05-01
ROV Chapter 1 mob moves

zach2good:
**_I affirm:_**

If I open a WIP PR, maybe it'll shame me into finishing this chunk of work so it can be merged into canary 👀 

UPDATE: I've filled in the missing moves and the bosses are starting to feel a bit more right. They will at least not sit around like potatoes doing 0 damage all the time. 

It looks like some of the animations are missing/wrong (but nothing is borked), so as work item for me later is to get more caps of these fights so I can fix them.


scripts/globals/mobskills/entice.lua
```diff
@@ -11,9 +12,19 @@ function onMobSkillCheck(target,mob,skill)
 end
 
 function onMobWeaponSkill(target, mob, skill)
-    local typeEffect = tpz.effect.SLEEP_II
+    local typeEffect = tpz.effect.CHARM_I
+    local power = 0
 
-    skill:setMsg(MobStatusEffectMove(mob, target, typeEffect, 1, 0, 20))
+    if (not target:isPC()) then
```
TeoTwawki: just a heads up: outer parens aren't needed
```lua
if not target isPC() then
```
works

kaincenteno:
omg aweome :D now with this the ROV branch can be merged into canary :+1:

scripts/globals/mobskills/entice.lua
zach2good: This is what I get for aggressive copy/paste...

TeoTwawki: :spaghetti:

scripts/globals/mobskills/lunatic_voice.lua
```diff
@@ -1,15 +1,23 @@
...
 require("scripts/globals/settings")
 require("scripts/globals/status")
 ---------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
-    return 0
+    if mob:hasStatusEffect(tpz.effect.SOUL_VOICE) then
+        return 0
+    end
+    return 1
```
ibm2431: Check for Soul Voice here is in the script for Lunatic Voice instead of Entice~ 😉

scripts/globals/mobskills/soulshattering_roar.lua
```diff
@@ -11,5 +11,17 @@ function onMobSkillCheck(target,mob,skill)
...
+    local duration = 10
+    local dmgmod = 2.0
+
+    MobPhysicalStatusEffectMove(mob, target, skill, typeEffect, 1, 0, 30)
+
+    local info = MobMagicalMove(mob, target, skill, mob:getWeaponDmg() * 4, tpz.magic.ele.DARK, dmgmod,TP_MAB_BONUS, 1)
+    local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.MAGICAL, tpz.damageType.DARK, MOBPARAM_WIPE_SHADOWS)
+    target:takeDamage(dmg, mob, tpz.attackType.MAGICAL, tpz.damageType.DARK)
+
+    -- TODO: Temporary immunity to a single weapon damage type
```
ibm2431, line 24: After merge someone will have to file an issue so we don't forget that this aspect isn't implemented.

We do have damage type immunities, but those are "set and forget" in nature, and the methods involved with making them temporary would be ugly. Will need dedicated effects for them in the future so limited duration can be baked into the initial call.

ibm2431 (changes requested):
Thanks, Zach! Move the [Soul Voice check over to Entice](https://www.bg-wiki.com/bg/Siren_(Notorious_Monster)) and I'm happy with this for now (and merging `rov` into `canary`)~

ibm2431:
Actually, does Siren have a spell list?

ibm2431 (approved):

sql/mob_skills.sql
```diff
@@ -2744,9 +2744,9 @@ INSERT INTO `mob_skills` VALUES (2885,2021,'beautiful_death',1,18.0,2000,1000,4,
 -- INSERT INTO `mob_skills` VALUES (2920,2664,'.',0,7.0,2000,1500,4,0,0,0,0,0,0);
 -- INSERT INTO `mob_skills` VALUES (2921,2665,'.',0,7.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (2922,2094,'soulshattering_roar',1,18.0,2000,1000,4,0,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (2923,2095,'calcifying_claw',1,18.0,2000,1000,4,0,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (2924,2096,'divesting_stampede',1,18.0,2000,1000,4,0,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (2925,2097,'bonebreaking_barrage',1,18.0,2000,1000,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (2923,2095,'calcifying_claw',0,18.0,2000,1000,4,0,0,1,0,0,0);
```
59blargedy, line 2747: was going to approve this but got confused by changing these values to 0 - does that mean no target? should it be 4 for target enemy for these two entries? apologies if i am missing something!

zach2good: That column corresponds to mob_skill_aoe:
https://github.com/project-topaz/topaz/blob/release/sql/mob_skills.sql#L33

I can't find the enum that describes it, but 0 is single-target:
https://github.com/project-topaz/topaz/blob/4d09173b57dd5867dd3997436fa3ee66eeea0fbb/src/map/mobskill.cpp#L61

59blargedy: ahh just me misreading the sql lines - sorry about that, thought it was mob_valid_target

59blargedy (approved):

PR #565, opened 2020-05-02
Elemental Dots functional with BLM merited abilities. Dot, and stat d…

Brierre:
…own tested on player casts and mob casts. Duration merits also working.

**_I affirm:_**

### DIFF
--- a/scripts/globals/spells/burn.lua
+++ b/scripts/globals/spells/burn.lua
@@ -33,7 +33,6 @@ function onSpellCast(caster,target,spell)
             end
             local sINT = caster:getStat(tpz.mod.INT)
             local DOT = getElementalDebuffDOT(sINT)
-            local DOTp = 0
             local effect = target:getStatusEffect(tpz.effect.BURN)
             local noeffect = false
             if (effect ~= nil) then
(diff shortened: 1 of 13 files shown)


m241dan:
As the original author of the code this person is fixing, I support this PR. Don't know how I glitched so hard and didn't just expand the original function.

TeoTwawki:
Note: We eventually still need to separate the dot from the stat down. While basing both off the power value works out using this function on the black mage spells it does not work accurately for the mob skills that inflict these status effects.

This is a pre-existing problem separate from what is being fixed however.

Brierre:
@TeoTwawki I've made a change to the magic.lua that should allow the dots to scale instead of the prescribed levels in the first commit (which will make Gates of Hades, etc. have the proper values).  I am unsure how to squish that in here so that it is clean.

TeoTwawki:
we prolly later (you don't need to just because of me) make it possible to send a different value than what the formula based on dot dmg is. Like we have dia/bio sending diff dot and status down by having one stored in power and the other in subpower. Just so that if you need to, you _can_ break the pattern without having to do workarounds. We'd still do what you are doing now by default.

(this should merge as is, imo, with possible exception to the FLAG_ERASABLE argument because I dunno what thats doing yet)

scripts/globals/spells/burn.lua
```diff
@@ -53,9 +52,8 @@ function onSpellCast(caster,target,spell)
 
                 local mbonus = caster:getMerit(tpz.merit.ELEMENTAL_DEBUFF_EFFECT)
                 DOT = DOT + mbonus/2 -- Damage
-                DOTp = DOTp + mbonus -- Stat Enfeeb
 
-                target:addStatusEffect(tpz.effect.BURN,DOT, 3, duration, tpz.effect.BURN, DOTp)
+                target:addStatusEffect(tpz.effect.BURN,DOT, 3, duration,FLAG_ERASABLE)
```
TeoTwawki: FLAG_ERASABLE is a nil value
tpz.effectFlag.ERASABLE would have been 2, but I do not no why you'd set that argument to the value of an effect flag enum. Sorry I didn't catch that earlier. :x

ibm2431 (approved):
Thanks, Brierre! Happy to have you with us!

PR #570, opened 2020-05-02
Add configurable subjob cap ratio..

TeoTwawki:
..and an option to apply to mobs that I hope is eventually deprecated by the ability to manually set mobs sj lv range. :smiley: 

The player options also apply to pets with the exception of charm pets, since those are mobs.

if `subjob_ratio` is 1, player subjobs are normal (one half) as in retail. This is the default option.
2 will set cap of two-thirds, so a lv 75 player can have a subjob up to lv 50, or a lv 99 player up to lv66
3 will make subjobs capped even to mainjob.
0 will make subjobs lv 0 and grant nothing.

I did not want to do this as a float value divisor or multiplier for several reasons, not the least of which is explaining to people the proper math with get the ratio they intended.

`include_mob_sj` determines if the adjustments also apply to monsters.

NOTE: since `SetSLevel()` handled the cap in its entirely, all existing attempts to set the subjob should lead back to here and rely on this. However any instance where a previous editor did not know this probably ***already had incorrect sub levels before I changed anything.***

**_I affirm:_**

### DIFF
--- a/conf/default/map.conf
+++ b/conf/default/map.conf
@@ -136,6 +136,16 @@ player_mp_multiplier: 1.0
 #Sets the fraction of MP a subjob provides to the main job. Retail is half and this acts as a divisor so default is 2
 sj_mp_divisor: 2.0
 
+# Modify ratio of subjob-to-mainjob
+# 0 = no subjobs
+# 1 = 1/2   (default, 75/37, 99/49)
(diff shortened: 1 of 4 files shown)


ibm2431 (approved):
Looks legit!

PR #573, opened 2020-05-02
Caskets: Split area loot

Omnione:
This allows areas with split mob levels to assign loot to caskets,
previously left out from a mistake on a rebase.

**_I affirm:_**

### DIFF
--- a/scripts/globals/caskets.lua
+++ b/scripts/globals/caskets.lua
@@ -58,6 +58,15 @@ local casketInfo =
         167, 169, 172, 173, 174, 176, 177, 178, 184, 190, 191, 192,
         193, 194, 195, 196, 197, 198, 204, 205, 207, 208, 212, 213
     },
+    splitZones = set{
+       tpz.zone.ZERUHN_MINES,
+       tpz.zone.KORROLOKA_TUNNEL,
+       tpz.zone.KING_RANPERRES_TOMB,
+       tpz.zone.ORDELLES_CAVES,
+       tpz.zone.OUTER_HORUTOTO_RUINS,
+       tpz.zone.GUSGEN_MINES,
+       tpz.zone.MAZE_OF_SHAKHRAMI
+    },
     cs =
     {
         [0]  = 1000, [1]  = 1003, [2]  = 1006, [3]  = 1009, [4]  = 1012, [5]  = 1015,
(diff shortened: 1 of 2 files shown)


scripts/globals/caskets.lua
```diff
@@ -401,10 +402,23 @@ function getDrops(npc, dropType, zoneId)
         local tempCount    = 1
         local randomTable  = {1,3,1,2,1,2,1,1,3,1,2,1}
 
+        if zoneId == 172 or zoneId == 173 or zoneId == 190 or
+           zoneId == 191 or zoneId == 193 or zoneId == 194 or
+           zoneId == 196 or zoneId == 198 then
```
ibm2431: Only thing I think might be improved on is defining these zones as a Set in the caskets global:
```lua
casketInfo.splitZones = set{
    tpz.zone.ZERUHN_MINES,
    tpz.zone.KORROLOKA_TUNNEL
}
```

And then your check here becomes:
`if casketInfo.splitZones[zoneId] then`

scripts/globals/caskets.lua
```diff
@@ -436,10 +450,22 @@ function getDrops(npc, dropType, zoneId)
         local itemCount    = 1
         local randomTable  = {1,4,1,3,1,1,2,1,3,1,2,1}
 
+        if zoneId == 172 or zoneId == 173 or zoneId == 190 or
+           zoneId == 191 or zoneId == 193 or zoneId == 194 or
+           zoneId == 196 or zoneId == 198 then
```
ibm2431: And with a Set already defined, the same check can be used here too: `if casketInfo.splitZones[zoneId] then`

Omnione: sounds legit, ill get on that

scripts/globals/caskets.lua
Omnione: yep sounds good

Omnione:
Ok tested the changes requested, all seems good, didnt even know we had set{}, i like that, ill have to remember that one.

ibm2431:
I slipped it in a while back. I've kept an eye out for opportunities to share its joys ever since!

ibm2431 (approved):
Thanks, Omnione!

Omnione:
lol nice

PR #574, opened 2020-05-03
Make Cactrot go Rapido again!

TeoTwawki:
- Fixed script logic (onMobRoam to onPath at spawn time, then recurring check to continue pathing in onMobRoam, with onPath sterting the patrol through nodes)
- Greatly reduced pathnodes while re-plotting several of them. These are not 100% accurate.
- nuked MOD_MOVE from  mob_pool_mods.sql and set speed in onMobSpawn instead. I am assuming when wiki says 250% they just mean its movement rate is 250, as they use the terminology interchangeably in many places. If I am wrong, his movement would be 125 assuming a base speed of 50, or 100 assuming a base speed of 40.Honestly none of these feel right, and I suspect he's 150 or 200. But feelings aren't good practice :P

Pending fresh retail captures, an overhaul to the pathing methods, and a change to send the entity update packets more often, this is as good as he gets.

P.S. I also corrected spelling on some unrelated comments my spellchecker caught while I was looking for his mods.

**_I affirm:_**
_I followed that punk around the desert for **three hours** man. **Three hours!** I got sand in places I didn't know I had, and now I just wanna kick back and enjoy a refreshing cactus cola._


ibm2431 (approved):
🌵 ☁️ ☁️

59blargedy (approved):

PR #575, opened 2020-05-03
Add key trading to mystery box

cocosolos:
Adds key trading feature for special, abjuration, fortune, and anniversary dials. Anniversary dial loot is currently not populated and will either give an item from the special dial or give the key back. Fortune key loot is populated based on best guess for what fits (wikis are incomplete).

**_I affirm:_**

### DIFF
--- a/scripts/globals/gobbiemysterybox.lua
+++ b/scripts/globals/gobbiemysterybox.lua
@@ -87,26 +87,20 @@ local gobbieJunk =
     19220
 }
 tpz.mystery.onTrade = function (player, npc, trade, events)
-    if not trade then
-        return false
-    elseif trade:getItemCount() == 1 then
-        local keyID = trade:getItemId(0)
-        if keyID == nil or keyID < 1 or keyID > 65535 or trade:getItemCount() ~= 1 or trade:getSlotQty(0) ~= 1 then
-            return false
-        else
+    if trade:getItemCount() == 1 then
+        local tradeID = trade:getItemId(0)
+        if keyToDial[tradeID] ~= nil then
             if player:getFreeSlotsCount() == 0 then
-                player:startEvent(events.FULL_INV, keyID, keyToDial[keyID])
-            elseif keyToDial[keyID] ~= 0 then
-                if player:getFreeSlotsCount() == 0 then
-                    player:startEvent(events.FULL_INV, keyID, keyToDial[keyID])
-                else
-                    player:setLocalVar("gobbieBoxKey", keyID)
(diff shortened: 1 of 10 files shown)


scripts/globals/gobbiemysterybox.lua
```diff
@@ -18,6 +18,97 @@ local costs =
...
+    15203,
+    18180,
+    19220
+}
+tpz.mystery.onTrade = function (player, npc, trade, events)
+    if not trade then
+        return false
+    elseif trade:getItemCount() == 1 then
+        local keyID = trade:getItemId(0)
+        if keyID == nil or keyID < 1 or keyID > 65535 or trade:getItemCount() ~= 1 or trade:getSlotQty(0) ~= 1 then
```
ibm2431: We don't need to check if `trade:getItemCount() ~= 1` because we've already confirmed the item count is 1 back on L92.

(And we may not need the nil or ID range checks either, because the trade container is already saying there's at least one valid item in it.)

scripts/globals/gobbiemysterybox.lua
```diff
@@ -18,6 +18,97 @@ local costs =
...
+            return false
+        else
+            if player:getFreeSlotsCount() == 0 then
+                player:startEvent(events.FULL_INV, keyID, keyToDial[keyID])
+            elseif keyToDial[keyID] ~= 0 then
+                if player:getFreeSlotsCount() == 0 then
+                    player:startEvent(events.FULL_INV, keyID, keyToDial[keyID])
+                else
```
ibm2431: Also don't need to check ` if player:getFreeSlotsCount() == 0 then` here on L100 because we already did so on L97 (so L100-L102 are impossible to reach).

scripts/globals/gobbiemysterybox.lua
```diff
@@ -53,9 +144,41 @@ tpz.mystery.onEventUpdate = function (player, csid, option, events)
...
 
-    if csid == event.DEFAULT then
+    if csid == event.KEY_TRADE then
+        if option == 1 then
+            local keyID = player:getLocalVar("gobbieBoxKey")
+            player:setLocalVar("gobbieBoxKey", 0)
+            switch (keyID): caseof
+            {
+                [8973] = function() itemID = SelectDailyItem(player, 6) end,  -- special dial
+                [9217] = function() -- abjuration
```
ibm2431: Thoughts on checking against keyToDial[keyID instead of the item IDs themselves?

ibm2431 (approved):
Thanks, cocosolos!

PR #576, opened 2020-05-03
#569 - onMobSpawn (Linux) no spells

### LINKED ISSUE #569: trust not using magic or job ability
**_I have:_**

just tested the canary branch to see the changes made by PR #454 . Kupipi doesn't cast magic and naji still doesnt provoke. (Curilla no casting either, didn't try shantoto)

brianmask:
**_I affirm:_**

change GetName() from packet_name to name
resolves #569 trust spells do not work on Linux

### DIFF
--- a/src/map/entities/trustentity.cpp
+++ b/src/map/entities/trustentity.cpp
@@ -49,11 +49,6 @@ CTrustEntity::CTrustEntity(CCharEntity* PChar)
     PAI = std::make_unique<CAIContainer>(this, std::make_unique<CPathFind>(this), std::make_unique<CTrustController>(PChar, this), std::make_unique<CTargetFind>(this));
 }
 
-const int8* CTrustEntity::GetName()
-{
-    return (const int8*)packetName.c_str();
-}
-
 void CTrustEntity::PostTick()
 {
     CBattleEntity::PostTick();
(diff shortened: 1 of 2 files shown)


src/map/entities/trustentity.cpp
```diff
@@ -49,11 +49,6 @@ CTrustEntity::CTrustEntity(CCharEntity* PChar)
     PAI = std::make_unique<CAIContainer>(this, std::make_unique<CPathFind>(this), std::make_unique<CTrustController>(PChar, this), std::make_unique<CTargetFind>(this));
 }
 
-const int8* CTrustEntity::GetName()
-{
-    return (const int8*)packetName.c_str();
```
TeoTwawki, line 54: Does the packetName still get sent in the packet to the client? As in if you cast `Trust: Selh'teus` does he pop on screen as `Selh'teus` or as `Selh_teus`/`Selh teus` ?

I think we need to do what allied mobs do, and have name and packetName as separate things, because both get used but for different purposes. Possibly this already happens because trusts are still mobs:
https://github.com/project-topaz/topaz/blob/8985d2af68876367592222836db1f8e5b15c8175/src/map/entities/mobentity.h#L244

If there's nothing to worry about then just mark my comment as resolved. :)

brianmask: Looks like it's properly pulling in the display name for the client.  I just checked Selh'teus, and his punctuation is correct.

ibm2431 (approved):
Thanks for looking into this, brianmask! Happy to have you here with us at Project Topaz!

PR #581, opened 2020-05-04
ensure trust spells are level appropriate

brianmask:
**_I affirm:_**
- x[] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

Purpose of this pull request is to ensure Trust Spells are both level appropriate and pass required checks in order to cast.

In the current implementation gambits registered with the SELECT_SPECIFIC reaction modifier will be cast regardless of level, recast timer and current MP.

This pull request adds one method to mob_spell_container (GetAvailable) in the same style as (GetBestAvailable) in order to check the trust has the spell in a spell container, has the appropriate mp to cast the spell, and the spell is currently not on cooldown.

This will still allow the trust lua files to register any needed gambits without regard to the trust level or other conditions and keep the flow as simple and straight forward as possible.

If a more appropriate method already exists rather than the new method added, I'm happy to modify the PR accordingly.

### DIFF
--- a/src/map/ai/helpers/gambits_container.cpp
+++ b/src/map/ai/helpers/gambits_container.cpp
@@ -135,7 +135,11 @@ void CGambitsContainer::Tick(time_point tick)
             {
                 if (action.reaction_mod == G_REACTION_MODIFIER::SELECT_SPECIFIC)
                 {
-                    controller->Cast(target->targid, static_cast<SpellID>(action.reaction_arg));
+                    auto spell_id = POwner->SpellContainer->GetAvailable(static_cast<SpellID>(action.reaction_arg));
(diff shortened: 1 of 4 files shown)


brianmask:
going to make a few more enhancements to not load the gambit in the first place if not applicable instead of checking every time if is applicable

brianmask:
modified logic slightly to check if spell is level appropriate during call to AddGambit
modified spell::CanUseSpell slightly to take in account for trust level during check
added GetAvailable to mob_spell_container to return SpellID if passed id is available and null if spell is on cooldown or trust does not have appropriate MP to cast

ibm2431 (approved):
Thanks, brianmask!

PR #583, opened 2020-05-04
Fix chain exp applying to non-chained kills.

tankfest:
**_I affirm:_**

Fixes the issue where an active exp chain always applies the current chain bonus exp to Decent Challenge and below mobs without giving a message.  Also eliminates several unnecessary conditional checks if the mob isn't EM+.  Thanks to cocosolos and aether for their help diagnosing and fixing this issue.


ibm2431 (approved):
This checks out as far as I can tell! Thanks, bruisefest!

Issue #325, opened 2020-01-08
Damage multiplying weaponskills not benefitting from TP


ibm2431:
**_I have:_**
- [x] searched existing issues (http://github.com/project-topaz/topaz/issues/) to see if the issue has already been opened or fixed
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Weaponskills which get an attack multiplier based on the player's current TP, like Tachi: Gekko, [do not have the player's current TP passed into the fTP function](https://github.com/project-topaz/topaz/blob/3eb1c1570f276c8510300a2993fc225fc50d4ba7/scripts/globals/weaponskills.lua#L671). `ftp()` ends up taking the WS's TP 1000 multiplier as the first argument, [but then bumps it up to 1000](https://github.com/project-topaz/topaz/blob/3eb1c1570f276c8510300a2993fc225fc50d4ba7/scripts/globals/weaponskills.lua#L641-L643), averting [the nil of ftp3 being undefined](https://github.com/project-topaz/topaz/blob/3eb1c1570f276c8510300a2993fc225fc50d4ba7/scripts/globals/weaponskills.lua#L646-L648).

ibm2431:
#585

PR #585, opened 2020-05-04
Fixed issue #325

Xlaits:
**_I affirm:_**

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -638,9 +638,6 @@ function getRangedHitRate(attacker,target,capHitRate,bonus)
 end
 
 function fTP(tp,ftp1,ftp2,ftp3)
-    --if (tp < 1000) then
-    --    tp = 1000
-    --end
     if (tp >= 1000 and tp < 2000) then
         return ftp1 + ( ((ftp2-ftp1)/1000) * (tp-1000))
     elseif (tp >= 2000 and tp <= 3000) then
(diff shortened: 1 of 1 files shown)


scripts/globals/weaponskills.lua
```diff
@@ -638,9 +638,9 @@ function getRangedHitRate(attacker,target,capHitRate,bonus)
 end
 
 function fTP(tp,ftp1,ftp2,ftp3)
-    if (tp < 1000) then
-        tp = 1000
-    end
+    --if (tp < 1000) then
```
zach2good: If these aren't being used anymore, they can just be deleted 👍

zach2good (commented):
Congrats on your first PR, welcome to being a contributor! Just one point about commented out code and you should be good to go

scripts/globals/weaponskills.lua
TeoTwawki: I think the intent was to floor at the minimum tp you could have in order to ws "just in case" some server decided to let you ws with less. I have no idea if thats even a problem but if it is I think that should be taken care of by whoever is customizing :)

ibm2431: And I believe it's not a necessary check either, as this function defaults to returning an ftp of 1 for invalid TP values _anyway_.

So I support the outright removal of the comment parts, @Xlaits ! 👍

ibm2431 (changes requested):
Yep, let's remove the commented lines. Otherwise this is good!

Congratulations on your first PR~! Happy to have you here with us!

ibm2431 (approved):
Pushed a commit that removed the previous commented lines! Approved.

Xlaits:
Awesome. Forgot to remove commented code when pushing the comit initially.

PR #587, opened 2020-05-05
Add configurable multiplier for normally-dropped gil

Kreidos:
Topaz lacked a really good way of controlling Gil payouts for those who don't like adding Gil to every single mob kill. This PR adds a modifier for the naturally-dropped Gil amounts. This multiplier does not apply to the gil from *all_mobs_gil_bonus*, only the normally-dropped amounts.

This combined with the all_mobs_gil_bonus affords much needed economic flexibility, including the option to set it to 0 and tweak all_mobs_gil_bonus to offer flat per-level Gil payouts server-wide.

**_I affirm:_**

### DIFF
--- a/conf/default/map.conf
+++ b/conf/default/map.conf
@@ -150,6 +150,9 @@ blood_pact_shared_timer: 0
 #Adjust mob drop rate. Acts as a multiplier, so default is 1.
 drop_rate_multiplier: 1.0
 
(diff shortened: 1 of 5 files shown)


ibm2431 (approved):
Looks legit! Thanks, Kreidos!

ghost (approved):

PR #564, opened 2020-05-01
Quest: Tuning In

InoUno:
**_I affirm:_**
- [x] that I agree to Project Topaz's [Limited Contributor License Agreement](http://project-topaz.com/blob/release/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [x] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

Added the [Tuning In](https://ffxiclopedia.fandom.com/wiki/Tuning_In) quest.

[Test video](https://www.youtube.com/watch?v=iwGovfoC8F8)

First time making a PR here. Let me know if anything should be done differently!

PR #590, opened 2020-05-06
Quest: Tuning Out

InoUno:
Next quest after the Tuning In quest, which was added in: #564

[Wiki page](https://ffxiclopedia.fandom.com/wiki/Tuning_Out)
[Test Video](https://www.youtube.com/watch?v=Kosx845G82o)

Notes:
* I tried the fight as a 75WAR to see how the difficult the Nasus are, and it matched the testimonials on the wiki page pretty well.
* ~~The Nasus positions I've added are just a semi-circle around the QM. I haven't been able to try the quest in retail, so I don't know how they spawn there.~~
* I took inspiration from other places in the codebase with regards to the progress of the quest and kill count on the Nasus. Is there a better approach to this now, or is this the way to go?

**_I affirm:_**


59blargedy:
Just a quick initial comment, you can find captured data for the quest here to adjust the spawn points:
https://discordapp.com/channels/443544205206355968/446470173008855050/644777354165616653
and you can make requests for future captures so that you have exact data

InoUno:
Cool, didn't know that existed. Will check it out and change the PR accordingly

scripts/zones/Yuhtunga_Jungle/mobs/Nasus.lua
```diff
@@ -0,0 +1,14 @@
+-----------------------------------
+-- Area: Yuhtunga Jungle
+--  Mob: Nasus
+-----------------------------------
+local ID = require("scripts/zones/Yuhtunga_Jungle/IDs")
+
+function onMobDisengage(mob)
+    DespawnMob(mob:getID(), 120)
```
ibm2431: Could be wrong, but I believe that once this fires, the mob will always despawn at that the end of the timer, even if it re-engages later.

We do have a mobMod for idle despawn, usage looks something like this:
```lua
-- mobs_script.lua
function onMobInitialize(mob)
    mob:setMobMod(tpz.mobMod.IDLE_DESPAWN, 120)
end
```
And it'll handle idle despawn for you automagically after that, without you having to worry about engage/disengage.

ibm2431 (commented):
This looks good! I _especially_ appreciate the comments of what's going on in an event so I could more easily follow the quest progression~

Only thing that jumps out to me is the despawn behavior:

ibm2431:
I'll go ahead and label this as WIP since you intend to check out the capture. Feel free to yell at us if we fail to notice you've added new commits!

scripts/zones/Yuhtunga_Jungle/mobs/Nasus.lua
InoUno: Nice, I'll use that instead then.
The code I had was something I found on another mob, and just reused it expecting it to be what I needed.

InoUno:
Changes are in.

I noticed while looking at the capture that some of the IDs don't match up with current server/client IDs. I'm assuming this is because there's been updates since the capture was done, that has moved/changed some of these?

* The QM has ID 172815589 in the capture, but in topaz the QM in the same spot has ID 17281590. The one with ID *89 does exist on topaz, but it's a bit further from the edge and is not visible by default. 
* In the capture, the message "A swarm appeared" has ID 40593, while in the latest retail client it has ID 40588 (which I've added to this PR).

ibm2431:
Yeah, client updates will sometimes shift both NPC IDs _and_ Message IDs! So sometimes if a capture is older the IDs may not align with what the current client uses.

ibm2431 (approved):
Thanks, InoUno!

scripts/zones/Yuhtunga_Jungle/IDs.lua
```diff
@@ -43,6 +43,7 @@ zones[tpz.zone.YUHTUNGA_JUNGLE] =
         NO_COMBINATION              = 7868,  -- You were unable to enter a combination.
         REGIME_REGISTERED           = 10046, -- New training regime registered!
         COMMON_SENSE_SURVIVAL       = 12040, -- It appears that you have arrived at a new survival guide provided by the Adventurers' Mutual Aid Network. Common sense dictates that you should now be able to teleport here from similar tomes throughout the world.
+        SWARM_APPEARED              = 40588, -- A swarm has appeared!
```
cocosolos: This ID looks like it should be `7820`. Think everything else looks good though.

scripts/zones/Yuhtunga_Jungle/IDs.lua
```diff
@@ -63,6 +65,7 @@ zones[tpz.zone.YUHTUNGA_JUNGLE] =
         OVERSEER_BASE         = 17281600,
         CERMET_HEADSTONE      = 17281625,
         PEDDLESTOX            = 17281640,
+        TUNING_OUT_QM         = 17281590,
```
ibm2431: Oh, on closer examination of the ID file, could you move this `17281590` definition up a bit to be in numerical order after the `BLUE_RAFFLESIA_OFFSET = 17281586,` offset?

scripts/zones/Yuhtunga_Jungle/IDs.lua
InoUno: Small test: 
image

I guess the higher bit is used for something that hides the NPC who's saying the message?
(`40588 - 2^15 = 7820`)

InoUno: In the capture, it doesn't have the `???: ` infront.

cocosolos: Don't pass `true` in the call to `messageText`, that's what's showing the NPC name.

Edit: Pass `false` instead, it defaults to `true`

ibm2431: [The messageText packet when called with `false` actually adds that offset you've found to the passed-in messageID to set that higher bit.](https://github.com/project-topaz/topaz/blob/release/src/map/packets/message_text.cpp#L36-L38)

InoUno: Aha, interesting. Will change ID to the lower one and pass `false` then.

Question, where did you find the 7820 ID?

cocosolos: [POLUtils](https://github.com/Windower/POLUtils)

InoUno:
ID changes are in.

cocosolos (approved):

PR #591, opened 2020-05-06
Add Threnody II + add NM Siren's spell list

ibm2431:
Adds the second tier Threnody spells. As these are awarded by gifts, there is no valid method for players to obtain them yet. There also seems to be a client lockout preventing them from appearing if !addallspells is used. However, mobs are capable of using them, including Siren.

**_I affirm:_**


TeoTwawki:
Does she actually cast in captures? The other avatars all fake their spells via mobskills. For both player pet and the primes, I believe. If she's actually a for-realz-caster ima custom the heck outa this :smiling_imp:

ibm2431:
In this instance, it's the NM Siren, not the summoned avatar:
siren

ibm2431:
"Hey self, don't you put 'merge ready' labels on stuff you approve?"
"Why yes self, I do. Why do you ask?"
"Do you not approve your own work?"
"Why yes, I do!"

zach2good:
LGTM 👍

cocosolos:
Looks good!

PR #596, opened 2020-05-08
Improvements for xiloader error reporting

Kreidos:
Topaz previously only used one account creation status code for any and all errors, which was reported in xiloader as "Username is taken" in all cases. Many times, this was not true; particularly if account_creation is set false in the config, or if an SQL error had occurred server-side.

This PR adds 2 new return codes, and corrects their usage. This new verbosity is dependent on using the new xiloader, but it is 100% backwards compatible; old versions can still work normally with their existing reduced clarity in error reporting.

This serves to make the account-creation flag a wholly complete and properly handled feature.

**This PR should probably be held until my corresponding [PR for improvements in xiloader](https://github.com/zircon-tpl/xiloader/pull/2) is merged.**

**_I affirm:_**


cocosolos (approved):

ibm2431 (approved):
Looks legit!

PR #601, opened 2020-05-09
Added missing zone to Tidal Talisman

Noikar:
**_I affirm:_**

You couldn't use it in Port Jeuno before.


59blargedy (approved):

ibm2431 (approved):
Thanks, Xephera!

PR #607, opened 2020-05-10
Full Speed Ahead + Packet 0x3A + Packet 0x75 + Packet 0x77

### LINKED ISSUE #603: Full Speed Ahead! not starting correctly.
**_I have:_**

After obtaining the Chocobo License, and the Map of the Jeuno Area, talking to Mapitoto plays the correct cutscene, and does dump the PC in Batallia Downs, but the PC is NOT mounted on the Raptor. The Markers for the quest are still in the map, however.

EDIT:
Returning to Mapitoto causes the cutscene to replay, and the PC to be dumped back out into Batallia Downs.

zach2good:
**_I affirm:_**

**Rewritten Ticket 2: Electric Boogaloo:**
I've bounced around a bunch from this being a single timer packet, to the full quest, to just two packets, to the whole quest. Now it's the whole quest, a bunch of bindings and 3 new packets. Oops.

Everything is kept inside the `full_speed_ahead.lua` helper. It is ticked by a `FULL_SPEED_AHEAD` status, which wraps `MOUNTED` status and the other logic. 

There is a binding for `onPlayerEmote`. 

Everything is tracked with localVars and a single charVar for status between zones. It doesn't look like this appears in the quest log and I can't see any caps or packets to confirm otherwise.

Fixes #603


src/map/packets/independant_animation.cpp
```diff
@@ -0,0 +1,25 @@
...
+	this->type = 0x3A;
+	this->size = 0x14;
+
+	if (PEntity)
+	{
+        ref<uint32>(0x04) = PEntity->id;
+        ref<uint32>(0x08) = PEntity->id;
+
+        ref<uint16>(0x0C) = PEntity->targid;
+        ref<uint16>(0x0E) = PEntity->targid;
```
cocosolos: Looks like this should take a target and pass that as `0x08` and `0x0E`. See [Windower docs](https://github.com/Windower/Lua/blob/308ec1d44e94250477db634a8a0552d7e687a88a/addons/libs/packets/fields.lua#L2123-L2131).

scripts/zones/Batallia_Downs/Zone.lua
```diff
@@ -28,27 +46,34 @@ function onZoneIn( player, prevZone)
...
     if (triggerLightCutscene(player)) then -- Quest: I Can Hear A Rainbow
         cs = 901;
-    elseif (player:getCurrentMission(WINDURST) == tpz.mission.id.windurst.VAIN and player:getCharVar("MissionStatus") ==1) then
+    elseif (player:getCurrentMission(WINDURST) == tpz.mission.id.windurst.VAIN and player:getCharVar("MissionStatus") == 1) then
         cs = 903;
     end
 
+    if player:getCharVar("[QUEST]FullSpeedAhead") == 1 then
+        player:addStatusEffect(tpz.effect.FULL_SPEED_AHEAD, 0, 2, tpz.fsa.duration)
+    end
```
cocosolos: Should this be first thing in this function and include a `return`? So that the other cutscenes don't fire when you zone in for this quest.

src/map/packets/timer_bar_util.cpp
```diff
@@ -0,0 +1,44 @@
...
+    ref<uint16>(0x04) = 0xFFFF;
+
+    ref<uint32>(0x08) = CVanaTime::getInstance()->getVanaTime();
+    ref<uint32>(0x0C) = seconds;
+
+    // Static data ("_unknown1" in Windower fields)
+    ref<uint8>(0x10) = 0x3C;
+    ref<uint8>(0x24) = 0x03;
+}
+
```
cocosolos, line 25: Can you add another function here to set the [battlefield and render radius](https://github.com/Windower/Lua/blob/308ec1d44e94250477db634a8a0552d7e687a88a/addons/libs/packets/fields.lua#L3079-L3080) at `0x1C` and `0x20` just so it's documented and done?

cocosolos (changes requested):

cocosolos (approved):

PR #610, opened 2020-05-11
Remove RUN flag from Vhana Ehgaklywha in Jeuno.

tankfest:
**_I affirm:_**

Vhana now walks (as intended) smoothly between lamps in Lower Jeuno.


kaincenteno:
is this the same person that walks outside jeuno xD?

tankfest:
It's the npc that walks in lower jeuno, lighting the lamps if no one does it before the timer.

PR #615, opened 2020-05-13
Minor settings.lua update.

ghost:
**_I affirm:_**

Hello.

Enabled all expansions by default and turned the CoP areas level cap off to be retail accurate.

Thanks!

### DIFF
--- a/scripts/globals/settings.lua
+++ b/scripts/globals/settings.lua
@@ -71,7 +71,7 @@ DARK_POWER      = 1.000 -- Multiplies amount drained by Dark Magic.
 ITEM_POWER      = 1.000 -- Multiplies the effect of items such as Potions and Ethers.
 WEAPON_SKILL_POWER  = 1.000 -- Multiplies damage dealt by Weapon Skills.
 WEAPON_SKILL_POINTS = 1.000 -- Multiplies points earned during weapon unlocking.
-USE_ADOULIN_WEAPON_SKILL_CHANGES = false -- true/false. Change to toggle new Adoulin weapon skill damage calculations
+USE_ADOULIN_WEAPON_SKILL_CHANGES = true -- true/false. Change to toggle new Adoulin weapon skill damage calculations
 
 HARVESTING_BREAK_CHANCE = 33 -- % chance for the sickle to break during harvesting.  Set between 0 and 100.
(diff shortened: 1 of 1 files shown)


ibm2431 (approved):
Just because the missions aren't finished doesn't mean we should lock what is.

Kreidos (approved):
I agree. Everything should probably be unlocked by default.

kaincenteno:
can the crafting be increase to the previous default (2.6) as it's difficult to skill up with 1.0? 

I'm not sure if Weapon skill skillup also needs to be reverted back to its previous default (2.5)

I know we are targetting current retail, but with the current settings for skill ups becomes impossible ^^;; (unsure about weaponskill though but for sure for crafting)

TeoTwawki:
> can the crafting be increase to the previous default (2.6) as it's difficult to skill up with 1.0?
> 
> I'm not sure if Weapon skill skillup also needs to be reverted back to its previous default (2.5)
> 
> I know we are targetting current retail, but with the current settings for skill ups becomes impossible ^^;; (unsure about weaponskill though but for sure for crafting)

how about we fix the crafting to be retail accurate at 1.0 instead. 1.0 is ***supposed to be*** retail, and moving it back to 2.6  just hides the underlying problem. Like being told to clean your room and shoving everything under the bed before mom & dad come back. Nearly every server customizes this  - just change yours.

kaincenteno:
stop looking under my bed @TeoTwawki  :panda_face:

TeoTwawki:
@kain the monster that lives there is complaining, says to get a storage locker instead of using his home for all your stuff.

Kreidos:
If we want 1.0 to be retail, we should in fact change the defaults in map.cpp to 1.0 as well; looks like whoever changed it in the default conf missed changing it there.

TeoTwawki:
> If we want 1.0 to be retail, we should in fact change the defaults in map.cpp to 1.0 as well; looks like whoever changed it in the default conf missed changing it there.

<.<  
\>.>  
~~You can't prove nothin~~ _git log rats Teo out_
You'll neva take me alive coppa!
:running_man: :cloud: :cloud: :cloud: 

.  
.  
.  
The synth utils does need at least a partial rewrite though. Has for some time.

ibm2431 (commented):
While you're here, Nyu:
`USE_ADOULIN_WEAPON_SKILL_CHANGES` is currently defaulting to false

PR #616, opened 2020-05-13
Missed a tpz.casket_loot. from line 493

Omnione:
**_I affirm:_**


ibm2431 (approved):
Kind of an inarguable fix, so I'm _breaking the two-staff rule_~!

PR #617, opened 2020-05-13
Fix some ENABLE settings

ibm2431:
COP: `cop-log` branch will change the starting ID from zero
WOTG: Just making its usage consistent with other ENABLE settings
SOA: 0 evaluates as true in lua

**_I affirm:_**


TeoTwawki (approved):
_I press button!_

ghost (approved):

PR #619, opened 2020-05-14
Ensure Enmity is Applied

brianmask:
**_I affirm:_**

Easy one liner...

Enmity generating trust skills currently are not generating enmity towards the user.  This is due to a missing call to state.ApplyEnmity().  Added the missing line to trustentity.cpp -> OnAbility.

A page of old commented out BST logic buried this call in charentity.cpp OnAbility which is what trustentify OnAbility seems to be based on.


ibm2431 (approved):
Thanks, brianmask!

cocosolos (approved):

PR #628, opened 2020-05-16
Fix mission ID when none accepted

cocosolos:
Current mission ID was being truncated to 8 bits in core previously, but now uses the full 16 bits available. Not having a current mission sets current mission ID to `-1` or `65535` which was being truncated to `255` previously. This should be `65535` now.

**_I affirm:_**


TeoTwawki:
[`!checkmission` also needs adjustments to account for this.](https://github.com/project-topaz/topaz/blob/release/scripts/commands/checkmission.lua#L49)

cocosolos:
Thanks, fixed!

PR #634, opened 2020-05-17
Table some shared mission IDs

cocosolos:
Tabled shared mission IDs, updated scripts to use these instead of hard coded numbers. Think I got them all.

**_I affirm:_**


scripts/globals/missions.lua
```diff
@@ -53,6 +53,17 @@ tpz.mission.logEx =
 
 tpz.mission.id =
 {
+    -----------------------------------
+    --  All Nations
+    -----------------------------------
+    ['nation'] =
+    {
+        MAGICITE    = 13,
+        VAHZL       = 14,
```
TeoTwawki: VAHZL? How about ARCHLICH so everyone knows what it is at a glance.

I see Vahzl I think of promy vahzl or teleport vahzl, the archlich 5-1 nation mission does not at all come to mind. You say Archlich I immediately know what you mean.

cocosolos: Yep you're totally right idk what I was thinking there, fixing now.

ghost (approved):

ibm2431 (approved):

PR #635, opened 2020-05-17
[FIX] Trusts are considered inside battlefields & instances

zach2good:
**_I affirm:_**

Got an interesting bug report from Kain that Trusts are acting strangely in the Shadow Lord fight, after some investigation:
```
Ok, I've done a full rebuild and db update from canary, indeed: shadow lord fight: Kupipi freaks out:
- When she cast Protectra it doesn't land on the player, which means she's constantly trying to cast it (which is her logic if someone in the party is missing protect/shell)
- When she runs out of MP her casting particle effects stay on for a while (but this is probably the same as retail, look at when a battle ends and their casting is "interrupted", the effect stays on even through the casting ends)
- I think BCNM with Trusts might be using player allegiance in strange ways, causing some issues
Well, my hunch seems correct, there are a bunch of things in the battlefield handler that deal with PC/MOB/NPC/PET
but not trust
Which means  their allegiance is never set
```

@cocosolos spotted it before me, there's a bunch of information that isn't being set on spawnTrust. I'm surprised we made it this far without noticing...

Also a little misc cleanup.


kaincenteno:
Thanks for your fixing this so quick @zach2good  :cat:

src/map/entities/charentity.cpp
```diff
@@ -1443,6 +1443,14 @@ bool CCharEntity::IsMobOwner(CBattleEntity* PBattleTarget)
         return true;
     }
 
+    for (auto& trust : PTrusts)
+    {
+        if (PBattleTarget->m_OwnerID.id == trust->id)
```
cocosolos: Is there any mechanism setting the trust as the mob owner? AFAIK right now only a player can be the owner of a mob, and this ID is used in packets so it probably shouldn't be set to a trust. This branch could really use a merge of the claim code.

src/map/battlefield.cpp
```diff
@@ -307,7 +308,7 @@ bool CBattlefield::InsertEntity(CBaseEntity* PEntity, bool enter, BATTLEFIELDMOB
         PEntity->loc.zone->PushPacket(PEntity, CHAR_INRANGE, new CEntityUpdatePacket(PEntity, ENTITY_SPAWN, UPDATE_ALL_MOB));
         m_NpcList.push_back(static_cast<CNpcEntity*>(PEntity));
     }
-    else if (PEntity->objtype == TYPE_MOB || PEntity->objtype == TYPE_PET)
+    else if (PEntity->objtype == TYPE_MOB || PEntity->objtype == TYPE_PET || PEntity->objtype == TYPE_TRUST)
```
cocosolos: Is the intent to treat trusts like battlefield allies? Right now I think allies are only instanced mobs that help during the fight, like Prishe in the Promathia fight. Player pets aren't stored in the battlefield object at all. I would probably remove both of these additions and change [this line](https://github.com/project-topaz/topaz/blob/7bc656841594711c48f8bbef80e3f66cbc7670ee/src/map/battlefield.cpp#L392) to handle trusts as well.

src/map/entities/charentity.cpp
TeoTwawki: should be the trusts "master" like with pets, even if they do appear on the party roster visually they aren't players and we need to be cautious what we send their way.

(where/how are pets even being handled for owner purposes in battlefields??)

zach2good: So, not in this PR, but I wanted to sneak this in ahead of my _next_ PR. In the next PR Trusts are using full player abilities with all the trimmings. When Naji 'vokes the player would be locked off of the mob without this snippet.

cocosolos: Anything other than players should never be set as the owner of a mob. Instead, it should be setting the player as the owner when the trust acts, avoiding this problem completely. This stuff is all centralized in the claim code that needs to get merged into this branch.

cocosolos (approved):

ibm2431 (approved):

PR #638, opened 2020-05-18
[FIX] Remove 1 day wait for Ranperre's Final Rest quest

MarkWaldron:
**_I affirm:_**

I've removed the 1 day check and CharVar, and added a zone requirement. 

Closes issue: #245

### DIFF
--- a/scripts/zones/Chateau_dOraguille/npcs/_6h0.lua
+++ b/scripts/zones/Chateau_dOraguille/npcs/_6h0.lua
@@ -24,7 +24,6 @@ function onTrigger(player,npc)
     local MissionStatus = player:getCharVar("MissionStatus");
     local infiltrateDavoi = player:hasCompletedMission(SANDORIA,tpz.mission.id.sandoria.INFILTRATE_DAVOI);
 
-    local Wait1DayRanperre = player:getCharVar("Wait1DayForRanperre_date");
     local osdate = tonumber(os.date("%j"));
 
     if (player:getCharVar("aBoysDreamCS") == 8) then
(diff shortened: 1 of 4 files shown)


scripts/zones/Northern_San_dOria/npcs/Grilau.lua
```diff
@@ -62,7 +62,7 @@ function onTrigger(player,npc)
             player:startEvent(1000); -- Start First Mission "Smash the Orcish scouts"
         elseif (player:hasKeyItem(tpz.ki.ANCIENT_SANDORIAN_BOOK)) then
             player:startEvent(1035);
-        elseif (CurrentMission == tpz.mission.id.sandoria.RANPERRE_S_FINAL_REST and player:getCharVar("MissionStatus",4) and tonumber(os.date("%j")) == player:getCharVar("Wait1DayForRanperre_date")) then
+        elseif (CurrentMission == tpz.mission.id.sandoria.RANPERRE_S_FINAL_REST and player:getCharVar("MissionStatus",4) and player:needToZone()) then
```
cocosolos, line 65: Should this be checking `not player:needToZone()`? I'm not sure what these 3 CS's are, but it doesn't look like they're executing correctly and setting the `MissionStatus` var correctly. Everything up until this part looks right.

scripts/zones/Chateau_dOraguille/npcs/_6h0.lua
```diff
@@ -41,7 +39,7 @@ function onTrigger(player,npc)
         player:startEvent(547);
     elseif (currentMission == tpz.mission.id.sandoria.RANPERRE_S_FINAL_REST and MissionStatus == 0) then
         player:startEvent(81);
-    elseif (currentMission == tpz.mission.id.sandoria.RANPERRE_S_FINAL_REST and MissionStatus == 4 and Wait1DayRanperre ~= osdate) then -- Ready now.
+    elseif (currentMission == tpz.mission.id.sandoria.RANPERRE_S_FINAL_REST and MissionStatus == 4 and player:needToZone() == false) then -- Ready now.
         player:startEvent(21);
```
cocosolos, line 43: I don't think either of these lines are required anymore, once the `MissionStatus` var is correctly being set by the guard.

scripts/zones/Northern_San_dOria/npcs/Grilau.lua
TeoTwawki: want to avoid needToZone anyway since its global. we can use set/getLocalVar to tell is player zoned since the preceding step..

zircon-tpl:
Hello, MarkWaldron. Thank you for your interest in Project Topaz!

Another Contributor had submitted a Pull Request which addresses the quest in question. That Pull Request was merged, and I have been informed by Staff that it would make the changes in this Pull Request redundant. Therefore I must unfortunately close this Pull Request.

Should you feel that this determination is in error, please do not be afraid to say so!

I apologize on behalf of the project, and hope that this does not discourage you from further contributions.

Issue #484, opened 2020-04-10
Eco-Warrior Windurst quest exploit


tankfest:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

After following the quest as intended and receiving the KI from the ??? after killing flies, if you warp and directly talk to Lumomo for your reward instead of Ahko, your char variable for having killed the NMs never resets, so next week you do this you will just keep getting the KI without having to fight.  We fixed this on our end by moving player:setCharVar("ECOR_WAR_WIN-NMs_killed", 0) from Ahko to Lumomo when you receive the reward.

ibm2431:
#639

PR #639, opened 2020-05-18
[FIX] Set CharVar for EcoWarrior NMs Killed to 0 after completing quest

MarkWaldron:
**_I affirm:_**

As per issue, an exploit existed where you could complete the Ecowarrior quest without having a CharVar reset, allowing you to just complete it each week without having to defeat the monster. Now, on successful completion of the question, the CharVar is reset. 

Closes issue: #484


59blargedy (approved):
ECOR vs ECO is driving me nuts but not your error XD - thanks for submitting!!!

ghost (approved):

PR #645, opened 2020-05-21
Hunt System

### LINKED ISSUE #240: hunt regimes and dominion ops
_Thursday Oct 11, 2018 at 16:08 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 5302_

----

**_I have:_**

n/a

n/a

These can be easily scripted right now. 

Hunt Regimes are very similar to Fields of Valor but the targets are NMs and the rewards are evoliths instead of exp. This is separate from Elite FoV pages. The evolith item carries augments on it as if it were equipment. The only part we can't do is synergy to transfer the evolith's augment data onto equipment **because synergy doesn't exist yet.**

Dominion ops pages are similar to Fields of Valor for Abyssea zones. We don't yet have regional influence for dominion op zones and the mobs themselves aren't retail accurate\* yet but for testing the mob only needs to spawn and die properly and that already works 

\* (retail abyssea mobs have level changes when killed - the quicker you mow them down the stronger they get, and will delevel themselves back to normal once left idle for a time).

Apocarypse:
**_I affirm:_**

Functional hunt system implementation. The only thing missing is Evoliths, which require synergy to be working. The VNMs also do not have scripts, as they are unimplemented at this time.

Please do not squash when merging this PR as there are collaborative commits which should remain separate.

I would like to thank Kreidos for all his help on this project. He was there to assist every step of the way and definitely deserves half the credit.

Fixes #240

### DIFF
--- a/scripts/globals/hunts.lua
+++ b/scripts/globals/hunts.lua
@@ -1,0 +1,1616 @@
+-----------------------------------------------------
+-- Hunt Registry
+
+-- https://www.bg-wiki.com/bg/Hunts
+-- https://ffxiclopedia.fandom.com/wiki/Hunt_Registry
+-- https://www.bg-wiki.com/bg/Hunt_Registry
+-----------------------------------------------------
+require("scripts/globals/zone")
+require("scripts/globals/msg")
+require("scripts/globals/regimes")
+
+tpz = tpz or {}
+tpz.hunts = tpz.hunts or {}
+
+local zone =
+{
+   [tpz.zone.RULUDE_GARDENS] =
+   {
+      msg =
+      {
+        6869, -- Hunt accepted!
+        6870, -- You use <fee> scylds. Scyld balance: <scylds>.
+        6887, -- Hunt canceled.
+        6851, -- Current training regime canceled.
+        6881, -- You record your hunt.
+        6883, -- You obtain <scylds>! Current Balance: <scylds>.
+      },
+      -- these params display the correct number of hunts
+      huntMenu =
+      {
+        [   1] =    2031616,
+        [   9] = 2147483616,
+        [  17] = 2147483616,
+        [  25] = 2147483616,
+        [  33] = 2147483616,
+        [  41] = 2147483616,
+        [  49] = 2147483616,
+        [  57] = 2147483632,
+        [  65] = 2147483520,
+        [  73] = 2147483616,
+        [  81] = 2147483632,
+        [  89] = 2147483632,
+        [  97] = 2147483632,
+        [ 105] = 2147483616,
+        [ 113] = 2147483616,
+        [ 121] = 2147483392,
+      },
+        -- these params display info after hunt selection
+        -- Qufim Island
+        [ 266] = { huntId =  52531, optionId = 2457 },
+        [ 522] = { huntId =  69940, optionId = 2465 },
+        [ 778] = { huntId =  86325, optionId = 2473 },
+        [1034] = { huntId = 121142, optionId = 2481 },
+        -- Beaucedine Glacier
+        [ 274] = { huntId =  69943, optionId = 2489 },
+        [ 530] = { huntId =  86328, optionId = 2497 },
+        [ 786] = { huntId =  88377, optionId = 2505 },
+        [1042] = { huntId = 105786, optionId = 2513 },
+        -- Xarcabard
+        [ 282] = { huntId =  86331, optionId = 2521 },
+        [ 538] = { huntId =  86332, optionId = 2529 },
+        [ 794] = { huntId = 104765, optionId = 2537 },
+        [1050] = { huntId = 122174, optionId = 2545 },
+        -- Uleguerand Range
+        [ 290] = { huntId = 120127, optionId = 2553 },
+        [ 546] = { huntId = 121152, optionId = 2561 },
+        [ 802] = { huntId = 139585, optionId = 2569 },
+        [1058] = { huntId = 158018, optionId = 2577 },
+        -- The Sanctuary of Zi'tah
+        [ 298] = { huntId =  86339, optionId = 2585 },
+        [ 554] = { huntId = 103748, optionId = 2593 },
+        [ 810] = { huntId = 105797, optionId = 2601 },
+        [1066] = { huntId = 105798, optionId = 2609 },
+        -- Ro'Maeve
+        [ 306] = { huntId = 120135, optionId = 2617 },
+        [ 562] = { huntId = 122184, optionId = 2625 },
+        [ 818] = { huntId = 121161, optionId = 2633 },
+        [1074] = { huntId = 139594, optionId = 2641 },
+        -- Upper Delkfutt's Tower
+        [ 314] = { huntId =  69963, optionId = 2649 },
+        [ 570] = { huntId = 120140, optionId = 2657 },
+        [ 826] = { huntId = 155981, optionId = 2665 },
+        -- Middle Delkfutt's Tower
+        [ 322] = { huntId =  52558, optionId = 2673 },
+        [ 578] = { huntId =  52559, optionId = 2681 },
+        [ 834] = { huntId =  52560, optionId = 2689 },
+        [1090] = { huntId =  52561, optionId = 2697 },
+        [1346] = { huntId =  52562, optionId = 2705 },
+        [1602] = { huntId =  52563, optionId = 2713 },
+        -- Lower Delkfutt's Tower
+        [ 330] = { huntId =  52564, optionId = 2721 },
+        [ 586] = { huntId =  52565, optionId = 2729 },
+        [ 842] = { huntId =  52566, optionId = 2737 },
+        [1098] = { huntId =  68951, optionId = 2745 },
+        -- Ranguemont Pass
+        [ 338] = { huntId =  69976, optionId = 2753 },
+        [ 594] = { huntId =  86361, optionId = 2761 },
+        [ 850] = { huntId = 122202, optionId = 2769 },
+        -- Fei'Yin
+        [ 346] = { huntId = 103771, optionId = 2777 },
+        [ 602] = { huntId = 121180, optionId = 2785 },
+        [ 858] = { huntId = 140637, optionId = 2793 },
+        -- Castle Zvahl Baileys
+        [ 354] = { huntId = 103774, optionId = 2801 },
+        [ 610] = { huntId = 121183, optionId = 2809 },
+        [ 866] = { huntId = 122208, optionId = 2817 },
+        -- Castle Zvahl Keep
+        [ 362] = { huntId = 120161, optionId = 2825 },
+        [ 618] = { huntId = 120162, optionId = 2833 },
+        [ 874] = { huntId = 120163, optionId = 2841 },
+        [1130] = { huntId = 120164, optionId = 2849 },
+        -- The Boyahda Tree
+        [ 370] = { huntId = 104805, optionId = 2857 },
+        [ 626] = { huntId = 122214, optionId = 2865 },
+        [ 874] = { huntId = 122215, optionId = 2873 },
+        [1130] = { huntId = 139624, optionId = 2881 },
+        -- Habitat Unknown (Ru'lude Gardens)
+        [ 378] = { huntId = 157226, optionId = 2889 },
+        [ 634] = { huntId = 157227, optionId = 2897 },
+        [ 890] = { huntId = 157229, optionId = 2905 },
+        [1146] = { huntId = 157230, optionId = 2913 },
+        [ 634] = { huntId = 174633, optionId = 2921 },
+        [ 890] = { huntId = 174636, optionId = 2929 },
+        [1146] = { huntId = 174639, optionId = 2937 },
+      -- actual data for each hunt
+      hunt =
+      {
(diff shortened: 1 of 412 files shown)


scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1616 @@
...
+function tpz.hunts.onTrigger(player,npc,event)
+  local huntZone = player:getCharVar("[hunt]zone")
+  local huntId = player:getCharVar("[hunt]id")
+  local huntStatus = player:getCharVar("[hunt]status")
+  local scyldBits = bit.lshift(player:getCurrency("scyld"), 14)
+  local lockBit = bit.lshift(1, 24)
+  local registryZone = zone[player:getZoneID()]
+
+  -- one vana'diel day lockout timer after completing a hunt
+  if (VanadielYear() * 360) + VanadielDayOfTheYear() < player:getCharVar("[hunt]nextHunt") then
```
cocosolos: Could use `if os.time() < player:getCharVar("[hunt]nextHunt") then`

scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1616 @@
...
+
+  -- cancels hunt
+  elseif option == 3 then
+    player:messageSpecial(msg[3])
+    player:setCharVar("[hunt]status", 0)
+
+  -- cancels training regime and clears all vars
+  elseif option == 4 then
+    player:messageSpecial(msg[4])
+    tpz.regime.clearPlayerVars(player)
```
cocosolos: Was this changed to `clearRegimeVars`?

scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1616 @@
...
+  elseif option == 5 then
+    local huntZone = player:getCharVar("[hunt]zone")
+    local huntId = player:getCharVar("[hunt]id")
+    local scyldBounty = zone[huntZone].hunt[huntId].bounty
+
+    player:messageSpecial(msg[5])
+    player:messageSpecial(msg[6], scyldBounty, scylds)
+    player:addCurrency("scyld", scyldBounty)
+    -- give player evoliths here
+    player:setCharVar("[hunt]nextHunt", (VanadielYear() * 360) + (VanadielDayOfTheYear() + 1))
```
cocosolos: Could use `player:setCharVar("[hunt]nextHunt", getVanaMidnight())`

scripts/globals/hunts.lua
Apocarypse: how about `if player:getCharVar("[hunt]nextHunt") < getVanaMidnight()then`

scripts/globals/hunts.lua
Apocarypse: Yes! I'm glad you caught that.

scripts/globals/hunts.lua
Kreidos: I like this idea. Ideally we should compare against a consistent source.

scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1616 @@
...
+function tpz.hunts.onTrigger(player,npc,event)
+  local huntZone = player:getCharVar("[hunt]zone")
+  local huntId = player:getCharVar("[hunt]id")
+  local huntStatus = player:getCharVar("[hunt]status")
+  local scyldBits = bit.lshift(player:getCurrency("scyld"), 14)
+  local lockBit = bit.lshift(1, 24)
+  local registryZone = zone[player:getZoneID()]
+
+  -- one vana'diel day lockout timer after completing a hunt
+  if player:getCharVar("[hunt]nextHunt") < getVanaMidnight() then
```
cocosolos: This should be checking `os.time()`, reason being that `getVanaMidnight()` very slightly varies in what it returns, even for the same day, so if you're lucky then `getVanaMidnight` will return a midnight that's maybe 1s before the `getVanaMidnight` that is saved in the char var. This should check `if os.time() < player:getCharVar("[hunt]nextHunt") then` (it's also checking the requirement backwards, this block should execute if the char is blocked)

Kreidos: I have a question. Since os.time() is lua native it always returns the system time; whereas getVanaMidnight ties into the time calculations and so could be affected by any offsets which are (or might) be implemented in the core. Couldn't that pose an issue to using os.time() which is completely independent of server time calculations?

scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1616 @@
...
+        [ 378] = { huntId = 157220, optionId = 2001 },
+        [ 634] = { huntId = 157221, optionId = 2009 },
+        [ 890] = { huntId = 174627, optionId = 2017 },
+
+      hunt =
+      {
+        [1577] = { pageId = 197, mobId = 17211561, bounty =  5, fee =   0 }, -- Stinging Sophie
+        [1585] = { pageId = 198, mobId = 17211714, bounty =  5, fee =   0 }, -- Maighdean Uaine
+        [1593] = { pageId = 199, mobId = 17211666, bounty =  5, fee =   0 }, -- Bedrock Barry
+        [1601] = { pageId = 200, mobId = 17215888, bounty =  5, fee =   0 }, -- Leaping Lizzy
```
cocosolos, line 466: I'm not sure if this is very common, but I happened to test this on Lizzy and she actually has 2 IDs, so might need to add the other one `17215868`.

scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1616 @@
...
+
+    player:messageSpecial(msg[5])
+    player:messageSpecial(msg[6], scyldBounty, scylds)
+    player:addCurrency("scyld", scyldBounty)
+    -- give player evoliths here
+    player:setCharVar("[hunt]nextHunt", getVanaMidnight())
+    tpz.hunts.clearHuntVars(player)
+
+    -- scylds cap at 1000
+    if player:getCurrency("scyld") + bounty > 1000 then
```
cocosolos: `bounty` is nil, is this supposed to be `scyldBounty`?

scripts/globals/hunts.lua
Apocarypse: I asked about this and got no reply unfortunately. There are some NMs that have a single spawn point under one ID, and like 50 spawn points under another ID. In those cases, I opted for the 50. I don't know what the logic is there exactly.

scripts/globals/hunts.lua
Apocarypse: Yes, I think you're right. I'll change it.

scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1616 @@
...
+
+  -- completes hunt
+  elseif option == 5 then
+    local huntZone = player:getCharVar("[hunt]zone")
+    local huntId = player:getCharVar("[hunt]id")
+    local scyldBounty = zone[huntZone].hunt[huntId].bounty
+
+    player:messageSpecial(msg[5])
+    player:messageSpecial(msg[6], scyldBounty, scylds)
+    player:addCurrency("scyld", scyldBounty)
```
cocosolos: Adding currency here and below?

scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1616 @@
...
+    tpz.regime.clearRegimeVars(player)
+
+  -- completes hunt
+  elseif option == 5 then
+    local huntZone = player:getCharVar("[hunt]zone")
+    local huntId = player:getCharVar("[hunt]id")
+    local scyldBounty = zone[huntZone].hunt[huntId].bounty
+
+    player:messageSpecial(msg[5])
+    player:messageSpecial(msg[6], scyldBounty, scylds)
```
cocosolos: This message should be moved to after you distribute the currency, then use the updated total.

scripts/globals/hunts.lua
cocosolos: It's still using getVanaMidnight to set the time, so that's all accounted for. This check is just seeing if the current time is past what has been determined is the next vana midnight when that var was set.

scripts/globals/hunts.lua
```diff
@@ -1586,19 +1586,19 @@ if huntEntry then
...
-    if player:getCurrency("scyld") + bounty > 1000 then
+    if player:getCurrency("scyld") + scyldBounty > 1000 then
        player:setCurrency("scyld", 1000)
+       player:messageSpecial(msg[5])
+       player:messageSpecial(msg[6], scyldBounty, scylds)
     else
-       player:addCurrency("scyld", bounty)
+       player:addCurrency("scyld", scyldBounty)
+       player:messageSpecial(msg[5])
+       player:messageSpecial(msg[6], scyldBounty, scylds)
```
cocosolos: This will still give the old total, go with this:
```lua
    -- scylds cap at 1000
    if player:getCurrency("scyld") + scyldBounty > 1000 then
       player:setCurrency("scyld", 1000)
    else
       player:addCurrency("scyld", scyldBounty)
    end
    player:messageSpecial(msg[5])
    player:messageSpecial(msg[6], scyldBounty, player:getCurrency("scyld"))
```

scripts/globals/hunts.lua
```diff
@@ -0,0 +1,1615 @@
...
+function tpz.hunts.onEventFinish(player, csid, option)
+  local scylds = player:getCurrency("scyld")
+  local registryZone = zone[player:getZoneID()]
+  local huntEntry = registryZone.hunt[option]
+  local msg = registryZone.msg
+
+-- accepting hunt
+if huntEntry then
+    player:messageSpecial(msg[1])
+    player:messageSpecial(msg[2], huntEntry.fee, scylds)
```
cocosolos: These should be moved to after the fee is applied, and updated with `player:getCurrency("scyld")`

cocosolos (approved):
Tried out a few hunts, everything looks like it's working as it should. Good job!

scripts/globals/hunts.lua
TeoTwawki: ~~tabling all the IDs of all these NMs is the wrong approach. Followup what all existing regimes do and include the regime ID at the mob, then match on regime ID instead of performing a lookup to see if that specific mob is in a table. We don't need to keep track of them all because its never going to change, and you already have that line there anyway. model it after the existing regimes instead of trying to centralize every bit of data to one place as this is a case where that just gets in the way.~~

~~`tpz.hunts.checkHunt(mob, player, isKiller)`
compare with
`tpz.regime.checkRegime(player, mob, 80, 1, tpz.regime.type.FIELDS)`
and
`tpz.regime.checkRegime(player, mob, 640, 2, tpz.regime.type.GROUNDS)`
you're trying to reinvent the wheel to separate hunts from these (understandable, a lot of code doesn't apply here) but missed why the regime ID was included.~~

now looking at this:
`tpz.hunts.checkHunt(mob, player, isKiller)`
We should not need isKiller at all as in retail all member who have the hunt regime active get credit and its not going to stack up (onMobDeath code will run once-per-member), so I suggest just go with:
`tpz.hunts.checkHunt(mob, player, RegimeIDhere)` to minimize your changes.

Edit: this could have been worded better by me, see coco's post which got straight to what it'd look like and didn't risk misunderstandings

TeoTwawki: p.s. if you're thinking "but then we need to keep track of which NMs go in which regime!" no, because the regime IDs will never shift. set once done for life. :)

Kreidos: I gotta disagree here on part of that. isKiller isn't necessary, you're right, but frankly I think having all the mobs tabled here with a simple lookup to a table is a much faster and better method than breaking out every ID into separate mob scripts. Regimes takes the ID it gets and combs through a table using a for loop anyway to find the mob which is far from elegant. This solution involves one direct check to one table and I gotta go with the performance method here.

Kreidos: I think this works well, and if you believe you have a better solution I welcome you to PR the files after this merges. :)

TeoTwawki: yeah no, I'm no opening a competing PR just because you prefer listing every NMs ID into the table and then trying to make workaround for the inherent shortcomings of that approach. 

A given NM can have multiple IDs. at a minimum would would need multiple ID entries. There is no reason we could not match on regime ID instead of mobID, loop or no loop.

cocosolos: Switching to `tpz.hunts.checkHunt(mob, player, RegimeIDhere)` would actually solve the multiple ID mobs issue all at once and would require minimal changes. `checkHunt` becomes:
```lua
    function tpz.hunts.checkHunt(mob, player, RegimeIDhere)
        local huntZone = player:getCharVar("[hunt]zone")
        local huntId = player:getCharVar("[hunt]id")
        local playerHunt = zone[huntZone].hunt[huntId]

        -- required NM has been defeated
        if player:getCharVar("[hunt]status") == 1 and huntId == RegimeIDhere then
           player:messageBasic(tpz.msg.basic.FOV_DEFEATED_TARGET + 20)
           player:setCharVar("[hunt]status", 2)
        end
    end
```
but that would also require removing all the mob IDs from the tables and updating all the mob files as well.

TeoTwawki: @cocosolos that is what I was going for. It does suck to re-edit all those mob scripts, but they all did have to have an edit to start with so its not like these didn't have code going to them. That said I will volunteer to make the mob script changes if apoc wants to take this road.

the alternative is harder - alterations to support multiple mobID elements and then the inevitable future problems that come with mobIDs later


((if anyone can come up with a 3rd option, I'm all ears)

PR #649, opened 2020-05-22
Take zone and distance into account for Trust Gambits

zach2good:
**_I affirm:_**

From Discord:
```
Galieon (wiseowl)
we're testing out the initial trusts over on Dawnbreak and found a bug with kupipi
if the party member is not in range, out of zone, etc, her AI spams rotation of Protectra/Shellra and nothing else
```

Nice catch, all other gambits rely on the fact that a trust cannot exist outside of the same zone as its master. This selector views the whole party, regardless of the zone (or distance).

This should fix that. I'll test this tomorrow, and possibly put a 60-second cooldown on Kupipi's Protectra/Shellra, so if strange conditions happen she's able to do other things.

### DIFF
--- a/src/map/ai/helpers/gambits_container.cpp
+++ b/src/map/ai/helpers/gambits_container.cpp
@@ -121,7 +121,9 @@ void CGambitsContainer::Tick(time_point tick)
             for (uint8 i = 0; i < master->PParty->members.size(); ++i)
             {
                 auto member = master->PParty->members.at(i);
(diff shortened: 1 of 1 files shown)


kaincenteno:
This is also true if one of the party members is in another zone. For some reason when testing this, the hp/mp on the lower right listing all party members would show as if they are not in the same zone, showing only the name. dunno if someone could replicate this, or if its even related.

59blargedy (approved):

PR #651, opened 2020-05-23
[WIP] Refactor effort

m241dan:
**_I affirm:_**

This is a WIP PR. That being said, these commits will be cleaned up but I wanted to get this out the door so reviewers could start looking it over.

**New Lua System**
This code marks the possible beginning of a massive refactor effort for the Topaz project. It could be the model for all future lua code, including proper integration testing and changing the way that the Core invokes lua code.

This design heavily leverages lua table mechanics. The benefits of this is that the code can be written in a more clean and concise manner, it is easier to test, and it supports a module system for people that are interested in customization.

It also follows a C++ language core principle, where possible, of not paying for what you do not use. You will see that there are very few decision paths: ie, each function is responsible for one thing and one thing only. There is no, "if I'm this, do X, if I'm that, do Y." This design focuses on, "I'm me, and this is what I do" type design (super technical terminology).

This design will eventually be extended to weaponskills, blue magic, ninjutsu, summoning magic, avatar damage, puppet master damage, etc, etc. 

**C Interface Functions**
This method seems like it might be a bit extra, and it is definitely different than how the design currently looks (especially considering the design moves away from individual files for things like spells, weaponskills, items, etc, and uses lua tables instead). At first, you may think, "why not just call the spell directly from C++ instead of loading the interface function and THEN calling the spell." The reason to not do it that way is that it excludes people from writing modules that could adjust the process of how C interacts with Lua.

**Testing**
The current design of the testing rig can be run from the base director using lua directly. I have fabbed up libraries that it will use to simulate calls to the Core.

*lua scripts/test_suite/test_magic.lua* 

The tests themselves are very non-committal in terms of numbers. They mostly focus on relative numbers to validate that things are working logically and that there are no bugs in any of the paths. The reason for this non-committal approach is that so people who want to customize their code do not have to continually update tests to match their specific numbers just because they want AM2 to do more damage.

**Modules**
I have included a single module. The idea of the modules is that people who want to customize their code do not have to edit the main project code (potentially creating conflicts when updating, etc). The example module is FAR from a polished product. It is simply to show how the mechanics should/could work, but there should be some formalization process put towards how a Topaz module should look / be executed. For now, I've just hacked it in to show the potential of the new system.

**Work List (not comprehensive, yet...):**
 * [x] Outline basics of new lua code design / testing / modules
 * [ ] Create a unified pDIF table for both the Core and the Lua stuff. Core will load pDIF table from Lua so that modules can affect pDIF reliant code in the Core
 * [ ] Extend design to cover ...
     * [ ] physical weaponskills
     * [ ] magical weaponskills
     * [ ] "unique" weaponskills
     * [ ] enfeebles (magic)
     * [ ] enfeebles (songs)
     * [ ] buffs (white magic)
     * [ ] buffs (bard)
     * [ ] blue physical spells
     * [ ] blue magical spells
     * [ ] blue buffs
     * [ ] avatar damage
     * [ ] automatons
 * [ ] Back fill all the data (plans to automate this process, doing it by hand is insane)
     * [ ] elemental nukes
     * [ ] physical weaponskills
     * [ ] "unique" weaponskills
     * [ ] enfeebles (magic)
     * [ ] enfeebles (songs)
     * [ ] buffs (white magic)
     * [ ] buffs (bard)
     * [ ] blue physical spells
     * [ ] blue magical spells
     * [ ] blue buffs
     * [ ] avatar damage
     * [ ] automatons
 * [ ] Testing
     * [ ] elemental nukes
     * [ ] physical weaponskills
     * [ ] "unique" weaponskills
     * [ ] enfeebles (magic)
     * [ ] enfeebles (songs)
     * [ ] buffs (white magic)
     * [ ] buffs (bard)
     * [ ] blue physical spells
     * [ ] blue magical spells
     * [ ] blue buffs
     * [ ] avatar damage
     * [ ] automatons
 * [ ] Migrate spell.lua to replace magic.lua in order to unify the language. Anything from the "magic" menu should be in magic.lua. Same thing from weaponskill, etc, etc.
 * [ ] remove all hardcoded numbers in favor of "customizable" tables.

**Supporting Changes**
Things considered "final" commits are preprended with "PR"

In order...
 * Unified the language and mechanics of elemental resistances vs. specific damage taken
 * Added support for the nuke wall mechanic that exists on Notorious Monsters
 * Changed the way magic burst damage is handled due to certain portions being capped and others not.
 * Added support for the cumulative magic effects that -ja spells do.
 * Mod'ified Ebullience (for you Teo!)
 * Mod'ified Focalization (for you Teo!)
 * Unified Elemental Staff mod language to be more inline with how bg-wiki outlines it.


src/map/utils/battleutils.cpp
```diff
@@ -1206,7 +1206,7 @@ namespace battleutils
                     damage = std::clamp(damage,0,50);
                     damage += 10; //10~60
                     damage += WELL512::GetRandomNumber(8); //10~67 randomised
-                    damage += (float)damage * ((float)PDefender->getMod(Mod::WINDRES)/-100);
+                    damage += (float)damage * ((float)PDefender->getMod(Mod::WINDSDT)/-100);
```
m241dan, line 1209: ah... should be /1000...
Missed these...

m241dan: ah this is in a comment, which is why...

ShelbyZ (changes requested):
Easy top-level comment - newlines at the end of files

src/map/instance_loader.cpp
```diff
@@ -177,19 +177,19 @@ CInstance* CInstanceLoader::LoadInstance(CInstance* instance)
...
-
-            PMob->setModifier(Mod::FIRERES, (int16)((Sql_GetFloatData(SqlInstanceHandle, 39) - 1) * -100)); // These are stored as floating percentages
-            PMob->setModifier(Mod::ICERES, (int16)((Sql_GetFloatData(SqlInstanceHandle, 40) - 1) * -100)); // and need to be adjusted into modifier units.
-            PMob->setModifier(Mod::WINDRES, (int16)((Sql_GetFloatData(SqlInstanceHandle, 41) - 1) * -100)); // Higher RES = lower damage.
-            PMob->setModifier(Mod::EARTHRES, (int16)((Sql_GetFloatData(SqlInstanceHandle, 42) - 1) * -100)); // Negatives signify lower resist chance.
-            PMob->setModifier(Mod::THUNDERRES, (int16)((Sql_GetFloatData(SqlInstanceHandle, 43) - 1) * -100)); // Positives signify increased resist chance.
-            PMob->setModifier(Mod::WATERRES, (int16)((Sql_GetFloatData(SqlInstanceHandle, 44) - 1) * -100));
-            PMob->setModifier(Mod::LIGHTRES, (int16)((Sql_GetFloatData(SqlInstanceHandle, 45) - 1) * -100));
-            PMob->setModifier(Mod::DARKRES, (int16)((Sql_GetFloatData(SqlInstanceHandle, 46) - 1) * -100));
+            PMob->setModifier( Mod::SLASHSDT, (uint16)(Sql_GetFloatData( SqlInstanceHandle, 35) * 1000));
```
ShelbyZ, line 180: Spaces between identifiers feels like style creep from local settings

src/map/lua/luautils.cpp
```diff
@@ -2093,36 +2104,122 @@ namespace luautils
     {
         TPZ_DEBUG_BREAK_IF(PSpell == nullptr);
 
-        lua_prepscript(
-            PSpell->getSpellGroup() == SPELLGROUP_BLUE ? "scripts/globals/spells/bluemagic/%s.lua" :
-            PSpell->getSpellGroup() == SPELLGROUP_TRUST ? "scripts/globals/spells/trust/%s.lua" :
-            "scripts/globals/spells/%s.lua", PSpell->getName()
-        );
+        /**
```
ShelbyZ, line 2107: Why stop at a partial implementation (hack) and not force all to use the new system.  You get a broader range of spells to test out and find bugs related to them.  Assuming this system is to replace the old all spells will need to be tested at some point.

m241dan: Definitely will be. I stopped at partial because I didn't have a script to automate grabbing all the spell values and dumping them into a table.

PR #652, opened 2020-05-24
Working Curio Vendor Moogle

kaincenteno:
Added all shop items (except keyitems) for all curio moogles (UP to RHAPSODIES IN AZURE as the other quests have not been scripted, but should be easy to just populate with correct information once they are)
Made it a global to not have to have those scripts all over the place.
Thank you Hiro from the Canaria server for testing and making sure that the prices and order was correct.

**_I affirm:_**

### DIFF
--- a/scripts/globals/shop.lua
+++ b/scripts/globals/shop.lua
@@ -6,7 +6,23 @@
 require("scripts/globals/conquest")
 require("scripts/globals/settings")
 require("scripts/globals/crafting")
+require("scripts/globals/keyitems")
 -----------------------------------
+
+-----------------------------------
+-- IDs for Curio Vendor Moogle
+-----------------------------------
+
+curio =
+{
+    ["medicine"]        = 1,
+    ["ammunition"]      = 2,
+    ["ninjutsuTools"]   = 3,
+    ["foodStuffs"]      = 4,
+    ["scrolls"]         = 5,
+    ["keys"]            = 6,
+    -- keyitems not implemented yet
+}
 
 tpz = tpz or {}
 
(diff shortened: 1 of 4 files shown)


Kreidos:
Great work, kain!

cocosolos (approved):

scripts/globals/shop.lua
```diff
@@ -383,5 +403,157 @@ tpz.shop =
                 8828,    7600,      tpz.craftRank.AMATEUR,      -- Smithing Kit 50
                 9247, 1126125,      tpz.craftRank.AMATEUR       -- Niobium Ore
         }
+    },
+
+    curioVendorMoogleStock =
+    {
+        ["medicine"] =
```
TeoTwawki: quote-string index keys?

kaincenteno: something wrong with it @TeoTwawki ?

Kreidos: It can be done either way, though it isn't strictly required unless you're using words which can't be _proper identifiers_... I'm personally okay with it either way.

TeoTwawki: I'd kinda like to keep with the existing style rather than having one thats just different without needing to be

like how you did the guild ones

kaincenteno: Done @TeoTwawki  please give it your blessing.
Just a heads-up in the if statement, I made it nested so whomever takes ownership of keyitems can just add the extra option for it #notIt lol

TeoTwawki: Ima preist now, I bless PRs.

:latin_cross:

cocosolos (approved):

ibm2431 (approved):
Items, prices, and KI item requirements check out. Gonna assume the item IDs check out. If one doesn't, I'm sure someone will say something before `release`~

PR #653, opened 2020-05-25
Artisan Moogles + Mog Sack

Kreidos:
Fully implemented code for the 4 Artisan Moogles and accompanying Mog Sack purchase/expansion abilities. They also now provide Warp Scrolls to players once per day as in retail. As retail-like capability exists now for purchasing Mob Sacks, this also sets default Mog Sack size for new players to 0 (as retail). Existing DBs will not be migrated, so as to not cause surprise issues for server admins & players.

**_I affirm:_**

### DIFF
--- a/scripts/globals/artisan.lua
+++ b/scripts/globals/artisan.lua
@@ -22,11 +22,6 @@ local menuFlags = {
 	aexpand = 0x2,
 }
 
-local getEpochDays = function()
-    local time = os.date("*t")
-    return time.year*365 + time.yday
-end
-
 tpz.artisan.moogleOnTrigger = function(player, npc)
     local csid = event[player:getZoneID()]
     local menuMask = 0
@@ -62,13 +57,13 @@ tpz.artisan.moogleOnUpdate = function(player,csid,option)
         end
 
     elseif option == 3 then -- Client requests sack + scroll status
-        local scrollAvail = player:getCharVar("[artisan]nextScroll") <= getEpochDays() and 1 or 0
+        local scrollAvail = player:getCharVar("[artisan]nextScroll") < getMidnight() and 1 or 0
         local sackSize = player:getContainerSize(tpz.inv.MOGSACK)
         if sackSize > 0 then sackSize = sackSize + 1 end
         player:updateEvent(0, 0, 0, sackSize, 0, 0, 0, scrollAvail)
 
     elseif option == 4 then -- Main dialogue
-        local scrollAvail = player:getCharVar("[artisan]nextScroll") <= getEpochDays() and 1 or 0
+        local scrollAvail = player:getCharVar("[artisan]nextScroll") < getMidnight() and 1 or 0
         local sackSize = player:getContainerSize(tpz.inv.MOGSACK)
         if sackSize > 0 then sackSize = sackSize + 1 end
         player:updateEvent(0, 0, player:getGil(), sackSize, 0, 0, 0, scrollAvail)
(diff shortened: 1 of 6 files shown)


scripts/globals/artisan.lua
```diff
@@ -0,0 +1,91 @@
...
+local menuFlags = {
+	expand = 0x8,
+	abags = 0x4,
+	aexpand = 0x2,
+}
+
+local getEpochDays = function()
+    local time = os.date("*t")
+    return time.year*365 + time.yday
+end
```
cocosolos: Might wanna use `getMidnight()` instead, will keep this in line with everything else that resets at "JP" midnight.

Kreidos: I wasn't aware that function existed. I will get that changed.

Kreidos: Updated.

cocosolos (approved):

scripts/globals/artisan.lua
```diff
@@ -0,0 +1,86 @@
...
+local menuFlags = {
+	expand = 0x8,
+	abags = 0x4,
+	aexpand = 0x2,
+}
+
+tpz.artisan.moogleOnTrigger = function(player, npc)
+    local csid = event[player:getZoneID()]
+    local menuMask = 0
+    local mogVisited = player:getCharVar("[artisan]visited")
```
ibm2431: Let's make this mogVisited assignment if the sackSize is > 0 or the var is one. Then when the player purchases a sack (probably every player will), we can nuke the charVar, and not have a forever var lingering in char_vars table.

ibm2431 (changes requested):
Just a small tweak to avoid a forever var~!

scripts/globals/artisan.lua
TeoTwawki: note: in other similar cases where you can't avoid what would be a forever var, you can start newly created characters at 1 and then zero them after, instead of starting 0 and moving to one. (see existing examples in player.lua)

Here, you don't have to do player.lua because we can check what ibm said.

Kreidos: Great idea, ibm. I've changed it so it'll only be set 1 for as long as the player has no sack, and has talked to a moogle. It will be zeroed once they have a bag and check will be based on either or.

I'll push it once I've tested a bit later.

ibm2431 (approved):

Noikar:
This will still never be necessary until the change inventory code is removed from Bluffnix.
`
function onEventFinish(player,csid,option)

    local TheGobbieBag = gobQuest(player,player:getContainerSize(0));

    if (csid == 43 and option == 0) then
        if (player:getQuestStatus(JEUNO,TheGobbieBag[1]) == 0) then
            player:addQuest(JEUNO,TheGobbieBag[1]);
        end
    elseif (csid == 73) then
        if (gobbieBag == 5) then
            player:addTitle(tpz.title.GREEDALOX);
        elseif (gobbieBag == 10) then
            player:addTitle(tpz.title.GRAND_GREEDALOX);
        end

        player:changeContainerSize(0,5);
        player:changeContainerSize(5,5);
        player:changeContainerSize(6,5);
        player:addFame(JEUNO, 30);
        player:addExp(player:getContainerSize(0) * EXP_RATE);
        player:tradeComplete();
        player:completeQuest(JEUNO,TheGobbieBag[1]);
        player:messageSpecial(ID.text.INVENTORY_INCREASED);
    elseif (csid == 10056) then
        player:setMaskBit(player:getCharVar("WildcatJeuno"),"WildcatJeuno",12,true);
    end
end;
`

PR #658, opened 2020-05-27
Fix Time-of-Day events not firing reliably (Night mobs not spawn/despawn)

### LINKED ISSUE #282: Undead do not despawn if no player is in the zone to cause tick
<a href="https://github.com/Bizzle-Dapp"></img></a> **Issue by [Bizzle-Dapp](https://github.com/Bizzle-Dapp)**
_Monday Apr 22, 2019 at 22:52 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 5882_

----

**_I have:_**
Initially noticed a Wight roaming Rolanberry fields in the middle of the daytime. Have since spotted Zombies in Pashlow Marshlands and a previous issue: github/DarkstarProject/darkstar - Issue 2570 references a very similar issue. 

**_Client Version_** (type `/ver` in game) **:** 30190328_2

**_Source Branch_** (master/stable) **:** Initially from whasf installer - since slightly editted

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**  Likely all

Kreidos:
This fixes longstanding intermittent issues with Evening Mobs not spawning, not de-spawning, Day/Night Latents not always taking effect, guild shops not always restocking, and many more things which are dependent on reliable time-transition detections.

The long explanation of this problem is that the time_server runs based on a delay of 2400ms (one vanadiel minute), while the SyncTime() function uses the true system clock to generate the current Vanadiel Time. Since a delay-based sync cycle is variable given various system loads, conditions, and code execution, it will drift in and out of phase with the SyncTime() time generation, leading to vanadiel minutes periodically being skipped or repeated. If this skipped second was minute 0 of an hour, this would result in TOTD updates not firing.

TOTD updates are the only absolutely essential trigger based on this, and so should detect hour boundaries rather than "minute 0". I don't believe a rewrite of the clock mechanism is necessary or helpful, as the overall vanadiel timekeeping is still accurate; rather we just be cognizant of the granularity of its momentary accuracy.

**_I affirm:_**

Fixes #282

### DIFF
--- a/src/map/vana_time.cpp
+++ b/src/map/vana_time.cpp
@@ -196,8 +196,10 @@ TIMETYPE CVanaTime::SyncTime()
     m_vHour = (uint32)((m_vanaDate % VTIME_DAY)   / VTIME_HOUR);
     m_vMin  = (uint32)( m_vanaDate % VTIME_HOUR);
 
-    if (m_vMin == 0)
+    static int8 lastTickedHour = m_vHour;
+    if (m_vHour == (lastTickedHour+1)%24)
     {
+        lastTickedHour = m_vHour;
         switch (m_vHour)
         {
             case  0: m_TimeType = TIME_NIGHT;   return TIME_MIDNIGHT;


tankfest:
Such an annoying bug... such a seemingly simple fix... well done, sir.

Kreidos:
> Such an annoying bug... such a seemingly simple fix... well done, sir.
Ha ha My father told me a story about a mechanic charging a customer $100 to turn one loose bolt which fixed a car that wouldn't start. When asking why a 1 minute fix was so expensive he replied it's $2 for the turn, and $98 for knowing _which_ bolt to turn.

This is one of those almost embarrassingly small fixes. It was just realizing the cause. :)

src/map/vana_time.cpp
```diff
@@ -196,8 +196,10 @@ TIMETYPE CVanaTime::SyncTime()
     m_vHour = (uint32)((m_vanaDate % VTIME_DAY)   / VTIME_HOUR);
     m_vMin  = (uint32)( m_vanaDate % VTIME_HOUR);
 
-    if (m_vMin == 0)
+    static int8 lastTickedHour = m_vHour;
+    if (m_vHour == (lastTickedHour+1)%24)
```
ibm2431, line 200: Does this `if` still fire if we're assigning `lastTickedHour` to be `m_vHour` right before checking if the current hour is our (just-assigned) `lastTicked` hour plus one?

ibm2431 (commented):
The simplicity of the solution is/will be amazing; I'm just confused over timing of var assignments~

src/map/vana_time.cpp
Kreidos: Hey @ibm2431 ! It will, because I've declared `lastTickedHour` as static. Static variables are only initialized once the first time they're declared, and will have indefinite lifespan like globals do, but they still maintain local scope. I selected static rather than a class variable since this variable never needs to be used outside of this local function. :)

*Edit:* I felt I should clarify that this means line 199 only does something the first run-through.

cocosolos: I actually had this same question lol. So the declaration happens once, then the reassignment happens every time?

Kreidos: The declaration and assignment happens only once. Basically, after the first pass, line 199 is inert and does nothing. The only time it'll be assigned is after the following `if` statement is true, since I have an assignment statement in there.

cocosolos: Yeah I meant the assignment on line 202. Makes sense, thanks for clarifying! Good work.

cocosolos (approved):

ibm2431 (approved):

Issue #551, opened 2020-04-30
Enemies never drop more than one crystal.


tankfest:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

With current code, when a player has signet, there is a base 20% chance for an enemy that *can* drop a crystal to drop it.  Regardless of how many players are in the party and have attacked the enemy, only one crystal can drop, with the % rate being unaffected.  In retail, the # of players with signet that have attacked the mob should increase the amount of crystals that can/will drop.

TeoTwawki:
What I and @KnowOne134 found out is that retail will  load its "global drops" (like crystals) onlyin any unused drop slots that nothing already filled, up to the party size that killed the mob. This is quite different from what happens now at a static rate with a cooldown.

ibm2431:
> only in any unused drop slots that nothing already filled, up to the party size that killed the mob

For clarification, if a mob is killed by a party with two members, and the mob drops 1 rabbit hide, is it:

1) One crystal is added, bringing the total drops to 2 (the max party size);
or
2) Two crystals are added, bringing the total drops to 3 (hide + size, so long as total is kept to less than 8)

?

TeoTwawki:
its crystals added to existing pool as long as there are free spaces to do so. Party size doesn't limit the pool size, just determine show many xtal -can try- to fill in the empty slots of the pool. similar situations happen for seals geodes etc.

KnowOne134:
I think it has diminishing returns. A party of 6 can have 6 Crystal's drop. But is super rare. Cap is 6

TeoTwawki:
I believe they simply roll per party member. so to get 6 all 6 members rolls would would need to succeed in addition to have 6 unused loot pool slots after the mob dies.

PR #662, opened 2020-05-27
Fixes crystal drops for parties.

tankfest:
**_I affirm:_**

Adds functionality for multiple crystal drops from a mob while in a party.  Also re-orders drops to make seals occur before crystals as is retail.  The code was tested on my server in a party of 1 to 3 people and works well.  The code was styled a tiny bit differently for topaz, so this exact code wasn't built or tested.

Closes issue #551 

Co-contribution credit to Teotwawki and cocosolos for working on this with me.

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -919,19 +919,36 @@ void CMobEntity::DropItems(CCharEntity* PChar)
 
         uint8 effect = 0; // Begin Adding Crystals
         
-        if (conquest::GetRegionOwner(PChar->loc.zone->GetRegionID()) <= 2 && m_Element > 0) // Valid Signet Zone
+        if (m_Element > 0)
         {
-            effect = 1;
+            uint8 regionID = PChar->loc.zone->GetRegionID();
+            switch (regionID)
+            {
+                // Sanction Regions
+                case REGION_WEST_AHT_URHGAN:
+                case REGION_MAMOOL_JA_SAVAGE:
+                case REGION_HALVUNG:
+                case REGION_ARRAPAGO:
+                    effect = 2;
+                    break;
+                // Sigil Regions
+                case REGION_RONFAURE_FRONT:
+                case REGION_NORVALLEN_FRONT:
+                case REGION_GUSTABERG_FRONT:
+                case REGION_DERFLAND_FRONT:
+                case REGION_SARUTA_FRONT:
+                case REGION_ARAGONEAU_FRONT:
+                case REGION_FAUREGANDI_FRONT:
+                case REGION_VALDEAUNIA_FRONT:
+                    effect = 3;
+                    break;
+                // Signet Regions
+                default:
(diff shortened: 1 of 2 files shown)


src/map/entities/mobentity.cpp
```diff
@@ -924,6 +913,58 @@ void CMobEntity::DropItems(CCharEntity* PChar)
...
+                case 3:
+                    if (PMember->StatusEffectContainer->HasStatusEffect(EFFECT_SIGIL) && PMember->getZone() == getZone() && distance(PMember->loc.p, loc.p) < 100)
+                    {
+                        crystalRolls++;
+                    }
+                    break;
+                default:
+                    break;
+            }
+        });
```
ShelbyZ, line 977: Let's simplify this block as each case is checking:

- if the member is in zone
- if the member is within distance

```cpp

PChar->ForParty([this, &crystalRolls, &effect](CBattleEntity* PMember)
{
    If (PMember->getZone() == getZone() && distance(PMember->loc.p, loc.p) < 100)
    {
        switch (effect)
        {
            case 1:
                if (PMember->StatusEffectContainer->HasStatusEffect(EFFECT_SIGNET)
                {
                    crystalRolls++;
                }
                break;
            // other cases similar..
        }
    }
});
```

ShelbyZ (changes requested):

src/map/entities/mobentity.cpp
cocosolos: Each case except the default case, and then for each member we don't even bother checking zone and distance unless they also have the status.

src/map/entities/mobentity.cpp
```diff
@@ -924,6 +913,58 @@ void CMobEntity::DropItems(CCharEntity* PChar)
...
+            effect = 1;
+        }
+        else if (PChar->loc.zone->GetRegionID() >= 28 && PChar->loc.zone->GetRegionID() <= 32 && m_Element > 0) // Valid Sanction Zone
+        {
+            effect = 2;
+        }
+        else if (PChar->loc.zone->GetRegionID() >= 33 && PChar->loc.zone->GetRegionID() <= 40 && m_Element > 0) // Valid Sigil Zone
+        {
+            effect = 3;
+        }
```
cocosolos: I would maybe change this to:
```cpp
        if (m_Element > 0)
        {
            uint8 regionID = PChar->loc.zone->GetRegionID();
            switch (regionID)
            {
                // Sanction Regions
                case 28:
                case 29:
                case 30:
                case 31:
                    effect = 2;
                    break;
                // Sigil Regions
                case 33:
                case 34:
                case 35:
                case 36:
                case 37:
                case 38:
                case 39:
                case 40:
                    effect = 3;
                    break;
                // Signet Regions
                default:
                    effect = (conquest::GetRegionOwner(PChar->loc.zone->GetRegionID()) <= 2) ? 1 : 0;
                    break;
            }
        }
```
This would avoid a lot of unnecessary comparisons being done in the if/else stack.

cocosolos (approved):

PR #666, opened 2020-05-30
Rov valor buffs

kaincenteno:
**_I affirm:_**

Changes made are for oneventstart to show correct price and
oneventfinish to charge the new lower price.

Dunno why its also commiting old commits? (i just forked this)

### DIFF
--- a/scripts/globals/regimes.lua
+++ b/scripts/globals/regimes.lua
@@ -14,6 +14,7 @@ require("scripts/globals/status")
 require("scripts/globals/utils")
 require("scripts/globals/zone")
 require("scripts/globals/msg")
+require("scripts/globals/keyitems")
 ------------------------------------
 
 tpz = tpz or {}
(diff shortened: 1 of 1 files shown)


kaincenteno:
any soul willing to review this? all i did was just add an if statement to check for a param that decides if u have key item or not, then copied one of the tables but with the newer price deducted if u have keyitem ^^;;

cocosolos:
I've looked through this and I believe it works. I'm not sure if we should duplicate the table just for a different price, or if it would be better for the table to return functions, something like this:
```lua
[117] = function (rhapsodyInWhite) if rhapsodyInWhite then return {act = "DRIED_MEAT",      cost = 25, food = true} else return {act = "DRIED_MEAT",      cost = 50, food = true} end end,
```
or
```lua
[117] = function (rhapsodyInWhite) return {act = "DRIED_MEAT",      cost = function() if rhapsodyInWhite then return 25 else return 50 end end, food = true} end,
```
then check the key item like this
```lua
    for k,v in pairs(regimeInfo[regimeType].finishOptions) do
        out[k] = v(hasKI)
```
So I'll let someone chime in on that otherwise I have no problems merging.

TeoTwawki:
another possibility is to leave it one table and leave cost alone, but then add another entry for the RoV modified cost like
`[ 21] = {act = "REPATRIATION",    cost = 50, discounted = 10},`
and then change the for loop part to check if you have KI before deciding which one to use in the loop.

I'd go with the functions coco suggested if thats not easy to do.

cocosolos:
> `[ 21] = {act = "REPATRIATION", cost = 50, discounted = 10},`

I like this, much cleaner. Would have to update 2 places in `bookOnEventFinish` to check for KI and use either `opt.cost` or `opt.discounted`.

kaincenteno:
@cocosolos done :cat: 
That SQL file came back again >.< can you remove it please

scripts/globals/regimes.lua
```diff
@@ -1050,21 +1057,27 @@ tpz.regime.bookOnEventFinish = function(player, option, regimeType)
...
-        return
+    if hasKI then
+        if opt.discounted and opt.discounted > tabs then
+            player:showText(player, msgOffset + 1032) -- You do not have enough tabs.
+            return
+        elseif opt.cost and opt.cost > tabs then
+            player:showText(player, msgOffset + 1032) -- You do not have enough tabs.
+            return
+        end
     end
```
cocosolos: Should this be checking regular cost if no KI? Like this:
```lua
    if hasKI then
        if opt.discounted and opt.discounted > tabs then
            player:showText(player, msgOffset + 1032) -- You do not have enough tabs.
            return
        end
    else
        if opt.cost and opt.cost > tabs then
            player:showText(player, msgOffset + 1032) -- You do not have enough tabs.
            return
        end
    end
```

kaincenteno: fixed :)

cocosolos (approved):
Looks good, thank you!

ibm2431 (commented):
```lua
local cost = opt.cost

if cost and player:hasKeyItem(tpz.ki.RHAPSODY_IN_WHITE)
    cost = opt.discounted
end

if cost > tabs then
    -- Not enough
end

if opt.act then
    player:delCurrency("valor_point", cost)
```

kaincenteno:
it's the curse of the 666 

changes done.

ibm2431 (approved):

PR #672, opened 2020-05-31
Winsock updates

ShelbyZ:
- Removing _WINSOCK_DEPRECATED_NO_WARNINGS (connect/game)
- Remove unused functions host2ip, socket_getips
- Fixing log statements using MSGSERVTYPE and not casting to uint
- cmake updated

Tested on win/nix

**_I affirm:_**

### DIFF
--- a/src/common/socket.cpp
+++ b/src/common/socket.cpp
@@ -276,12 +276,12 @@ bool _vsocket_final(void){
 }
 
 // hostname/ip conversion functions
-const char* ip2str(uint32 ip, char ip_str[16])
+std::string ip2str(uint32 ip)
 {
     uint32 reversed_ip = htonl(ip);
-    char* address = new char[INET_ADDRSTRLEN];
+    char address[INET_ADDRSTRLEN];
     inet_ntop(AF_INET, &reversed_ip, address, INET_ADDRSTRLEN);
-    return (ip_str == NULL) ? address : strncpy(ip_str, address, 16);
+    return std::string(address);
 }
 
 uint32 str2ip(const char* ip_str)
--- a/src/common/socket.cpp
+++ b/src/common/socket.cpp
@@ -156,10 +156,6 @@ time_t last_tick;
 time_t tick_time;
 time_t stall_time = 60;
 
-uint32 g_addr_[16];   // ip addresses of local host (host byte order)
-
-int32 naddr_;   // # of ip addresses
-
 int32 makeConnection(uint32 ip, uint16 port, int32 type)
 {
 	struct sockaddr_in remote_address;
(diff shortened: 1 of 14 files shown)


zach2good (commented):
LGTM 👍

cocosolos:
I can't say I know exactly what this is doing but it builds/runs fine and nothing looks wrong with it, so I'm approving so it doesn't sit forever.

cocosolos (approved):

src/common/socket.cpp
```diff
@@ -373,22 +276,20 @@ bool _vsocket_final(void){
...
-
 const char* ip2str(uint32 ip, char ip_str[16])
 {
-	struct in_addr addr;
-	addr.s_addr = htonl(ip);
-	return (ip_str == NULL) ? inet_ntoa(addr) : strncpy(ip_str, inet_ntoa(addr), 16);
+    uint32 reversed_ip = htonl(ip);
+    char* address = new char[INET_ADDRSTRLEN];
+    inet_ntop(AF_INET, &reversed_ip, address, INET_ADDRSTRLEN);
+    return (ip_str == NULL) ? address : strncpy(ip_str, address, 16);
```
cocosolos: If `ip_str` is not null, does the memory for `address` need to be freed after `strncpy`?

ShelbyZ: I had to dig to find where it is used with an ip_str provided and found exactly 1 instance:

https://github.com/project-topaz/topaz/blob/45bfc2d4f658df99e9e30b3a88e78cfb2a440636/src/map/map.cpp#L439-L441

All other instances pass nullptr.  We do not use ip_str after calling the method.  It may be reasonable to modify the method to drop the second parameter and avoid strncpy entirely.

cocosolos: Would that still be a leak because none of the functions this is returning to frees the memory? Should we instead always pass the second param and get rid of the need for address being used here?
```cpp
const char* ip2str(uint32 ip, char ip_str[16])
{
    uint32 reversed_ip = htonl(ip);
    inet_ntop(AF_INET, &reversed_ip, ip_str, INET_ADDRSTRLEN);
    return ip_str;
}
```

ShelbyZ: All current calls are returning the results via logging inline.  If we had a reason that some function would be wanting to store that value long term it may make sense.  I'm not sure why I went with new char[...].

```cpp
std::string ip2str(uint32 ip)
{
    uint32 reversed_ip = htonl(ip);
    char address[INET_ADDRSTRLEN];
    inet_ntop(AF_INET, &reversed_ip, address, INET_ADDRSTRLEN);
    return std::string(address);
}
```

src/common/socket.cpp
```diff
@@ -373,22 +276,20 @@ bool _vsocket_final(void){
...
 
 // hostname/ip conversion functions
-uint32 host2ip(const char* hostname)
-{
-	struct hostent* h = gethostbyname(hostname);
-	return (h != NULL) ? ntohl(*(uint32*)h->h_addr) : 0;
-}
-
-const char* ip2str(uint32 ip, char ip_str[16])
+std::string ip2str(uint32 ip)
```
TeoTwawki, line 279: Death to `const char*` !  Long live `std::string` !

and you even remembered the nulls at the ends in the ShowWhatevers :heart:

PR #677, opened 2020-06-02
Update abilities.sql

illzbee:
added 295,'curing_waltz_v'

**_I affirm:_**


cocosolos:
Looks good, is there a reason you moved this out of order? It looked like you had it in the right place in the first commit.

TeoTwawki:
looks like travis was having a partial outage at the time of the 1st commit, guessing you mistakenly thought you needed to put new stuff at the bottom because of that error. the rows should stay sorted by primary key and the error was just the service being partly out.
> apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install software-properties-common g++-7 luajit-5.1-dev libzmq3-dev mysql-client-core-5.6 mysql-server-5.6 mysql-client-5.6 luarocks python3.7" failed and exited with 100 during .

TeoTwawki:
p.s. commit messages should be more descriptive than "update thing"

illzbee:
When I went on my test files it moved it on the bottom. Which made me think
that was the cause of the error. I won't make the mistake again. Thanks for
letting me know.

Wiggo32 (approved):

Wiggo32 (commented):
We wanted to put it back in order though right?

PR #678, opened 2020-06-03
Change updateClaim to clear claim when no args

cocosolos:
Also just return if used on something other than a mob, rather than crash.
Fixes issues with COR elemental shots, angon, and chi blast when used on
charmed players.

**_I affirm:_**

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -10360,10 +10360,17 @@ inline int32 CLuaBaseEntity::resetEnmity(lua_State *L)
 inline int32 CLuaBaseEntity::updateClaim(lua_State *L)
 {
     TPZ_DEBUG_BREAK_IF(m_PBaseEntity == NULL);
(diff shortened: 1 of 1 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -10360,10 +10360,17 @@ inline int32 CLuaBaseEntity::resetEnmity(lua_State *L)
 inline int32 CLuaBaseEntity::updateClaim(lua_State *L)
 {
     TPZ_DEBUG_BREAK_IF(m_PBaseEntity == NULL);
-    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
+    if(m_PBaseEntity->objtype != TYPE_MOB)
```
zach2good: Space after if 🙏

zach2good (approved):
Makes sense to me! Pretty clear whats going on here 👍

Wiggo32 (approved):

ibm2431 (approved):

PR #679, opened 2020-06-03
Fix Shen crash on pop

cocosolos:
Fixes `updateEnmity` call on wrong target. Mob logic wasn't touched.

**_I affirm:_**


zach2good (approved):
Nice catch ⭐

59blargedy (approved):

Wiggo32 (approved):

PR #683, opened 2020-06-04
Implement Rampart as -25% SDT rather than 100% magic shield

wrenffxi:
**_I affirm:_**

Rampart's effect was changed in Feb. version update to be a -25% SDT that's multiplicative with DT.
https://forum.square-enix.com/ffxi/threads/56444-February-12-2020-%28JST%29-Version-Update
https://www.ffxiah.com/forum/topic/46016/first-and-final-line-of-defense-v20/72/#3487802


Wiggo32 (approved):

ibm2431 (approved):
Thanks, Wren!

PR #685, opened 2020-06-04
Abyssea Lights system

Omnione:
This PR will introduce the visitant light system for Abyssea, it will currently do nothing but control the lights that a player receives after killing a mob in an  Abyssea zone, all mobs have been mapped for the lights they should give, and all light values are as retail unless no information could be found, in which case the value will be an estimate based on other mobs of the same family in the same tier.

The lights that drop will be dependent on the way the mob has been killed as per retail.

All information has been collated and checked on multiple sites and also from retail testing. 

I have included GM commands to add lights and reset lights accordingly (!addlights and !resetlights).

special thanks to Lynnea of then Dawnbreak server for the help on this.

**_I affirm:_**


zach2good:
At it again with the giant PRs ;)

Omnione:
Maybe lol

scripts/globals/abyssea.lua
```diff
@@ -399,3 +415,170 @@ function abysseaOnEventFinish(player,csid,option)
...
+
+    local tiermsg     = 0
+    local lightamount = amount or 0
+    local pearl       = player:getCharVar("pearlLight")
+    local azure       = player:getCharVar("azureLight")
+    local ruby        = player:getCharVar("rubyLight")
+    local amber       = player:getCharVar("amberLight")
+    local gold        = player:getCharVar("goldLight")
+    local silver      = player:getCharVar("silverLight")
+    local ebon        = player:getCharVar("ebonLight")
```
cocosolos: It looks like these are only being used in their respective blocks below, can we move these definitions down there so we only load the ones we need?

scripts/globals/abyssea.lua
```diff
@@ -399,3 +415,170 @@ function abysseaOnEventFinish(player,csid,option)
...
+}
+
+function isInAbysseaZone(player)
+    for index, zone in ipairs(abyssea_zones) do
+        if zone == player:getZoneID() then
+            return true
+        end
+    end
+    return false
+end
```
cocosolos: Can this just `return player:getCurrentRegion() == tpz.region.ABYSSEA`?

scripts/mixins/drop_lights.lua
```diff
@@ -0,0 +1,56 @@
...
+                    break
+                end
+            end
+            DropLights(user, target:getName(), wsType, target)
+        end 
+    end)
+
+    mob:addListener("ABILITY_TAKE", "ABYSSEA_ABILITY_DEATH_CHECK", function(mob, user, ability, action)
+        if mob:getHP() <= 0 then
+            DropLights(user, mob:getName(), "physical")
```
cocosolos: Should this be passing the mob?

scripts/mixins/drop_lights.lua
```diff
@@ -0,0 +1,56 @@
...
+    end)
+
+    mob:addListener("ATTACKED", "ABYSSEA_PHYSICAL_DEATH_CHECK", function(target, attacker)
+        if target:getHP() <= 0 then
+            DropLights(attacker, target:getName(), "physical", target)
+        end
+    end)
+
+    mob:addListener("DEATH", "ABYSSEA_DEATH_NO_ACTION", function(mob, player, isKiller)
+        DropLights(attacker, mob:getName(), "physical", mob)
```
cocosolos: Should `attacker` be `player`? Also is this one called for each member? Then DropLights adds light for each member as well, so are they getting hit twice? Should this check for `isKiller`?

scripts/zones/Abyssea-Altepa/Zone.lua
```diff
@@ -17,6 +18,8 @@ function onZoneIn(player,prevZone)
         player:setPos(435 ,0 ,320 ,136)
     end
 
+    ResetPlayerLights(player)
```
cocosolos: Will this reset lights if a player logs out then logs back in? Or accepts a tractor? I think light is supposed to persist that, at least logging out, IDK about tractor.

cocosolos (commented):
Looks pretty good, still need to actually test it but it's a good start. Another comment/suggestion I'd like to see is the same naming conventions used by Dynamis/Limbus and others. I see a lot of `AddPlayerLights` in files and it's unclear where those are defined. I think `tpz.abyssea.AddPlayerLights` would be more clear.

scripts/commands/addlights.lua
```diff
@@ -0,0 +1,60 @@
...
+        targ = GetPlayerByName(target)
+        if (targ == nil) then
+            error(player, string.format("Player named '%s' not found!", target))
+            return
+        end
+    end
+
+    local selectedLight = tostring(light)
+    local setLight = lightType[selectedLight]
+    if (light == nil or setLight == nil) then
```
whasf: I think you will get a lua "nil index" error if selectedLight is nil (you may want to bomb out separately before line 45)

scripts/globals/abyssea.lua
TeoTwawki: Why chavars when any zoning loses it on retail? setLocalVar() would fit with retail behavior.

cocosolos: Retail behavior saves lights on logout.

scripts/globals/moblights.lua
```diff
@@ -0,0 +1,601 @@
...
+tpz = tpz or {}
+tpz.abyssea_mob = tpz.abyssea_mob or {}
+
+-- if the amount is 100 or 50 it is a special number and is randomised as per retail
+tpz.abyssea_mob.info =
+{
+    ------------------------------------------------------------------------------------------------------------
+    -- Abyssea - Konschtat
+    ------------------------------------------------------------------------------------------------------------
+    ["Ab_xzomit"]              = {azure = 8,   pearl = 5,   ruby = 16,  amber = 8  }, -- Xzomits
```
TeoTwawki, line 14: I would handle this information the way we handle mob 2hrs - input passed to a mixin, rather than trying to name every mob in every abyssea zone in a global table like this.

You lose your single central location, but gain a cleaner more efficient code path.

scripts/globals/abyssea.lua
TeoTwawki: ..has it always been that way? Old man brain _swears_ I always lost them if I dc’d

scripts/globals/abyssea.lua
TeoTwawki: Seconded.

scripts/globals/abyssea.lua
cocosolos: Not sure, wikia has this line from 2011,
> Built-up lights are only lost when a character leaves Abyssea. Changing to a different enemy target or logging out/disconnecting does not remove lights, nor do they decay over time.

TeoTwawki: Sounds like we could potentially store them in a hidden status effect and then flag the effect to be silently lost on zonage - like the Prowess buffs.

Speaking of effects, past me likely did the cruor buffs incorrectly even though they work. >.>;;  (used dummy IDs) I should check that soonish

scripts/mixins/drop_lights.lua
TeoTwawki: Death listener is running on just the killer iirc

cocosolos: Yeah I think you're right. Wrong param is being passed still though.

cocosolos: Actually the DEATH listener only has `mob` and `killer` being passed to it, there is no `isKiller`, and `killer`/`player` can be nil.

scripts/mixins/drop_lights.lua
Omnione: man, how do i  do that, i mean i must have to do this stuff by accident, cos i know it was there before, must be lack of sleep.

scripts/globals/abyssea.lua
Omnione: good call sir's

scripts/globals/abyssea.lua
Omnione: hmm, i may have to look into what Teo suggested and add them to visitant status effect maybe?

Omnione: oh, thinking about it, i may have to store this in a sql, ugh.

scripts/commands/addlights.lua
Omnione: yep, good call, cheers whasf

scripts/zones/Abyssea-Altepa/Zone.lua
Omnione: Accepting Tractor will remove this effect. (Note that instead of taking you to where Tractor was cast upon you, you will instead end up back at the entry encampment of whatever Abyssea area you were in, with your Visitant status, Abyssea Lights, and mob (EXP or Gil) chain all removed as if you had just entered the area again.)

scripts/mixins/drop_lights.lua
cocosolos: `DropLights(user, mob:getName(), "physical", mob)` it's just missing the last arg.

scripts/zones/Abyssea-Altepa/Zone.lua
cocosolos: It will reset the lights when they logout then log back in though, which it shouldn't do.

scripts/globals/abyssea.lua
cocosolos: Status effects and char vars are both stored in the db, char vars read the db every time you use them, so if you keep using them then just make sure you're doing it efficiently as you can.

cocosolos: You would likely need a different status effect for each light type. You could theoretically fit 3 light types into 1 char var, if you wanna go into that.

Omnione: i'll think about it tommorow, after i get some sleep, cheers

scripts/globals/abyssea.lua
Omnione: done

scripts/mixins/drop_lights.lua
Omnione: fixed

scripts/commands/addlights.lua
Omnione: changed

scripts/mixins/drop_lights.lua
Omnione: fixed now

scripts/zones/Abyssea-Altepa/Zone.lua
Omnione: ok, ill figure it out tommorow, ty cocosolos

scripts/mixins/drop_lights.lua
cocosolos: Sorry just wanna reopen this because `isKiller` is still there. I just want to make sure someone doesn't see it and think it can be used, since only the mob and the killing player, if any, are passed to this function.

Omnione: Whoops, i've taken it out now.

scripts/globals/abyssea.lua
```diff
@@ -483,102 +474,83 @@ function AddPlayerLights(player, light, amount)
...
             player:setCharVar("amberLight",255)
         else
             player:setCharVar("amberLight",amber + lightamount)
         end
-
         player:messageSpecial(ID.text.BODY_EMITS_AMBER_LIGHT,tiermsg)
+    -------------------------------------------------------------------------------------
+    -- Golden Light
+    -------------------------------------------------------------------------------------
     elseif light == 5 then
```
TeoTwawki:
```lua

    -- Golden Light
    elseif light == 5 then
```
instead of brick houses for comments

TeoTwawki (changes requested):

scripts/globals/abyssea.lua
TeoTwawki: or enums /shrug

scripts/globals/abyssea.lua
TeoTwawki: even if you used every parameter addStatusEffect has I don't think you could cram it all into visitant. so 7 effects or 7 charVars I guess, that you then have to remember to kill anytime player leaves zone in anyway, which makes me think effect route is less effort.

scripts/globals/abyssea.lua
ibm2431: I haven't looked at any code, just checking notifications:

Enums~! No magic numbers...!

`tpz.abyssea.light.GOLDEN`

scripts/zones/Abyssea-Altepa/Zone.lua
Omnione: This should work like retail now, needs to be tested more, but seems right my end, and works well.

scripts/globals/abyssea.lua
Omnione: after disscussing this with others, it seems, the charVar method will be best suited for this task, so this should not be an issue now that the lights should be removed like retail, they stay on death, logout, only reset on zone or timeout or any other means of losing the visitant status.

scripts/globals/abyssea.lua
Omnione: All magic numbers should have been replaced with enums now

59blargedy:
@project-topaz/developer @project-topaz/reviewer  this appears to be in a good state and potentially ready for merging. as something on the older side, would love a review.

scripts/commands/addlights.lua
```diff
@@ -0,0 +1,66 @@
...
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!addlights <light type> <amount> {player}")
+end
+
+function onTrigger(player,light,amount,target)
+    -- validate target
+    local targ
+    if (target == nil) then
```
TeoTwawki, line 34:
```lua
    if target then
        targ = GetPlayerByName(target)
        if not targ then
            error(player, string.format("Player named '%s' not found!", target))
            return
        end
    else
        targ = player
    end
```
would be preferred formatting.

other similar examples using `== nil` and extra parentheses exist in this pr as well. Does not affect functionality, but I thought you'd like to know to save yourself extra typing in the future.

scripts/globals/abyssea.lua
TeoTwawki: can mark this convo resolved since it has enums now

scripts/mixins/zones/Abyssea-Altepa.lua
```diff
@@ -0,0 +1,7 @@
+---------------------------------------------------------
+
+---------------------------------------------------------
+mixins = 
+{
+    require("scripts/mixins/drop_lights"),
```
TeoTwawki, line 6: https://discordapp.com/channels/639659267003252746/639670779386134548/745655082934337536

59blargedy: do you want to handle these last two comments as an issue?

TeoTwawki: nrly..

PR #688, opened 2020-06-05
Compiler updates - cmake/makefile.am

ShelbyZ:
- Add -Werror for other builds
- .gitignore **/config.* - weird created /common/config.in~
- fix some unused variables
- fix parenthesis in if-statements
- add default where switch case not handled
- fix uint -> int casting

**_I affirm:_**


zach2good (approved):
Spoke about this a little bit in Discord, looks good to me!

PR #494, opened 2020-04-14
Geo Initial preview commit

Omnione:
**_I affirm:_**
- [X] that I agree to Project Topaz's [Limited Contributor License Agreement](http://project-topaz.com/blob/release/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [X] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

This PR adds the basic functions for geo, including the auras or bubbles as some like to call them.
only spells available are as follows:
• indi-regen
• indi-poison
• indi-refresh
• geo-regen
• geo-posion
• geo-refresh

Abilities:
• Full Circle returns a fraction of mp based on the intial cost of the geo spell and luopans remaing HP

This is a WIP so feel free to test and make changes accordingly.

Credit to zach2good for making this possible and to actually move ahead.

NOTE: the animations when casting will look odd untill the mID for the bells are added if they are not added already.

PR #690, opened 2020-06-06
GEO bubble work as of April

ibm2431:
This is #494 , but rebased on current release (ie: with SQL already merge), and limited to only as of the end of April; without the additional spell bubbles, -ara spells, job abilities, etc. (In other words at [the point when I was mostly okay with merging to start a new branch](https://github.com/project-topaz/topaz/pull/494#issuecomment-621176083))

This will let us rebase/retarget #494 and better see what was added after April, and then begin splitting up and reviewing the out-of-scope GEO stuff that goes beyond the original bubble implementation.


ibm2431:
I had sole hold on the _original_ Pull Request. We never determined what happens when PRs are split -- or when more stuff is added after a hold -- but _this_ PR is work that was done prior to my hold, so I'll go ahead and hit the merge button...!

PR #691, opened 2020-06-06
Trust HP/MP/Stat/Skill Multiplier Settings

zach2good:
**_I affirm:_**

Not yet tested! But these seem like they'll be useful for any server that wants to downplay the power of trusts, so they aren't so much of a crutch.


ibm2431 (approved):
Looks fine, two questions:

1) Are there ever any retail campaigns which boost fellows? Setting for trust does need to be separate from pets to be able to match retail's campaigns. I don't know if fellows are included in those campaigns.

2) Will gambits adjust for impossible weapon skills if the skill level is no longer sufficient due to setting?

TeoTwawki:
> Will gambits adjust for impossible weapon skills if the skill level is no longer sufficient due to setting?

I'm not convinced there is such a thing as impossible for their weaponskills. multiple trusts don't obey the player rules and will use ws they would not have had access to on retail when summoned low lv. Surprised this even got a setting separate from base stats.

PR #699, opened 2020-06-08
Add geomancer -ra spells and scrolls

ibm2431:
These are the scripts related to geomancer AoE elemental nukes as of #494.

I have not reviewed them yet.

As @Omnione nor @Lynnea1320 will be the author of this Pull Request - and therefore won't be able to push additional commits - review items will have to be addressed in a future Pull Request.

Please review as if this were a normal Pull Request. We'll compile a list of work items that needs to be done for this branch prior to a merge into canary.


scripts/globals/spells/fira.lua
```diff
@@ -0,0 +1,29 @@
...
+function onMagicCastingCheck(caster, target, spell)
+    return 0
+end
+
+function onSpellCast(caster, target, spell)
+    local spellParams = {}
+    spellParams.hasMultipleTargetReduction = true
+    spellParams.resistBonus = 1.0
+    spellParams.V = 216
+    spellParams.V0 = 140
```
ibm2431: bgwiki spell page lists V0 at 140, but the [magic damage page](https://www.bg-wiki.com/bg/Magic_Damage) lists it at 240.

V0 of 240 probably makes more sense when compared with V and V50~

ibm2431 (commented):
I'll push a commit to adjust Fira V0.

@Omnione Two questions:

1) What are the sub-zero V and M values based on? They don't jump out at me as wrong, just wondering if there's a source we can point to / compare against.

2) When these spells are cast, they hit AoE around the Geomancer, and not the target, ya?

PR #700, opened 2020-06-08
Geomancer abilities, merits, and adjustments

ibm2431:
This is the work relating to Geomancer abilities, merits, and other adjustments made since the initial April submission of #494.

I have not reviewed it yet.

@Omnione won't be the author of this Pull Request, but the compare is from his fork, so he might still be able to push commits to the branch. If not, review items will have to be addressed in a future Pull Request.

Please review as if this were a normal Pull Request. We'll compile a list of work items that needs to be done for this branch prior to a merge into canary.


ibm2431:
I will need push access, but no longer have it since I closed #494, and I was the one who opened _this_ PR. GitHub won't let me reopen #494  since a PR from omnione/geo is already open (this PR). Taking the entire branch so we can have control over it.

PR #705, opened 2020-06-10
Fix unable to continue RoV questline

kaincenteno:
Change RoV to take precedence over CoP missions
Removed not needed ";", "(", and ")"
Added Comments for easy understanding
Moved also OnEventFinish to match the same order as onTrigger

Thank you Cyrianna(Canaria Server) @ibm2431  and @ffxijuggalo 

**_I affirm:_**


ibm2431 (approved):
Looks legit, thanks kain~!

PR #706, opened 2020-06-11
Fix Flashbulb and mob Sleepga from crashing server when resisted

ibm2431:
**_I affirm:_**
[ x ] that I agree to Project Topaz's [Limited Contributor License Agreement](http://project-topaz.com/blob/release/CONTRIBUTOR_AGREEMENT.md), as written on this date
[🤞] that I've _tested my code_ since the last commit in the PR, and will test after any later commits


cocosolos (approved):

ibm2431:
Changed message, per Wiggo retail capture:

> Shiva uses Sleepga, but misses Fleshflayer Killakriq.

edit: And checked in PVLV too:
github

Issue #552, opened 2020-04-30
Two different settings for server JP midnight time.


tankfest:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

**scripts\global\settings.lua:**
TIMEZONE_OFFSET = 9.0 -- Offset from UTC used to determine when JP Midnight is for the server.  Default is JST (+9.0).

**conf\map.conf:**
#Determine when JP midnight is.
vanadiel_time_offset: 8

Redundant settings, only one can/should be used.

Also if possible, maybe an option should be to pick it directly from the OS time.

cocosolos:
I think `vanadiel_time_offset` is used to control the in-game time/year, while `TIMEZONE_OFFSET` is used for gating requirements.

Edit: `vanadiel_time_offset` doesn't actually look like it does anything right now, but that seems to have been the intention at one time.

tankfest:
The comments both state they are for determining when "JP midnight" is.  If vanadiel_time_offset isn't used for anything, perhaps it should be removed.

cocosolos:
I actually think the ~~setting should maybe be renamed~~ comment should be clarified and the feature implemented. It sounds like it would let you reboot your server and have it be days or years ahead in Vana'diel, and that could be useful.

PR #708, opened 2020-06-11
Custom epoch for Vana'diel time

cocosolos:
Rename setting to epoch from offset to hopefully clarify usage.
Adjust luautils function used with GM command.

Fixes issue #552

GM command `!timeoffset <offset>` will let you move around in time. Must zone for client to reflect changes. `!timeoffset 144` will take you an hour ahead, `timeoffset -144` will take you an hour back. Setting the conf setting to 1 will set the epoch to the farthest time back, making the client as far into the future as it can currently go. Setting the conf setting to current timestamp will set the time to 886/1/1 on Firesday. Anything greater than the current timestamp will set the date to 886/1/1 but will result in different day elements until current time is reached, on Firesday.

**_I affirm:_**

### DIFF
--- a/conf/default/map.conf
+++ b/conf/default/map.conf
@@ -76,8 +76,11 @@ exp_loss_rate: 1.0
 exp_party_gap_penalties: 1
 fov_allow_alliance: 1
 
-#Determine when "JP midnight" is.
-vanadiel_time_offset: 0
+#Determines Vana'diel time epoch (886/1/1 Firesday)
+# current timestamp - vanadiel_time_epoch = vana'diel time
+# 0 defaults to SE epoch 1009810800 (JP midnight 1/1/2002)
+#safe range is 1 - current timestamp
+vanadiel_time_epoch: 0
 
 #For old fame calculation use .25
 fame_multiplier: 1.00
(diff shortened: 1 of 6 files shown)


Kreidos:
Good stuff, Cocosolos!

I do have questions about how this will impact existing servers who choose to use the epoch; there are a fair number of scripts which save dates for timeouts and such in charvars. Many use functions directly linked to the current vanadiel year/month/day. (ie "No String Attached" or "Chocobo On The Loose") Any such quests could be softlocked for such a period of time that the offset catches up with it's pre-change value.

Edit: Also, I don't see the GM !timeoffset script in this branch. :(

cocosolos:
Anything using lockouts based on Vana'diel time will be effected by changing the epoch setting in map.conf or usage of the `!timeoffset` command since functions such as `VanadielDayOfTheYear` return a timestamp of Vana'diel time. Lockouts based on real time or JP midnight are based on the time offset supplied in settings.lua and use real time timestamps. The `!timeoffset` command is an existing command that currently does nothing since this functionality was removed at some point. Real usage for the map.conf setting would be setting a "birth" date for your server so the in game clock makes sense compared to the dates mentioned for major events. The `!timeoffset` command is more useful for testing development based on Vana'diel time.

ibm2431:
Talk to us about the removal/replacement of the time offset setting~!

edit: Nevermind, was discussed on the issue; the original setting didn't do what the comment implied it did.

ibm2431 (approved):

PR #718, opened 2020-06-13
Table spell IDs + move other magic enums out of magic global

ibm2431:
No more putting it off.

Convention is as follows:
- Spaces replaced with underscore
- Hyphens replaced with underscore. Lua does allow hyphens in table keys, but they need to be defined `['like-this'] =` and can't be defined `like-this =`. This also matches our [current hyphen handling in Core](https://github.com/project-topaz/topaz/blob/release/src/map/spell.h#L502-L508) 
- `'` and `:` were dropped
- Abbreviations were spelled out. We are not currently [consistent](https://github.com/project-topaz/topaz/blob/release/src/map/spell.h#L867) on [this](https://github.com/project-topaz/topaz/blob/release/src/map/spell.h#L888). Consistency on abbreviation with core will have to be addressed later. (Core typically uses the raw spell ID from the spell object as assigned by the DB _anyway_)
- Gaps in IDs are represented by an empty line for visual distinction that there is supposed to be a gap
- I'll want consistency with file names as well in the future, but that's outside the scope of finally getting an enum table in Lua for `trust`, `geo`, and wherever else we've been putting it off

@zach2good 

**_I affirm:_**


zach2good (approved):
Awww yeaaaahhhhhh

cocosolos (approved):

TeoTwawki:
if it must be done, at least its got its own home now instead of making magic or status globals astronomical proportions (I has a sadness anytime we have to enum the same IDs in all 3 of core lua and sql)

PR #721, opened 2020-06-13
Add conditions to delivery box block

cocosolos:
Check if player is in residential area.

**_I affirm:_**

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -1611,7 +1611,7 @@ void SmallPacket0x04D(map_session_data_t* session, CCharEntity* PChar, CBasicPac
         return;
     }
 
-    if (PChar->m_GMlevel == 0 && !PChar->loc.zone->CanUseMisc(MISC_AH) && !PChar->loc.zone->CanUseMisc(MISC_MOGMENU))
+    if (!zoneutils::IsResidentialArea(PChar) && PChar->m_GMlevel == 0 && !PChar->loc.zone->CanUseMisc(MISC_AH) && !PChar->loc.zone->CanUseMisc(MISC_MOGMENU))
     {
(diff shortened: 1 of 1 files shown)


src/map/packet_system.cpp
```diff
@@ -1611,7 +1611,7 @@ void SmallPacket0x04D(map_session_data_t* session, CCharEntity* PChar, CBasicPac
         return;
     }
 
-    if (PChar->m_GMlevel == 0 && !PChar->loc.zone->CanUseMisc(MISC_AH) && !PChar->loc.zone->CanUseMisc(MISC_MOGMENU))
+    if (!zoneutils::IsResidentialArea(PChar) && PChar->m_GMlevel == 0 && !PChar->loc.zone->CanUseMisc(MISC_AH) && !PChar->loc.zone->CanUseMisc(MISC_MOGMENU))
```
TeoTwawki, line 1614: was the situation not covere dby chekcing the misc flags? and nomad mogs say hi, same menu but you aren't inside a mog house

TeoTwawki, line 1614: is this intended to fix pulling up menu inside town but outside of mog house? ignore previous message, I'm still waking up I guess *yawns*

cocosolos: Actually the opposite, inside town and outside of moghouse is already blocked, but you're currently being blocked if you enter your moghouse from a zone that doesn't have these things, such as Port Bastok. I guess the moghouse is part of the last zone you were in? Not really sure how that all works, but you seem to use the misc flags of whatever zone you entered from.

TeoTwawki: yep. the mog house is part of the zone you entered from. I thought the town zones were flagged to allow and not block for this very reason.

ibm2431 (approved):

PR #722, opened 2020-06-14
Add var and binding for isDualWielding + remove presumptuous granting of mod to all NIN NMs

ibm2431:
As previously discussed, without adding dual wield mobs to the NIN NMs which _should_ have them, this will result in some NIN NMs becoming easier than they should be.

Here is a list of NMs which are currently automatically being given Dual Wield mobmod (note: _not trait_) which will make them _always_ have a second hit:
github

Some of these mobs should always be attacking twice; others should not. Unfortunately, _the mob's model **can not** be used to distinguish which is which_. Some NIN NMs with the _same_ model will always attack twice per round, and others only once. Determining which are "dual wielding" (ie: which will always attack twice per combat round) would require video captures from retail.

**Future selves: The dual wield mobmod is bad, and we need to make it go away.** Mobs don't wield weapons, and the assumption that they do leads to mistakes like the one currently being fixed in this PR (NIN NMs with a great sword model being given "dual wield"). We will need to add some sort of numHits column to the DB which dictates how many hits a mob has per attack round. This is _not_ the same as Double Attack, which procs on weaponskill/TP moves.

**_I affirm:_**

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -10015,6 +10015,30 @@ int32 CLuaBaseEntity::checkImbuedItems(lua_State* L)
 }
 
 /************************************************************************
+*  Function: isDualWielding()
+*  Purpose : Returns true if entity is wielding two weapons
+*  Example : if player:isDualWielding() then
+*  Notes   : 
+************************************************************************/
+
+inline int32 CLuaBaseEntity::isDualWielding(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    CBattleEntity* PBattleEntity = dynamic_cast<CBattleEntity*>(m_PBaseEntity);
+    if (PBattleEntity)
+    {
+        lua_pushboolean(L, PBattleEntity->m_dualWield);
+        return 1;
+    }
+    else
+    {
+        lua_pushboolean(L, false);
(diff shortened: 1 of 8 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -9924,6 +9924,12 @@ int32 CLuaBaseEntity::checkImbuedItems(lua_State* L)
...
 {
     TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    if (m_PBaseEntity->objtype == TYPE_NPC)
+    {
+        lua_pushboolean(L, false);
+        ShowError("lua::isDualWielding :: NPCs don't wield weapons!\n");
+        return 1;
+    }
 
     CBattleEntity* PBattle = (CBattleEntity*)m_PBaseEntity;
```
TeoTwawki: Reminder for tomorrow about c-style cast thing we discussed in discord, but otherwise :+1:

TeoTwawki (approved):

cocosolos:
Regarding the dual wield mobmod, would `tpz.mobMod.MULTI_HIT` work instead?

ibm2431:
I think that's similar to the idea that we were going to go with, but with a db column in mob_pool instead, as an alternative to defining a mobmod for individual mobs or more entries in pool_mobmods. Teo and I have gone in circles over how to handle multi-hit mobs... multiple... times, and at this point I'm not 100% positive what if anything was decided on.

cocosolos (approved):

PR #725, opened 2020-06-15
Double-up cooldown adjustments

tankfest:
**_I affirm:_**

Removes the cooldown added to double-up after a corsair uses a roll.  Also reduces the cooldown on double-up in line with the retail change to 5 seconds.

### DIFF
--- a/scripts/globals/ability.lua
+++ b/scripts/globals/ability.lua
@@ -467,7 +467,7 @@ function corsairSetup(caster, ability, action, effect, job)
     caster:setLocalVar("corsairRollTotal", roll)
     action:speceffect(caster:getID(), roll)
     if (checkForElevenRoll(caster)) then
-        action:recast(action:recast()/2)
+        action:recast(action:recast()/2) -- halves phantom roll recast timer for all rolls while under the effects of an 11 (upon first hitting 11, phantom roll cooldown is reset in double-up.lua)
     end
     checkForJobBonus(caster, job)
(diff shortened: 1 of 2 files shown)


ibm2431 (approved):
I'm assuming the recast for Phantom Roll itself is added by core... somewhere.

If you want to go the extra mile:
```lua
    if (checkForElevenRoll(caster)) then
        action:recast(action:recast()/2)
    end
```
Is wrong. An 11 roll should instantly cause Phantom Roll to be available again, not at half the current recast time.

tankfest:
> 
> 
> I'm assuming the recast for Phantom Roll itself is added by core... somewhere.
> 
> If you want to go the extra mile:
> 
> ```lua
>     if (checkForElevenRoll(caster)) then
>         action:recast(action:recast()/2)
>     end
> ```
> 
> Is wrong. An 11 roll should instantly cause Phantom Roll to be available again, not at half the current recast time.

So does only the _next_ phantom roll get 0 cooldown, or do all phantom rolls while under the effects of an 11 give 0 cooldown?

edit: it's on wiki -- Rolling a total of 11 will immediately reset the Phantom Roll ability timer. While under the effect of an 11-valued roll, your Phantom Roll recast timer will be reduced by 50%. In addition, any rolls of 12+ will still remove the roll's effect, but you will not suffer a Bust penalty and may re-roll freely once your recast timer is up.

PR #728, opened 2020-06-15
Fixed spawn window on HAWKEYED_DNATBAT

Noikar:
**_I affirm:_**

Fixed spawn window (was 1 hr when it should have been 1-3 hrs)


cocosolos:
I think since this is a lottery spawn then it only needs the minimum time value.

TeoTwawki:
as presently in release branch, it'd just have a 10% shot at popped after the 1hr cooldown passes, which could be 1 spawn cycle or 500 spawn cycles. Some other NMs have a variable randomly chosen window open point, but there's been no decided standard thus far for how to deal with wiki's purely anecdotal accounts of "it took me 6 hours to pop this guy!" and we do not know if an upper limit even exists (if an upper limit does exist, we shoudl be forcing them to spawn next cycle when that window ends)

cocosolos:
Are there any lottery mobs we do this with? I know we do it with some timed mobs like Fafnir, but imo all lottery mobs should just have their minimum time, everything else is just bad luck.

Noikar:
Okay the wiki showed 1-3 so I was thinking that the spawn window was not
guaranteed after one hour, I thought it was determined by a time in the
window. I wasn't sure how it if this is not how we do it, you can close the
pr

TeoTwawki:
> 
> 
> Are there any lottery mobs we do this with? I know we do it with some timed mobs like Fafnir, but imo all lottery mobs should just have their minimum time, everything else is just bad luck.

Several, but I think we _should not._
https://github.com/project-topaz/topaz/blob/8985d2af68876367592222836db1f8e5b15c8175/scripts/zones/Beaucedine_Glacier/mobs/Tundra_Tiger.lua#L15

PR #735, opened 2020-06-16
Fixes pet not despawning in combat when master dies.

tankfest:
Co-contribution credit for diagnosis and fix to cocosolos!

**_I affirm:_**

DoRoamTick is where the original despawn pet code was when the master dies, this doesn't fire as intended while the pet is in combat.  It's been improved and moved to on Tick so that it's always being checked, regardless of pet combat status.

Tested with BST charmed/jug, shiva, puppet, and wyvern.

### DIFF
--- a/src/map/ai/controllers/pet_controller.cpp
+++ b/src/map/ai/controllers/pet_controller.cpp
@@ -36,21 +36,17 @@ CPetController::CPetController(CPetEntity* _PPet) :
 
 void CPetController::Tick(time_point tick)
 {
-    if (PPet->isCharmed && tick > PPet->charmTime)
+    if ((PPet->isCharmed && tick > PPet->charmTime) || ((PPet->PMaster == nullptr || PPet->PMaster->isDead()) && PPet->isAlive()))
     {
         petutils::DespawnPet(PPet->PMaster);
         return;
     }
+    
     CMobController::Tick(tick);
 }
 
 void CPetController::DoRoamTick(time_point tick)
 {
-    if ((PPet->PMaster == nullptr || PPet->PMaster->isDead()) && PPet->isAlive()) {
-        PPet->Die();
-        return;
-    }
-
     //automaton, wyvern
     if (PPet->getPetType() == PETTYPE_WYVERN || PPet->getPetType() == PETTYPE_AUTOMATON) {
         if (PetIsHealing()) {


TeoTwawki:
_Carbuncle: No! I will avenge my summoner!_

:rofl:

kaincenteno:
that's awesome =) we just experienced this today

ibm2431 (commented):
So we talked about this a bit on Discord, interested parties can search the server for the first occurrence of `DetachPet` for the start of a longish conversation about orphans.

To summarize my thoughts:
- [despawnPet in petutils literally only calls detachPet](https://github.com/project-topaz/topaz/blob/release/src/map/utils/petutils.cpp#L1249-L1254) and can be removed in core (be sure to not delete [the lua binding itself](https://github.com/project-topaz/topaz/blob/release/src/map/lua/lua_baseentity.cpp#L12066-L12083))
- I don't think any alive checks are necessary in pets' ticks, as [the master battle entity's die() will be called prior to the next roam or engaged tick](https://github.com/project-topaz/topaz/blob/release/src/map/entities/battleentity.cpp#L1772-L1779)
- In [Charentity die()](https://github.com/project-topaz/topaz/blob/release/src/map/entities/charentity.cpp#L1659-L1660), call detachPet; this function [already handles](https://github.com/project-topaz/topaz/blob/release/src/map/utils/petutils.cpp#L1164-L1173) various [pet types](https://github.com/project-topaz/topaz/blob/release/src/map/utils/petutils.cpp#L1209-L1216)
- detachPet can't go in mob beastmaster entities due to it (correctly) [throwing a debug break if master isn't a player](https://github.com/project-topaz/topaz/blob/release/src/map/utils/petutils.cpp#L1155-L1158) due to all the [player specific work](https://github.com/project-topaz/topaz/blob/release/src/map/utils/petutils.cpp#L1236-L1240) the function does; so leave detachPet out of [battleentity die](https://github.com/project-topaz/topaz/blob/release/src/map/entities/battleentity.cpp#L1209-L1210) (and any controllers which might inadvertently call detachPet on a mob)
- [mobentities which die currently check if they have a pet, and invoke die() on that pet directly if it's an avatar](https://github.com/project-topaz/topaz/blob/release/src/map/entities/mobentity.cpp#L1067-L1075); this is correct, but _unengaged_ non-avatar pets will also need a die() (see Discord discussion for videos - search: `Phomiuna`)

ibm2431:
Alternatively, `despawnPet` can stay in petutils, and then the type of the master can be checked as a player before calling `detachPet`. We could then call the pet's die() directly here (in all cases) inside `despawnPet`, allowing `despawnPet` to be used by both mob masters and player masters. This would also allow the binding in `lua_baseentity` to be used on mobs without crashing the server.

ibm2431:
I didn't consider this last night, but aggroed orphaned pets will despawn if aggro is lost through Hide or Warp. So if beastmen pets are governed by pet_controller, we might still need checks in roamTick (still make sure `DetachPet` doesn't get called with a beastman parent).

zircon-tpl:
@project-topaz/developer : @tankfest has requested assistance in making suggested changes and opening a new Pull Request.

PR #736, opened 2020-06-17
Adds scripting and DB edits for NM Backoo

tankfest:
**_I affirm:_**

I'll be doing live testing on Gold Saucer over the next few days for this; holding here for critiquing as well.

Thanks to cocosolos for assistance with spawn conditions!

### DIFF
--- a/scripts/zones/Buburimu_Peninsula/Zone.lua
+++ b/scripts/zones/Buburimu_Peninsula/Zone.lua
@@ -4,60 +4,84 @@
 --
 -----------------------------------
 local ID = require("scripts/zones/Buburimu_Peninsula/IDs")
-require("scripts/globals/icanheararainbow");
-require("scripts/globals/chocobo_digging");
-require("scripts/globals/conquest");
+require("scripts/globals/icanheararainbow")
+require("scripts/globals/chocobo_digging")
+require("scripts/globals/conquest")
 require("scripts/globals/helm")
-require("scripts/globals/zone");
+require("scripts/globals/zone")
 -----------------------------------
 
 function onChocoboDig(player, precheck)
     return tpz.chocoboDig.start(player, precheck)
-end;
+end
 
 function onInitialize(zone)
+    local hour = VanadielHour()
+    
+    if hour >= 6 and hour < 16 then
+        GetMobByID(ID.mob.BACKOO):setRespawnTime(1)
+    end
+
     tpz.conq.setRegionalConquestOverseers(zone:getRegionID())
 
     tpz.helm.initZone(zone, tpz.helm.type.LOGGING)
-end;
+end
 
-function onZoneIn( player, prevZone)
-    local cs = -1;
+function onZoneIn(player, prevZone)
+    local cs = -1
 
-    if (player:getXPos() == 0 and player:getYPos() == 0 and player:getZPos() == 0) then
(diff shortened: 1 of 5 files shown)


scripts/zones/Buburimu_Peninsula/Zone.lua
```diff
@@ -4,60 +4,81 @@
...
 function onChocoboDig(player, precheck)
     return tpz.chocoboDig.start(player, precheck)
-end;
+end
 
 function onInitialize(zone)
+    local hour = VanadielHour()
+    
+    if hour >= 6 and hour < 16 then
+        DisallowRespawn(ID.mob.BACKOO, false)
```
cocosolos: Should just need to change this to `GetMobByID(ID.mob.BACKOO):setRespawnTime(1)`

scripts/zones/Buburimu_Peninsula/Zone.lua
```diff
@@ -4,60 +4,81 @@
...
 function onRegionEnter(player,region)
-end;
+end
+
+function onGameHour(zone)
+    local hour = VanadielHour()
+    local nmBackoo = GetMobByID(ID.mob.BACKOO)
+
+    if hour == 6 then -- backoo time-of-day pop condition open
+        DisallowRespawn(ID.mob.BACKOO, false)
```
cocosolos, line 58:
```lua
local backoo = GetMobByID(ID.mob.BACKOO)
if backoo:getRespawnTime() == 0 then
    backoo:setRespawnTime(1)
else
    DisallowRespawn(ID.mob.BACKOO, false)
end
```

cocosolos (approved):

sql/mob_groups.sql
```diff
@@ -7927,6 +7927,7 @@ INSERT INTO `mob_groups` VALUES (51,3148,118,'Pixie',0,128,2001,0,0,51,54,0);
 INSERT INTO `mob_groups` VALUES (52,2213,118,'Ketos',1,0,0,0,0,1,1,0);
 INSERT INTO `mob_groups` VALUES (53,4734,118,'Botulus_Rex',0,128,0,0,9999,94,95,0);
 INSERT INTO `mob_groups` VALUES (54,1642,118,'Goblin_Bounty_Hunter',0,0,1030,0,0,17,20,0);
+INSERT INTO `mob_groups` VALUES (55,5736,118,'Backoo',0,128,3140,2500,0,37,37,0);
```
ibm2431, line 7930: Might need a different pool. This is the pool for Koropokkur, which has a skill list of _only_ Wild Oats.

tankfest:
It should be noted that its spawn position input in the database is a rough guesstimate based on what ffxiclopedia had to say.  Could use a capture (should be fairly easy given his regular spawn timing) to determine more accurate spawn coords.

edit: Nyu provided a retail capture for his position, thanks!

sql/mob_groups.sql
tankfest: Swapped him to Tom Tit Tat's pool, who uses no special abilities outside of standard mandy moves.

TeoTwawki: but then its using all the same pool data as Tom, which may not be correct. More than likely a flag value will be different (probably size bits) and those are stored in pool. Recommend just making a new pool for new mobs.

PR #738, opened 2020-06-17
Fix Half Partition synth level requirements

tankfest:
**_I affirm:_**

Thanks to Eren@GoldenSaucer for bringing this up.


cocosolos:
[Bg-wiki](https://www.bg-wiki.com/bg/Half_Partition) and [jp-wiki](http://wiki.ffo.jp/html/17287.html) both say WW 88. Only wikia has WW 83 listed with a subskill. I would say drop the subskill but leave the WW at 88.

PR #742, opened 2020-06-17
fixing merit values based on 1000 tp model and fixing overflow

59blargedy:
Savagery and enchainment were giving 10 tp bonus instead of 100. meritvalue variable set at int8 caused overflow issues with change, so modified to 16 to accommodate.

Did not note any areas where the overflow change would cause issues.

**_I affirm:_**


ibm2431 (approved):
Merits in char_merits are stored as charid -> merit id, number of upgrades. Not a blob, so we don't need to worry about size expansion.

GetMeritValue here only returns the bonus value as tabulated by how many upgrades in a category which a character has, multiplied by the value granted by the merit. (And some caps and level checks are applied).

Reason you needed to expand the size is that the max size of uint8 is 256. And 5 * 100 = 500, so...!

Anything affected by getMeritValue returning uint16 instead of uint8 was already using too small of a space for whatever it's calculating.

edit: These merits currently don't seem to be used anywhere, so we don't need to worry about existing code already doing `GetMeritValue(MERIT_ENCHAINMENT) * 10`.

PR #744, opened 2020-06-18
Fix BRD song overwrite bug

InoUno:
This PR fixes a bug where BRDs are able to keep up many more than the allowed max amount of songs.

With the current implementation, if BRD A overwrites a song of the same tier/type from BRD B, the check for maximum amount of songs allowed for BRD A is never performed, since it just overwrites it immediately.
This leaves BRD B free to sing a new song, since he doesn't own that active song anymore. Rinse and repeat, and you can get as many songs active as the casting time allows with just 2 BRDs.

Example:
```
A sings March1
A sings March2
B sings Minuet1
B sings Minuet2
A overwrites Minuet1
A overwrites Minuet2
[at this point there's 4 songs up, all owned by BRD A]
B sings Madrigal1
B sings Madrigal2
[at this point there's 6 songs up between 2 BRDs]
[rinse and repeat, with BRD A overwriting songs and BRD B singing new songs]
```

**_I affirm:_**


ibm2431 (approved):
Sorry for the delay, InoUno!

cocosolos (approved):

PR #746, opened 2020-06-18
Updated Binary to let RUN use their innate spells

mrmclaw:
Updated binary code to allow RUN to cast spells they have. 

Crusade, and Regen IV are not currently coded and not able to update for run specifically. Foil is still commented out and I'm not sure if that is because RUN is technically NYI at this time or because something is wrong with it.

Utilized windows 10 programmer calculator to convert. (Thank you for the tip Mr. Sent)

**_I affirm:_**

Thank you~!


Omnione:
I checked all the numbers compared to what wiki has, they all check out, good job

TeoTwawki:
> binary code 

Job mask. It's a hexadecimal mask for what player jobs can use (mobs don't obey this just players)

cocosolos (approved):

ibm2431 (approved):
Thanks, mrmclaw!

PR #751, opened 2020-06-20
Added Scholar AF quest Seeing Blood-Red

Noikar:
**_I affirm:_**

BCNM still needs to be coded.

### DIFF
--- a/scripts/zones/Grauberg_[S]/npcs/34.lua
+++ b/scripts/zones/Grauberg_[S]/npcs/34.lua
@@ -1,0 +1,31 @@
+-----------------------------------
+-- Area: Grauberg [S]
+--  NPC:
+--  Quest - Seeing Blood-red
+-- pos -283 -55 -97
+-----------------------------------
+local ID = require("scripts/zones/Grauberg_[S]/IDs")
+require("scripts/globals/quests")
+require("scripts/globals/keyitems")
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    if player:getCharVar("SeeingBloodRed") == 3 then
+        player:startEvent(14)
+    end
+end
+
+function onEventUpdate(player,csid,option)
+
+end
+
+function onEventFinish(player,csid,option)
+    if csid == 14 then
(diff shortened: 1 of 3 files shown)


scripts/zones/Grauberg_[S]/npcs/34.lua
```diff
@@ -0,0 +1,31 @@
...
+    end
+end
+
+function onEventUpdate(player,csid,option)
+
+end
+
+function onEventFinish(player,csid,option)
+    if csid == 14 then
+      --player:delKeyItem(tpz.ki.PORTING_MAGIC_TRANSCRIPT)
```
Comment (Noikar) [inline scripts/zones/Grauberg_[S]/npcs/34.lua:27]: Once the BCNM is in, this should be uncommented, as it consumes the transcript on cutscene finish.

scripts/zones/Grauberg_[S]/npcs/34.lua
```diff
@@ -0,0 +1,31 @@
...
+      --TODO: Teleport player to Ruhotz Silvermines
```
Comment (Noikar) [inline scripts/zones/Grauberg_[S]/npcs/34.lua:28]: The landing place needs to be determined.

scripts/zones/Grauberg_[S]/npcs/34.lua
```diff
@@ -0,0 +1,31 @@
...
+      --AFTER BCNM:player:setCharVar("SeeingBloodRed",5)
```
Comment (Noikar) [inline scripts/zones/Grauberg_[S]/npcs/34.lua:29]: At the end of the BCNM fight, this line should execute. I put it here because the BCNM isn't scripted. Once this is set, the quest will continue as I've written it.

scripts/zones/Grauberg_[S]/npcs/34.lua
```diff
@@ -0,0 +1,31 @@
...
+require("scripts/globals/quests")
+require("scripts/globals/keyitems")
+-----------------------------------
+
+function onTrade(player,npc,trade)
+end
+
+function onTrigger(player,npc)
+    if player:getCharVar("SeeingBloodRed") == 3 then
+        --player:startEvent(14)
```
Comment (Noikar) [inline scripts/zones/Grauberg_[S]/npcs/34.lua:17]: Once the BCNM is in, this should be uncommented so the BCNM-teleportation cutscene can be triggered.

PR #753, opened 2020-06-22
Fix wrong boot arg read

cocosolos:
**_I affirm:_**


Lynnea1320:
What the Hecate does this change?

PR #756, opened 2020-06-22
Remove automatic multipliers when mob is marked as an NM

ibm2431:
This was entirely made up.

NMs which have been CONFIRMED to have HP values that aren't in line with the standard growth scale should have their HP manually set instead.

**_I affirm:_**

### DIFF
--- a/src/map/utils/mobutils.cpp
+++ b/src/map/utils/mobutils.cpp
@@ -342,14 +342,6 @@ void CalculateStats(CMobEntity * PMob)
         if(PMob->MPmodifier == 0)
         {
             PMob->health.maxmp = (int16)(18.2 * pow(mLvl,1.1075) * scale) + 10;
-            if(isNM)
-            {
-                PMob->health.maxmp = (int32)(PMob->health.maxmp * 1.5f);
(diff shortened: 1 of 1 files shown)


59blargedy (approved):

TeoTwawki:
There was a 1.5x stat multiplier that was made up too, dunno if still exists

cocosolos:
I didn't see any stat bonuses but there is also an MP bonus applied right after this, dunno if you want to get that as well.

ibm2431:
I was originally tempted to remove the stat multiplier, but expected its removal to be met with resistance, so stuck with just the HP since there's a specific config setting I could point to.

But if we all want to remove the stat multiplier too, I will _happily_ delete it.

I didn't see the MP bonus, I'll delete that too. (Although "mob MP" will be a discussion for another time.)

edit: Oh, there's an NM stat multiplier in map.conf already. So out the vanilla stat multiplier goes.

cocosolos:
Didn't even see those multipliers next to the conf settings, good eye!

cocosolos (approved):

PR #757, opened 2020-06-22
[Fix] Outpost teleport (Unlock) options 1/2

Mortium:
With the unlocked option 1 outpost teleport the following showed up in the menu but did not work:
Li'Telor, Kuzotz, Vollbow, Elshimo Lowlands, Elshimo Uplands. (Animation happens, no currency taken and no teleport)

With OP teleports unlocked option 2 you should have access to all including: Tu'Lia, Tavnazia

**_I affirm:_**
- [] that I agree to Project Topaz's [Limited Contributor License Agreement](http://project-topaz.com/blob/release/CONTRIBUTOR_AGREEMENT.md), as written on this date


cocosolos:
This part should already be behaving properly, all you've done is make it allow all outposts if the setting is either 1 or 2. Do you meet the minimum level requirement for the outpost? If so the problem may be in `tpz.conquest.teleporterOnEventFinish` or possibly `teleports.lua`.

Mortium:
If you try to use it ingame, it does not work with the previous code. Like this it does. It only actually teleports to: Ronfaure, Zulkheim, Norvallen, Gustaberg, Derfland, Sarutabaruta, Kolshushu, Aragoneu, Fauregandi, Valdeaunia and Qufim Island regions. None of the above I posted will actually teleport you anywhere, they just show up on the list.

TeoTwawki:
~~Sidenote: I’d really love somebody making the setting just cause the event to treat you as having these unlocked, instead of its present form of causing nee players to start with them unlocked.~~

~~But I’m able to teleport just fine with the current code, which had not changed in awhile.~~

Change I was talking about happened and I didn’t know/remember ¯\\\_(ツ)\_/¯

Mortium:
> Sidenote: I’d really love somebody making the setting just cause the event to treat you as having these unlocked, instead of its present form of causing nee players to start with them unlocked.
> 
> But I’m able to teleport just fine with the current code, which had not changed in awhile.

This only applies to the unlocked options 1 or 2, not the new player no unlocked version.

cocosolos:
> Sidenote: I’d really love somebody making the setting just cause the event to treat you as having these unlocked, instead of its present form of causing nee players to start with them unlocked.

It should already be doing this. If you don't have the outpost (new players don't even with this setting) then it checks if the settings are enabled and just returns true.

Mortium:
In settings there are 3 options:

UNLOCK_OUTPOST_WARPS = 0 -- Set to 1 to give starting characters all outpost warps.  2 to add Tu'Lia and Tavnazia.

cocosolos:
1 is only supposed to be all outposts except Tu'Lia and Tavnazia, the way this is being done allows all outposts with either 1 or 2, there is no difference.

Mortium:
It does not work if you try it with the current code. The way I changed it, it actually does. :) Been tested.

cocosolos:
Ok but your fix is to change what the setting means, creating a redundancy with options 1 and 2, and disabling the option of allowing all teleports except Tu'Lia and Tavnazia.

Mortium:
With option 1 Tu'lia and Tavnazia do not show up on the list at the NPC the way I fixed it, with option 2, it shows and allows Tu'Lia and Tavnazia.

TeoTwawki:
> This only applies to the unlocked options 1 or 2, not the new player no unlocked version.

I know what the options are and was not at all commenting on what you did here. It just turns out the changes I wished for happened and I never got the memo.

cocosolos:
Those options show up in the list because of `getAllowedTeleports`, `hasOutpost` is only used in the event finish if you have those options enabled.

scripts/globals/conquest.lua
```diff
@@ -85,10 +85,10 @@ local function hasOutpost(player, region)
     local region = region + 5
     local hasOP = player:hasTeleport(player:getNation(), region)
     if not hasOP then
-        if UNLOCK_OUTPOST_WARPS == 2 then
+        if UNLOCK_OUTPOST_WARPS == 1 then
```
TeoTwawki, line 88:
```
        if UNLOCK_OUTPOST_WARPS == 1 then
            hasOP = true
```
if setting is 1, the check for if you have access is always true. that is what this says. That is not what should happen here.

Mortium:
> > This only applies to the unlocked options 1 or 2, not the new player no unlocked version.
> 
> I know what the options are and was not at all commenting on what you did here. It just turns out the changes I wished for happened and I never got the memo.

Some servers are a bit more quality of life style like newer MMO's (like the one I play on as well) sp they do allow more freedom but the good thing is that there is the options at least.

scripts/globals/conquest.lua
Mortium: I honestly can only say to try to use the current code or what I have posted. The current 1 doesn't allow Li'Telor, Kuzotz, Vollbow, Elshimo Lowlands, Elshimo Uplands. (Animation happens, no currency taken and no teleport) Why that is is maybe due to quest requirement besides the supply missions to be in those areas.

TeoTwawki:
> Some servers are a bit more quality of life style like newer MMO's (like the one I play on as well) sp they do allow more freedom but the good thing is that there is the options at least.

what does that even have to do with what I said??

I dunno what you think you read but I guarantee you misunderstood

scripts/globals/conquest.lua
TeoTwawki: this is what the code does. indisputably with even basic understanding of lua. The ONLY instance that should be always true is the value of 2 you changed it away from. You do not get access to tav+sky without:
setting=2, OR, actually have the unlock legit (and in one of those, there is no legit unlock)

Mortium:
> I dunno what you think you read but I guarantee you misunderstood

Possible, if so my apologies.

scripts/globals/conquest.lua
Mortium: With option 1 Tu'lia and Tavnazia do not show up on the list at the NPC.

cocosolos: Because the list is generated by `getAllowedTeleports`, not `hasOutpost`.

Mortium: So my question is I guess, why is there even a restriction on options 1 or 2 when it comes to `hasOutpost `? The list comes from `getAllowedTeleports` the last two teleports for option 2 won't be available anyways unless you choose that option. And the point of them is to have access to everything in those option 1 or 2 lists. I mean, it works like this.

cocosolos: `hasOutpost` is used in other places to calculate fees and in the event finish to verify that that the packet was not altered and the players choice is indeed a valid choice from the menu.

Mortium: `        elseif UNLOCK_OUTPOST_WARPS == 1 then
            hasOP = region <= tpz.region.ELSHIMOUPLANDS`

I ment this, not fees and such, I understand that.

cocosolos: The only outpost regions above Elshimo Uplands is Tu'Lia and Tavnazia, so if UNLOCK_OUTPOST_WARPS is 1 then it will return true for everything except those 2 regions, so if for some reason a player is able to select those options even though they're not in the menu, they will still be rejected.

Mortium: But it does not work for everything below, nothing above Quifim Island actually teleports you anywhere with option 1.

cocosolos: Okay the problem seems to be the `local region = region +5`. I'm not sure if this +5 is needed for the `hasTeleport` call, but it is screwing up the region comparison if `UNLOCK_OUTPOST_WARPS == 1`.

cocosolos:
```lua
local function hasOutpost(player, region)
    local hasOP = player:hasTeleport(player:getNation(), region + 5)
    if not hasOP then
        if UNLOCK_OUTPOST_WARPS == 2 then
            hasOP = true
        elseif UNLOCK_OUTPOST_WARPS == 1 then
            hasOP = region <= tpz.region.ELSHIMOUPLANDS
        end
    end
    return hasOP
end
```
that probably fixes the problem without breaking anything, assuming everything was working prior to [this change](https://github.com/project-topaz/topaz/commit/6507e3d88db53d750858af4f2978c46c4fbef8f7).

Mortium: I will test it out right now. Thanks!

Mortium: > ```lua
> local function hasOutpost(player, region)
>     local hasOP = player:hasTeleport(player:getNation(), region + 5)
>     if not hasOP then
>         if UNLOCK_OUTPOST_WARPS == 2 then
>             hasOP = true
>         elseif UNLOCK_OUTPOST_WARPS == 1 then
>             hasOP = region <= tpz.region.ELSHIMOUPLANDS
>         end
>     end
>     return hasOP
> end
> ```
> 
> that probably fixes the problem without breaking anything, assuming everything was working prior to [this change](https://github.com/project-topaz/topaz/commit/6507e3d88db53d750858af4f2978c46c4fbef8f7).

This seems to be working @cocosolos ! Would you mind commiting this please? :)

kaincenteno:
what happened her why did it get closed? =(

PR #760, opened 2020-06-23
Change NM hunt scripts to pass RegimeID

Kreidos:
Sister PR to hunts.lua + other changes coming from Cocosolos/Teo.

**_I affirm:_**

### DIFF
--- a/scripts/zones/Yuhtunga_Jungle/mobs/Mischievous_Micholas.lua
+++ b/scripts/zones/Yuhtunga_Jungle/mobs/Mischievous_Micholas.lua
@@ -7,7 +7,7 @@ require("scripts/globals/regimes")
 -----------------------------------
 
 function onMobDeath(mob, player, isKiller)
-    tpz.hunts.checkHunt(mob, player, isKiller)
+    tpz.hunts.checkHunt(mob, player, 362)
     tpz.regime.checkRegime(player, mob, 126, 1, tpz.regime.type.FIELDS)
     tpz.regime.checkRegime(player, mob, 128, 1, tpz.regime.type.FIELDS)
 end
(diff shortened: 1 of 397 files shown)


cocosolos (approved):

scripts/zones/Yuhtunga_Jungle/mobs/Mischievous_Micholas.lua
```diff
@@ -7,7 +7,7 @@ require("scripts/globals/regimes")
 -----------------------------------
 
 function onMobDeath(mob, player, isKiller)
-    tpz.hunts.checkHunt(mob, player, isKiller)
+    tpz.hunts.checkHunt(mob, player, 362)
```
ibm2431, line 10: Seeing:
```lua
tpz.hunts.checkHunt(mob, player, 362)
tpz.regime.checkRegime(player, mob, 126, 1, tpz.regime.type.FIELDS)
```
Makes me want to say that we should establish a standardized argument order at some point to avoid inadvertently crossing wires. "Such and such arguments always come before this other argument". There's a lot of combat functions which waffle between `attacker, defender` and `defender, attacker` too.

ibm2431 (approved):
Thanks, Kreidos!

scripts/zones/Yuhtunga_Jungle/mobs/Mischievous_Micholas.lua
TeoTwawki: Personal opinion: checkHunt does it better visually because of the order onMobDeath uses. Just “feels right”.

Edit: altho I did originally envision the existing GoV and FoV global being refactored to kill that loop it has and accommodate both hunts and dominion ¯\\\_(ツ)\_/¯

Kreidos: Indeed, I presume why @Apocarypse structured it so originally. :+1:

PR #766, opened 2020-06-24
Fix True Will title

ibm2431:
> Completing the ninja artifact three quest called "True will" did not change my title. Player title should be changed to "Paragon of ninja excellence", when quest is completed.

**_I affirm:_**
- [🤞] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

### DIFF
--- a/scripts/zones/Rabao/npcs/Leodarion.lua
+++ b/scripts/zones/Rabao/npcs/Leodarion.lua
@@ -4,11 +4,13 @@
 -- Involved in Quest: 20 in Pirate Years, I'll Take the Big Box, True Will
 -- !pos -50 8 40 247
 -----------------------------------
-require("scripts/globals/settings");
-require("scripts/globals/keyitems");
-require("scripts/globals/shop");
-require("scripts/globals/quests");
-local ID = require("scripts/zones/Rabao/IDs");
+require("scripts/globals/settings")
+require("scripts/globals/keyitems")
+require("scripts/globals/npc_util")
+require("scripts/globals/quests")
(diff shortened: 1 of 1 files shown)


scripts/zones/Rabao/npcs/Leodarion.lua
```diff
@@ -73,15 +75,14 @@ function onEventFinish(player,csid,option)
...
-            player:addFame(NORG,30);
-            player:completeQuest(OUTLANDS,tpz.quest.id.outlands.TRUE_WILL);
+        if npcUtil.completeQuest(player, OUTLANDS, tpz.quest.id.outlands.TRUE_WILL, {
+                item = 13782, -- Ninja Chainmail
+                fameArea = NORG,               
+                title = tpz.title.PARAGON_OF_NINJA_EXCELLENCE,
+                var = "trueWillCS"
+            })
+        then
+            player:delKeyItem(tpz.ki.LARGE_TRICK_BOX)
```
Kreidos, line 85: I think it's possible to specify 2 key items in completeQuest(), is there a technical reason this should happen only after?

Kreidos (commented):
Looks good! Just one small question with deleting key items. :)

scripts/zones/Rabao/npcs/Leodarion.lua
Kreidos: Sorry, I misunderstood this is taking an item whereas completeQuest is giving one.

Kreidos (approved):

cocosolos (approved):

PR #771, opened 2020-06-25
Added Gyre-Carlin

Noikar:
**_I affirm:_**


scripts/zones/PsoXja/mobs/Diremite.lua
```diff
@@ -0,0 +1,12 @@
...
+-- Area: Pso'Xja
+--  Mob: Diremite
+-----------------------------------
+local ID = require("scripts/zones/PsoXja/IDs");
+
+function onMobDeath(mob, player, isKiller)
+end;
+
+function onMobDespawn(mob)
+    tpz.mob.phOnDespawn(mob, ID.mob.GYRE_CARLIN_PH, 5, math.random(3600*4, 3600*5)) -- 4-5 hours.
```
cocosolos: Where did you get 4-5 hours? I'm only seeing as low as 15-30 mins on wikia, and again I don't think lottery mobs should have an upper bound on their spawn time.

scripts/zones/PsoXja/IDs.lua
```diff
@@ -31,6 +31,16 @@ zones[tpz.zone.PSOXJA] =
...
     {
+        GYRE_CARLIN_PH =
+        {
+            [16814317] = 16814320, -- -66.201 -0.352 189.181
+            [16814318] = 16814320, -- -56.059 -0.352 186.055
+            [16814324] = 16814331, -- -13.760 -0.352 224.323
+            [16814325] = 16814331, --  -6.011 -0.352 205.851
+            [16814328] = 16814344, --   9.065 -0.352 168.035
+            [16814329] = 16814344, --  34.530 -0.352 165.232
+            [16814330] = 16814344, --  34.057 -0.352 165.306
```
cocosolos: It seems that none of these NM IDs exist. Also the only placeholder seems to be 16814330 and is supposed to spawn in a few different locations (there's only 1 room with 3 Diremites, the 3rd one in that room is 16814330 and is supposed to spawn in a few different places.)

cocosolos: Sorry I might actually be wrong about this, they're there they just don't have any spawn points so they're not being loaded at all. Same with the other PHs, looks like there are multiples (assuming only 1 is supposed to be up at a time) but none of the other ones have spawn points set.

cocosolos: PHs should be:
```lua
[16814319] = 16814320,
[16814330] = 16814331,
[16814343] = 16814344,
```

scripts/zones/PsoXja/mobs/Diremite.lua
Noikar: Oh sorry, I thought I pasted the clean version not the Dawnbreak version.

cocosolos (approved):

ibm2431 (approved):

PR #774, opened 2020-06-26
Full Speed Ahead: Various Improvements

zach2good:
**_I affirm:_**

Tested a bunch, I'm getting pretty good at the minigame...

**New:**
Added tracking via Quest Log
Fixed missing Item and NPC IDs in completion CS
Changed motivation/pep relationship
Added clamping to motivation and pep to fix any weirdness with the bar rendering
Added failure CS when you dismount or run out of motivation, which lets you enable easy mode (or not)
Added easy mode: tracked through the effect's power (acts as a modifier for motivation decay and pep growth)
Fixed bug where the final Raptor Food region was acting like it was Syrillia's region
Added MEET_SYRILLIA text when you've eaten 5 foods

**Not doing in this PR:**
Additional Raptor animations
Varying pep and speed depending on terrain slope etc.
Best time tracking

**Once this PR goes in, I think it would be fair to open two new issues:**
**- [Good First Issue] Quest: Full Speed Ahead! doesn't track server best time**
Using `Get/SetServerVariable`

**- [Good First Issue][Core] Quest: Full Speed Ahead! raptor doesn't slow down as it runs out of pep**
Mount speed is hard-coded, and needs to be modifiable (through mods or otherwise) for Full Speed Ahead!

`battleentity.cpp`
```
uint8 CBattleEntity::GetSpeed()
{
    return (isMounted() ? 50 + map_config.speed_mod : std::clamp<uint16>(speed * (100 + getMod(Mod::MOVE)) / 100, std::numeric_limits<uint8>::min(), std::numeric_limits<uint8>::max()));
}
```

### DIFF
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -481,6 +481,8 @@ tpz.quest.id =
         VW_OP_115_VALKURM_DUSTER        = 168,
         VW_OP_118_BUBURIMU_SQUALL       = 169,
         PRELUDE_TO_PUISSANCE            = 170,
+
+        FULL_SPEED_AHEAD                = 179, -- + --
     },
 
     -----------------------------------
--- a/scripts/quests/full_speed_ahead.lua
+++ b/scripts/quests/full_speed_ahead.lua
@@ -3,7 +3,24 @@
 -----------------------------------
 local ID = require("scripts/zones/Batallia_Downs/IDs")
 require("scripts/globals/status")
+require("scripts/globals/utils")
+require("scripts/globals/zone")
 -----------------------------------
+
+--[[
+Debugging:
+Reset: !delquest 3 179
+
+CharVar: [QUEST]FullSpeedAhead == ...
+1 : Starting minigame on Normal Mode
+2 : Starting minigame on Easy Mode
+3 : Minigame active/playable
+4 : Minigame complete, time to hand in
+
+FULL_SPEED_AHEAD effect power:
+0 : Normal Mode
+1 : Easy Mode
+]]--
 
 tpz = tpz or {}
 tpz.full_speed_ahead = tpz.full_speed_ahead or {}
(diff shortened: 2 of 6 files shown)


scripts/globals/quests.lua
```diff
@@ -481,6 +481,8 @@ tpz.quest.id =
         VW_OP_115_VALKURM_DUSTER        = 168,
         VW_OP_118_BUBURIMU_SQUALL       = 169,
         PRELUDE_TO_PUISSANCE            = 170,
+
+        FULL_SPEED_AHEAD                = 179, -- + --
```
ibm2431, line 485: This gap worries me; we'll have to check dats later and add whatever is missing from Jeuno's quest log

scripts/quests/full_speed_ahead.lua
```diff
@@ -13,25 +30,30 @@ tpz.full_speed_ahead.motivation_decay = 2
...
 
-tpz.full_speed_ahead.onEffectLose = function(player)
+tpz.full_speed_ahead.onEffectLose = function(player, effect)
     player:delStatusEffectSilent(tpz.effect.MOUNTED)
     player:countdown(0) 
     player:enableEntities({})
+
+    -- If in Batallia Downs and didn't get the completion flag (failed/dismounted)
+    if player:getZoneID() == tpz.zone.BATALLIA_DOWNS and player:getCharVar("[QUEST]FullSpeedAhead") ~= 4 then
+        player:startEvent(26, 0, effect:getPower())
```
ibm2431, line 50: How are we handling the quest mounted status again? Did it get a special effect, or are we using regular mounted?

In other words: what if a player fails, quits, rides a chocobo in Batalia, then dismounts?

scripts/zones/Upper_Jeuno/npcs/Mapitoto.lua
```diff
@@ -31,15 +32,15 @@ function onTrade(player,npc,trade)
...
-    local hasTrainersWhistle = player:hasKeyItem(tpz.ki.TRAINERS_WHISTLE)
 
-    if hasTrainersWhistle then
+    if fsaQuest == QUEST_COMPLETED then
         player:startEvent(10226)
-    elseif fullSpeedAheadStatus == 2 then -- Retry
-        player:startEvent(10224, 1)
     elseif fullSpeedAheadStatus == 3 then -- Complete
-        player:startEvent(10225, tpz.ki.TRAINERS_WHISTLE)
+        player:startEvent(10225, tpz.ki.TRAINERS_WHISTLE, 15533, ID.npc.MAPITOTO)
```
ibm2431: Could be wrong, but I think you bumped up your quest vars, so talking to Mapitoto after trying once will result in completion. 😉

scripts/globals/quests.lua
TeoTwawki: ~~he dat index stops at 158 and shows this quest as 157 so I dunno wtf we did to get these values...~~
_our packet is fubar and we're gonna wind up with migrations again later, awesome_

ibm2431: More fuel for the dumpster fire that this one PR line has illuminated: every quest area is messed up at each interval of 32!

scripts/quests/full_speed_ahead.lua
zach2good: It has its own special effect which is a wrapper for regular mount and all the functions in `quests/full_speed_ahead.lua`. This handles the loss of the underlying mounted effect, it's own loss, ticks, power for difficulty etc.

I tested your scenario: nothing extra happens :+1:

ibm2431 (approved):
We'll have to fix our quest logs and packets later; I'm not going to dump that on you to fix before this can be merged!

cocosolos (approved):

PR #775, opened 2020-06-26
Trust Spawn/Despawn/Death/Teamwork & Error Messages 

zach2good:
Ended up completing this feature once I tracked down the required packet and the message offsets.

**Done:**
messageCombat and trust messages

Swapping out hardcoded values for spell table entries.

Swap out hard-coded trust ID values for tabled spellID values in these:

Table the system_messages used in trust.lua

Maybe rename getSeekingParty to isSeekingParty

Add timer for last time a player joined a party or when the party was formed, for the last trust error message

**_I affirm:_**


TeoTwawki:
might be worth renaming that packet/message while its still young: 
https://github.com/project-topaz/topaz/search?q=CMessageDebugPacket&unscoped_q=CMessageDebugPacket

scripts/globals/msg.lua
```diff
@@ -311,3 +311,17 @@ tpz.msg.basic =
     TRUST_NO_CAST_TRUST    = 700,  -- You are unable to use Trust magic at this time.
     TRUST_NO_CALL_AE       = 717,  -- You cannot call forth alter egos here.
 }
+
+------------------------------------
+-- Message System
+------------------------------------
+
```
TeoTwawki, line 318: since you renamed the paclket, maybe rename the table to match its new name?

scripts/globals/spells/trust/ajido-marujido.lua
```diff
@@ -27,10 +27,20 @@ function onMobSpawn(mob)
 
     mob:addSimpleGambit(ai.t.TARGET, ai.c.NOT_SC_AVAILABLE, 0,
                         ai.r.MA, ai.s.HIGHEST, tpz.magic.spellFamily.NONE, 30)
+
+    tpz.trust.synergyMessage(mob, {
```
TeoTwawki: I expect using the term "synergy" will be confusing once we have synergy craft skill work started.

scripts/globals/msg.lua
TeoTwawki: OH! Ima dunce these are yellow system3 type aren't they, i mistook this as an enum for the combatMessage packet you did.

src/map/packets/message_system.h
```diff
@@ -26,6 +26,16 @@
...
+enum MSGSYSTEM_ID : uint16
+{
+    MSGSYSTEM_TRUST_NO_SEEKING_PARTY         = 296,   // You cannot use Trust magic while seeking a party.
+    MSGSYSTEM_TRUST_DELAY_NEW_PARTY_MEMBER   = 297,   // While inviting a party member, you must wait a while before using Trust magic.
+    MSGSYSTEM_TRUST_MAXIMUM_NUMBER           = 298,   // You have called forth your maximum number of alter egos.
+    MSGSYSTEM_MSGSYSTEM_TRUST_ALREADY_CALLED = 299,   // That alter ego has already been called forth.
+    MSGSYSTEM_TRUST_NO_ENMITY                = 300,   // You cannot use Trust magic while having gained enmity.
+    MSGSYSTEM_TRUST_SOLO_OR_LEADER           = 301,   // You cannot use Trust magic unless you are solo or the party leader.
+};
+
```
Omnione: Does this need to be here rather than msg.lua?, im only asking as we usually put these messages in msg.lua and call from that, and it seems that you allready call tpz.msg for messages.

Omnione (commented):
Other than the messages, this looks really good.

src/map/packets/message_system.h
zach2good: It doesn't _need_ to be here, but this is following the pattern from basic messages since everything has to flow through core to be able to have the right packet sent out

Omnione: Ah ok, got ya, ok carry on lol

Omnione: and wow totaly missed that MSGSYSTEM_MSGSYSTEM lol, glad you caught that

scripts/globals/trust.lua
```diff
@@ -86,3 +115,43 @@ tpz.trust.spawn = function(caster, spell)
...
     return 0
 end
+
+tpz.trust.message = function(mob, id)
+    local master = mob:getMaster()
+    local offset = (mob:getTrustID() - 896) * 100
+    local party = master:getParty()
+
+    for _, member in ipairs(party) do
+        member:messageCombat(mob, offset + id, 0, 711)
```
ibm2431: I worry about starting message ID shifting. I don't know the last time it's happened on the low end, but I don't know if this trust global will be on the minds of people running text ID tools should it happen.

Would it be possible to define a very-first-starting-trust message ID in `msg`, and then calculate offsets based on that?
`local message_id = tpz.msg.combat.GLOBAL_TRUST_OFFSET + ((trust_id - x) * 100)`
The global would need to be adjusted if the _size_ of trust message blocks change, but they'd be a little more resistant if SE decides to throw a couple random messages before the trust block.

Variable naming could be a bit clearer - what's being passed in isn't a message ID, but a message offset. Complication in that you're also calculating a _trust_ offset. Maybe `trust_offset` + `message_offset`.

scripts/globals/trust.lua
```diff
@@ -86,3 +115,43 @@ tpz.trust.spawn = function(caster, spell)
...
+            for id, message in pairs(teamwork_messages) do
+                if member:getTrustID() == id then
+                    table.insert(messages, message)
+                end
+            end
+        end
+    end
+
+    if table.getn(messages) > 0 then
+        tpz.trust.message(mob, messages[math.random(#messages)])
```
ibm2431, line 145: Are we certain the message selected is random? There's no priority system? I've done a lot of partying with ROV versions of The Princesses, and I remember them _always_ commenting about Iroha.

I don't know how far your retail character is, so I'll tag @project-topaz/researcher :
1. Summon Iroha II
2. Summon The Princesses II
3. See if The Princesses ever comment about each other versus Iroha

scripts/globals/trust.lua
```diff
@@ -86,3 +115,43 @@ tpz.trust.spawn = function(caster, spell)
...
+        -- Defaults to regular spawn message
+        tpz.trust.message(mob, tpz.trust.message_offset.SPAWN)
+    end
+end
+
+-- For debugging and lining up teamwork messages
+tpz.trust.dumpMessages = function(mob)
+    for i=0, 20 do
+        tpz.trust.message(mob, i)
+    end
```
ibm2431, line 156: 👍

src/map/lua/lua_baseentity.cpp
```diff
@@ -533,6 +534,39 @@ inline int32 CLuaBaseEntity::messageSystem(lua_State* L)
...
+*  Notes   :
+************************************************************************/
+int32 CLuaBaseEntity::messageCombat(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+
+    CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
+
+    CBaseEntity* PTarget;
```
ibm2431: "Target" or "Speaker"?

src/map/packets/message_system.h
```diff
@@ -26,6 +26,16 @@
...
 
+enum MSGSYSTEM_ID : uint16
+{
+    MSGSYSTEM_TRUST_NO_SEEKING_PARTY         = 296,   // You cannot use Trust magic while seeking a party.
+    MSGSYSTEM_TRUST_DELAY_NEW_PARTY_MEMBER   = 297,   // While inviting a party member, you must wait a while before using Trust magic.
+    MSGSYSTEM_TRUST_MAXIMUM_NUMBER           = 298,   // You have called forth your maximum number of alter egos.
+    MSGSYSTEM_TRUST_ALREADY_CALLED           = 299,   // That alter ego has already been called forth.
+    MSGSYSTEM_TRUST_NO_ENMITY                = 300,   // You cannot use Trust magic while having gained enmity.
+    MSGSYSTEM_TRUST_SOLO_OR_LEADER           = 301,   // You cannot use Trust magic unless you are solo or the party leader.
+};
```
ibm2431: You're not using these enums in core~!

scripts/globals/trust.lua
```diff
@@ -28,14 +38,27 @@ tpz.trust.canCast = function(caster, spell, not_allowed_trust_ids)
...
     end
 
-    -- TODO: Block summoning trusts if someone recently joined party
+    -- Block summoning trusts if seeking a party
+    if caster:isSeekingParty() then
+        caster:messageSystem(tpz.msg.system.TRUST_NO_SEEKING_PARTY)
+        return -1
+    end
 
+    -- Block summoning trusts if someone recently joined party (120s)
```
ibm2431, line 51: I don't remember, but are we already blocking party invites to players if trusts are present? (Either a party leader using trusts sending an invite to a free person, or a player-only party sending an invite to someone with trusts out)

Not a requirement for this PR if we're not; just making sure we don't forget any loose threads~

ibm2431 (commented):
Thanks, Zach!

scripts/globals/trust.lua
TeoTwawki: if its a per zone thing (as in, its in the string dats of each zone, even if they all have the same ID now) it should really be in the IDs.lua anyway just for this reason. if its not in the zones dat, we generally won't need to worry about it shifting. (I'm still in favor of offsets where we can, vs enuming all the stuff and things, just FYI))

I most zones we have several IDs that are generally the same in every zone - we still IDs.lua them because they will eventually get shifted,
image

scripts/globals/trust.lua
zach2good: _Seems_ random...

src/map/packets/message_system.h
zach2good: I've decided to remove them (for now), because I don't want to chase down all of the other hard coded uses of messageSystem...

scripts/globals/trust.lua
zach2good: Not really, on the core end everything is pretty much unlimited and the only blocking happens on the lua side. The way trusts are handled in and around a player/mixed party is all sorts of wrong. That'll have to be a whole research and dev leg on its own before release :(

scripts/globals/trust.lua
TeoTwawki: ~~I believe there is a priority system of sorts based on what order they were summoned in on retail, but someone will need to confirm that~~ _Vidya says I'm wrong_

ibm2431 (approved):
Looks good to me! Thanks, Zach!

Issue #702, opened 2020-06-10
Zoning into BCNM zones crashing server


AvNytemare:
**_I have:_**

- [X] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [X] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Only addon loaded: Ashita Recast

There is about an 80% chance to crash the server upon zoning into an area containing a burning circle. It is not limited to just one specific BCNM area, it happens with all of them. There also seems to be a chance that if you successfully make it into the zone where you can enter the burning circle, that upon entering a BCNM fight it will also cause a crash, resulting in the loss of the fight and leaving you stuck in the zone, the only way out is to warp or get GM assistance if no warp is available.

TeoTwawki:
if this is a windows based server, I suggest you attach the visual studio debugger. when it crashes again you can have visual studio save the crashdump to a folder, copy your executable and .pdb (program debug build file) and pack those with something like 7zip, throw it on drop box or googledrive or something so someone can download and review the crash. You might also be able to spot the problem yourself looking at the call stack with some practice.

TeoTwawki:
if the server is linux based I've never done a crashdump there so we'd need to a ~~fireman~~ ~~police officer~~ ~~scientist~~ _linux programmer_ for help

kaincenteno:
Still happening with latest Canary from last week.

kaincenteno:
here you go @TeoTwawki  (but this doesnt come from VS)
```
#0  0x00007f893eb6cd2d in __GI___pthread_timedjoin_ex (threadid=140227398309632, thread_return=0x0, abstime=0x0, block=<optimized out>)
    at pthread_join_common.c:89
#1  0x00007f893e058933 in std::thread::join() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x000055c69a4521e5 in do_final (code=0) at src/map/map.cpp:280
#3  0x000055c69a39c45c in main (argc=1, argv=0x7ffd6374a058) at src/common/kernel.cpp:272
```

^thats the backtrace and the last message before the hange:

```
double free or corruption (out)
[26/Jun] [01:28:46][Debug] Message: Received message 0 from message server
```

TeoTwawki:
hmm.
```cpp

void do_final(int code)
{
    delete[] g_PBuff;
    delete[] PTempBuff;

[...]

PR #778, opened 2020-06-27
Fix double-free errors in exit/crash

Kreidos:
do_final was written in such a way that it was only ever intended to be called once. However, since the signal_handler calls do_abort and do_final in response to various error states, this means do_final could be (and often is) called multiple times if that signal/error occurs during (or as a result of) the shutdown process itself. This is a problem since there weren't any safeguards against memory attempting to be de-allocated a second time.

This PR ensures that called shutdown procedures are safe (or at least, resistant) against double-freeing. It should also reduce the chance do_final gets caught in a recursive error loop.

This will hopefully lend insight into the real sources of some difficult-to-reproduce crashes, since the recursive double-frees after an error were often trampling over the stack and obfuscating the original cause in those cases.

**_I affirm:_**

### DIFF
--- a/src/common/taskmgr.cpp
+++ b/src/common/taskmgr.cpp
@@ -24,15 +24,24 @@
 #include "../common/timer.h"
 #include "../common/taskmgr.h"
 
-CTaskMgr* CTaskMgr::_instance = NULL;
+CTaskMgr* CTaskMgr::_instance = nullptr;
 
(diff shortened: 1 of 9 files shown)


Kreidos:
Pinging issue #702 as an impetus for this; whose root cause may hopefully be more easily determined with a cleaner error log. :)

ibm2431 (approved):
Looks legit!

PR #779, opened 2020-06-28
Implement crafting skill limitation and various other fixes

Xaver-DaRed:
Among other things, this implements the limit imposed by retail in wich you can only have a certain amount of levels over level 70 in crafts.

**Main changes**
- *Implements "Craft Sepecialization" System*
   Retail only allows up to 40 levels across all main craft skills combined starting from level 70.
   This makes it imposible to have all skills maxed, forcing players to specialize on 1.

- *New configurable option:* Maximum amount of specialization points.
   This allows server to configure how many crafts can a character maximize.
   By default, 400 points are needed per craft.

- *New configurable option:* Maximum level of crafts before using the Specialization System.
   This is shared across all 8 crafts involved. It cannot be individualy set per craft.
   Retail = 700; Level 75 era = 600; Opt out of the system entirely = 1100;

**Other bugfixes**

- *Limit skill up gains amounts (not the chance of it happening) to 0.1 per level above level 70*
   As per https://www.bluegartr.com/threads/57123-Before-you-ask-a-stupid-crafting-question-read-this!
   This needs confirmation from a more recent era. It may have changed after the raise in level cap.
   For now I made this cap the same as the skill common cap, before the specialization system activates, becouse it made
   sense to me. (Level 70)

- *Fixes a badly placed random call*
   The place it was coded in undesirebly lowered the actual chance of 0.2+ skill ups from happening.

- *Removes crafting myths*
   As per discussion in discord.

- *Fixed skill up chance configuration variable*
   It actually works as a multiplier now and skill ups happen when set at 1.0

- *Lower skill up chance of desynths*

- *Cleaned up coments*

**Related Issues**
#66 #401 

**_I affirm:_**

### DIFF
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -450,8 +450,8 @@ int32 doSynthSkillUp(CCharEntity* PChar)
             {
                 int32  skillAmount = 1;
 
-                if (charSkill < 600) // no skill ups over 0.1 happen over level 60 https://www.bluegartr.com/threads/57123-Before-you-ask-a-stupid-crafting-question-read-this!
-                {                    // we need more recent confirmation since the "free" level cap was raised to 70
+                if (charSkill < map_config.craft_common_cap) // no skill ups over 0.1 happen over level 60 https://www.bluegartr.com/threads/57123-Before-you-ask-a-stupid-crafting-question-read-this!
+                {                                            // we need more recent confirmation since the "free" level cap was raised to 70
                     int32  satier = 0;
                     double chance = 0;
 
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -165,11 +165,6 @@ bool isRightRecipe(CCharEntity* PChar)
 
 double getSynthDifficulty(CCharEntity* PChar, uint8 skillID)
 {
-    uint8  ElementDirection = 0;
-    uint8  WeekDay = (uint8)CVanaTime::getInstance()->getWeekday();
-    uint8  crystalElement = PChar->CraftContainer->getType();
-    uint8  direction = (PChar->loc.p.rotation - 16)/32;
-    uint8  strongElement[8] = {2,3,5,4,0,1,7,6};
     Mod ModID = Mod::NONE;
 
     switch (skillID)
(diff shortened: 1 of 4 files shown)


Kreidos:
Nice, thanks Xaver! Would definitely like that config flag for something like this; I'm sure it would be a somewhat 'divisive feature' with some servers. :laughing: 

You may also want to mark this PR as draft, then nobody may merge it until you mark it as ready!

Xaver-DaRed:
I will
It does work, from my testing. that is... 1 hour -2

TeoTwawki:
Similarly some people are very insistent about their myths not being myths in spite of SE confirmation, which is why there were settings for those (altho I can’t speak to if they even still work)

Xaver-DaRed:
the way it was coded, one of those myths didnt even work.
I did ask in topaz discord before removing them

Kreidos:
Could just pull a page from Square's book... If they were removed would anyone even know? :wink: 

I do believe there was some discussion on the discord about that; if someone were really serious about it they could revert these changes. I haven't heard from Consul but I expect the number of servers who would even want/use the myths is very small.

Apocarypse:
If SE states they're myth and our goal is to be as retail accurate as possible, I don't see any other outcome. If we ever get a custom code section up and running, the myths can go in that pile and be configurable.

TeoTwawki:
@xaver-dared I’m surprised any of them still work.

@Apocarypse I’m the one that settingified them, used to be just hard coded on. While I disagree with you logic as no setting is a retail accurate thing, I did want these gone since the outset. But ppl didn’t like that so they became setting instead.

I wasn’t making a “no don’t do that” post. It was a history lesson to laugh about. Coz people still believe this stuff is more than random numbers. Fake news lord Odin wants my craft to HQ!

ghost:
Hello!

What Apocarypse said. It's clearly a customization thing, not an era setting. If people got accomodated to a bug for years, we wouldn't make a setting for it in the fixing PR. Let's take the recent Treasure Hunter values revealed by S-E for example, if people got used to the old ones and we fix them, we won't make a setting out of the old ones. Same goes here, this is not like Neo Dynamis or the old fishing system, etc.

Edit:

I hope I didn't misunderstood (wouldn't be the first time tho...), I'm sorry if it felt like I was reacting to anyone's message, I wasn't. Just my thoughts about a setting!

Thanks!

TeoTwawki:
I would say most of our settings aren’t era settings and are custom i would also say thats irrelevant seeing how busted up the crafting code has been for ages while trying to accommodate it

TeoTwawki:
@UynGH if its not clear i’m saying axe the myths and code related but don’t mistake “its not era” or “not retail” as a reason because then your logic applies to almost everything else too. Kill it coz its both broken and silly instead!

src/map/utils/synthutils.cpp
```diff
@@ -572,23 +499,132 @@ int32 doSynthSkillUp(CCharEntity* PChar)
...
+                if (charSkill > 700) // The cap from which you are limited to 400 points
                 {
-                    PChar->WorkingSkills.skill[skillID] += 0x20;
+                    uint16 skillCumulation = 0;
+                    uint8 skillHighest    = 0;
+
+                    // we are gonna set which is the OTHER highest skill over lvl 70 and the amount of points we have of the 400 to allocate
+                    if (PChar->RealSkills.skill[49] > 700) // Craft 1
+                    {
+                        skillCumulation += (PChar->RealSkills.skill[49] - 700);
```
Kreidos: Perhaps all these 8 skill checks could be replaced with one fairly good for loop!
```c++
for(int i = SKILL_WOODWORKING; i <= SKILL_COOKING; i++)
{
    if (PChar->RealSkills.skill[i] > 700)
    {
        skillCumulation += (PChar->RealSkills.skill[i] - 700);
        if (skillID != i && PChar->RealSkills.skill[i] > skillHighest)
            skillHighest = i;
    }
}
```

Kreidos: I'm also unsure if it's intentional, but wouldn't `skillCumulation` be reduced for every skill below 70? (ie 500 - 700 would be -200 to skillCumulation).

Xaver-DaRed: Thank you mate. Done

Noikar:
Sorry for the late response, Dawnbreak doesn't use the ridiculous crafting myths. We will be pulling this later in the week to test it.

src/map/utils/synthutils.cpp
```diff
@@ -502,64 +424,71 @@ int32 doSynthSkillUp(CCharEntity* PChar)
...
-        {
+        {                                                                                                     // if you break a recipe higher than 5 levels, no skill ups
             continue;
         }
 
-        if (charSkill < maxSkill)
+        if (charSkill < maxSkill) // Check if a character can skill up
         {
-            double skillUpChance = ((double)baseDiff*(map_config.craft_chance_multiplier - (log(1.2 + charSkill/100))))/10;
+            double skillUpChance = (double)baseDiff * map_config.craft_chance_multiplier * (3 - (log(1.2 + charSkill/100))) / 10;
```
ibm2431, line 433: Make sure I don't forget to compile my retail data~!

This is fine for now as it appears to be fixing the formula so that a setting of 1.0 is adjusted back to what the default used to be. But I'm not entirely convinced the original formula's _approach_ matches retail.

src/map/utils/synthutils.cpp
```diff
@@ -572,23 +501,59 @@ int32 doSynthSkillUp(CCharEntity* PChar)
...
+                    PChar->WorkingSkills.skill[skillHighest] -= 0x20;
+                    PChar->pushPacket(new CCharSkillsPacket(PChar));
                 }
 
                 charutils::SaveCharSkills(PChar, skillID);
+
+                if (skillHighest != 0 && skillCumulation > map_config.craft_specialization_points)
+                {
+                    charutils::SaveCharSkills(PChar, skillHighest);
+                }
```
ibm2431: Is restructuring possible so that all your `skillHighest != 0 && skillCumulation > map_config.craft_specialization_points` logic is contained in a single branch?

ibm2431 (commented):
Thanks, Xaver! I approve of fewer references to elements!

src/map/utils/synthutils.cpp
Xaver-DaRed: Sure thing

src/map/utils/synthutils.cpp
Xaver-DaRed: On a future PR I was planning on updating this(skill up rate) and the skill up range with retails current system. As of now, this is.... "good enough"

Xaver-DaRed:
If this gets merged to canary, server owners should be noticed that skill up rate config now works correctly at 1.0 and that they should check it

src/map/utils/synthutils.cpp
ibm2431: Agreed - the fewer things fixed per PR the better! 😄

ibm2431 (approved):

cocosolos (approved):

ibm2431:
Just linking this on #831 so it's not forgotten~

PR #786, opened 2020-06-30
tpz.zone referenced in casket_loot

ShelbyZ:
casket_loot.lua should require zone global with its usage of tpz.zone.  I only recently saw this error crop up on a mutli-server setup.  Something must behave differently when running on a single server that the log does not spam errors when loading a zone eligible for caskets.

**_I affirm:_**

### DIFF
--- a/scripts/globals/casket_loot.lua
+++ b/scripts/globals/casket_loot.lua
@@ -4,6 +4,9 @@
 -- this is because some zones have high mobs and low mobs,
 -- and the drops are level dependant.
 -----------------------------------------------------------
+
+require("scripts/globals/zone")
+
 tpz = tpz or {}
 tpz.casket_loot = tpz.casket_loot or {}
 


TeoTwawki:
~~is there a specific error message being addressed?~~ due to shenanigans, the zone global is always loaded at all times in all zones - every zone lua and id's lua make sure the table exists and the require is loaded

I phrased that stupidly: whats the error message, is what I mean to ask

ShelbyZ:
image

This may be related to the split for my setup (298 zones, 1 zone).  If zone global should be available for all lua there is a race condition that code is run before the zone table is available via lua parser

cocosolos:
> the zone global is always loaded at all times in all zones - every zone lua and id's lua make sure the table exists and the require is loaded

That is the `zones` global table, which is different than the `zone.lua` global file. I think the reason that this is happening is that when the zones are loaded, NPCs are loaded, then Mobs, then the actual Zone.lua is checked for init. The zone.lua global is included in the individual Zone.lua files, ~~and because of the order zones are loaded then it's already in memory when it gets to a zone with caskets~~ I have no idea _what_ is loading the zone.lua global, since it appears all npc and mobs and mixins are loaded before any Zone.lua file is accessed. On a split server where the first zone has a the casket zone mixin, that gets loaded before the Zone.lua is loaded which includes the zone.lua global include.

TeoTwawki:
@cocosolos yeah checked it out after getting the error message and its because the caskets lua needed the zone **name** not its table contents, so the require is the simplest fix (it still finds what mixin to look for because it knows where the player is, but theres a set that comes up nil without the require)

TeoTwawki (approved):

59blargedy (approved):

cocosolos (approved):

PR #790, opened 2020-07-01
Infill nm_spawn_points for all existing lottery mobs

### LINKED ISSUE #262: calling `UpdateNMSpawnPoint()` on NMs with no alternate spawn points
_Monday Feb 11, 2019 at 20:40 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 5729_

----

**_I have:_**

n/a

I AM THE **MASTER** /kungfumusic

"UpdateNMSpawnPoint: SQL error: No entries for mobid \<number here\> found."

~~17219886, Rampaging Ram~~
~~17211702, Maighdean Uaine~~
~~17220001, Bendigeit Vran~~
~~17309954, Despot~~
_thanks Wren._

~~All trigger that log error.~~ I'm sure there are more as well. These actually _should_  have multiple spawn pos, but  I also feel we could be handling this better than just assuming every NM has multiple pos entries in that sql table / auto calling this function in the global on every NM. These log messages aren't making anyone go capture retail pos and we can see which NM need pos data using a very simple SQL query.

wrenffxi:
**_I affirm:_**

Fixes #262 

This PR contains two commits.

* The first orders nm_spawn_points.sql by mobid (fixing just a few entries that aren't in order)
* The second infills from mob_spawn_points missing coordinates for all lottery-type mobs.  The lottery code calls UpdateNMSpawnPoint, which causes those logged errors in #262, so these entries will prevent errors for all existing mobs.

I understand that we may eventually move to a different/better spawn point system, but this at least fixes the immediate problem.

### DIFF
--- a/sql/nm_spawn_points.sql
+++ b/sql/nm_spawn_points.sql
@@ -138,10 +138,6 @@ INSERT INTO `nm_spawn_points` VALUES (16785676,46,86.146,-6.743,-526.943);
 INSERT INTO `nm_spawn_points` VALUES (16785676,47,87.834,-7.597,-520.502);
 INSERT INTO `nm_spawn_points` VALUES (16785676,48,76.962,-5.935,-514.680);
 INSERT INTO `nm_spawn_points` VALUES (16785676,49,102.059,-5.104,-559.392);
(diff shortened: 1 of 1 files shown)


ibm2431 (approved):
Looks legit, thanks Wren~!

cocosolos (approved):

PR #791, opened 2020-07-01
Trust Melee Balancing

zach2good:
**_I affirm:_**

Trusts were using monster-level weapon damage with no regard to their jobs. This is a quick patch to bring their weapon damage down and scale a bit around their jobs' default weapon. There are mods to adjust Excenmille and Naji's weapon damages since they don't use their jobs' default weapons.

NOTE: Currently only targets Trusts that are "available". Will be expanded in future PRs.


Wiggo32 (approved):

cocosolos (approved):

ibm2431 (approved):
I hate everything about mobs and having "weapons".

But none of that your fault, zach. You've done the best with what you've got~

PR #794, opened 2020-07-02
Fix incorrect id in item_latents.sql; sort

wrenffxi:
Imperial Bow: fix incorrect id, then sort by id

**_I affirm:_**

* fix incorrect item itemId for Imperial Bow
* sort the entries by itemId


ibm2431 (approved):
Thanks, Wren!

59blargedy (approved):

PR #796, opened 2020-07-02
I Can Hear a Rainbow: scripts/globals to scripts/quests

wrenffxi:
**_I affirm:_**

* remove from scripts/globals, and rewrite in scripts/quests
* make the color data more easily readable
* localVar the variable used in eventUpdate, which removes the need for onEventFinish
* clear the correct charVar on quest completion
* rainbow controlled by localVar on rainbow NPC, which removes need for server variable

update: per feedback,

* ignore weather set by SCH<sup>1</sup>
* use set constructor for color data
* swap order of quest check vs ruby check

<sup>1</sup>Two issues had to be solved here.  First, there wasn't a way to get weather from zone into LUA.  To resolve this, I added a zone:getWeather() binding.  Second, during onZoneIn, the players's zone is not yet set, causing player:getZone() to return null.  To resolve this, I added an optional bool isZoning parameter to player:getZone(), which will instead return the player's destination zone.


TeoTwawki:
thankyouwrenthatscriptnamealwaysbotheredme

zach2good (approved):
I was looking at this the other day hoping someone would clean it up, and here it is! Nice work!

cocosolos (approved):

scripts/quests/i_can_hear_a_rainbow.lua
```diff
@@ -0,0 +1,158 @@
...
+        {
+            [tpz.zone.VALKURM_DUNES] = true,
+            [tpz.zone.ROLANBERRY_FIELDS] = true,
+            [tpz.zone.CAPE_TERIGGAN] = true,
+            [tpz.zone.EASTERN_ALTEPA_DESERT] = true,
+            [tpz.zone.MERIPHATAUD_MOUNTAINS] = true,
+            [tpz.zone.YUHTUNGA_JUNGLE] = true,
+            [tpz.zone.YHOATOR_JUNGLE] = true,
+            [tpz.zone.WESTERN_ALTEPA_DESERT] = true,
+        },
```
ibm2431: We have a [set constructor](https://github.com/project-topaz/topaz/blob/release/scripts/globals/common.lua#L35-L45) which will do this for you from a simple table:
```lua
zones = set{
    tpz.zone.VALKURM_DUNES,
    tpz.zone.ROLANBERRY_FIELDS,
    tpz.zone.CAPE_TERIGGAN,
   -- etc
}

File: scripts/quests/i_can_hear_a_rainbow.lua [hunk 2]
```diff
@@ -0,0 +1,158 @@
...
+    },
+}
+
+-------------------------------------------------
+-- public functions
+-------------------------------------------------
+
+quests.i_can_hear_a_rainbow.onZoneIn = function(player)
+    if player:hasItem(1125, 0) and player:getQuestStatus(WINDURST, tpz.quest.id.windurst.I_CAN_HEAR_A_RAINBOW) == QUEST_ACCEPTED then
```
ibm2431: Maybe swap order of quest status check and `hasItem`. HasItem iterates over everything in the bag, whereas quest status is a quick return. Checking for status first means we only ping inventory if player is on the quest.

scripts/quests/i_can_hear_a_rainbow.lua
```diff
@@ -0,0 +1,158 @@
...
+        local trigger = false
+
+        -- get data for basic weather (e.g. downgrade GALES to WIND)
+        local weather = player:getWeather()
```
ibm2431: `entity:getWeather()` will return SCH weather too; are those lost on zoning? Does the effect loss include logging?

TeoTwawki: use zone entities weather instead. sch 'weather' doesn't work for smn quest

wrenffxi: I don't currently see a way to get zone's weather into LUA, so will have to either add a getWeather binding to lua_zone, and/or add an optional ignoreScholar bool parameter to player:getWeather.  Any preference?

TeoTwawki: when the object type isn't a player or mob, it assumed we want zone weather:
```lua
inline int32 CLuaBaseEntity::getWeather(lua_State *L)
{

    WEATHER weather = WEATHER_NONE;
    if (m_PBaseEntity->objtype & TYPE_PC || m_PBaseEntity->objtype & TYPE_MOB)
        weather = battleutils::GetWeather((CBattleEntity*)m_PBaseEntity, false);
    else
        weather = zoneutils::GetZone(m_PBaseEntity->getZone())->GetWeather();

    lua_pushinteger(L, weather);

    return 1;
}
```
so you can get the zone (from the player if need be) and then
`zone:getWeather()`
works

TeoTwawki: `player:getZone():getWeather()`

for example

TeoTwawki: Sorry I see what you are saying about it not being in lua_zone...I thought we already had this functionality in use elsewhere already.

Quick search says I am wrong

wrenffxi:
Updated PR per feedback.  Description lists changes.

ibm2431 (approved):
Thanks, Wren~

PR #803, opened 2020-07-03
Guild shop always open

kaincenteno:
**_I affirm:_**
- [x] that I've _tested my code_ since the last commit in the PR, and will test after any later commits (See comments below)

Changes guildshops to be less hacky and more retail accurate by sending the correct information

This is a teamwork with @Kreidos , he has worked on all c and cpp code. :cat:  Thank you @Kreidos 

updated guild vendors to use npc_util (I haven't verified that it doesn't break anything but would appreciate some feedback from a Tester as that's the only part that HASN'T been tested):
Cletae
Celestina

### DIFF
--- a/scripts/zones/Mhaura/npcs/Celestina.lua
+++ b/scripts/zones/Mhaura/npcs/Celestina.lua
@@ -6,57 +6,55 @@
 -- Guild Merchant NPC: Goldsmithing Guild
 -- !pos -37.624 -16.050 75.681 249
 -----------------------------------
-local ID = require("scripts/zones/Mhaura/IDs");
-require("scripts/globals/settings");
-require("scripts/globals/shop");
-require("scripts/globals/keyitems");
-require("scripts/globals/quests");
+local ID = require("scripts/zones/Mhaura/IDs")
+require("scripts/globals/settings")
+require("scripts/globals/shop")
+require("scripts/globals/keyitems")
+require("scripts/globals/quests")
+require("scripts/globals/status")
+require("scripts/globals/npc_util")
 -----------------------------------
 
 function onTrade(player,npc,trade)
-
-    if (player:getQuestStatus(OTHER_AREAS_LOG,tpz.quest.id.otherAreas.THE_SAND_CHARM) == QUEST_ACCEPTED) then
-        if (trade:hasItemQty(13095,1) and trade:getItemCount() == 1) then
-            player:startEvent(127); -- Finish quest "The Sand Charm"
+    if player:getQuestStatus(OTHER_AREAS_LOG,tpz.quest.id.otherAreas.THE_SAND_CHARM) == QUEST_ACCEPTED then
+        if npcUtil.tradeHasExactly(trade, 13095) then
+            player:startEvent(127) -- Finish quest "The Sand Charm"
         end
     end
 
-    if (player:getQuestStatus(JEUNO,tpz.quest.id.jeuno.RIDING_ON_THE_CLOUDS) == QUEST_ACCEPTED and player:getCharVar("ridingOnTheClouds_3") == 5) then
-        if (trade:hasItemQty(1127,1) and trade:getItemCount() == 1) then -- Trade Kindred seal
-            player:setCharVar("ridingOnTheClouds_3",0);
-            player:tradeComplete();
-            player:addKeyItem(tpz.ki.SOMBER_STONE);
-            player:messageSpecial(ID.text.KEYITEM_OBTAINED,tpz.ki.SOMBER_STONE);
+    if player:getQuestStatus(JEUNO, tpz.quest.id.jeuno.RIDING_ON_THE_CLOUDS) == QUEST_ACCEPTED and player:getCharVar("ridingOnTheClouds_3") == 5 then
+        if npcUtil.tradeHasExactly(trade, 1127) then -- Trade Kindred seal
+            player:setCharVar("ridingOnTheClouds_3",0)
+            player:confirmTrade()
+            npcUtil.giveKeyItem(player, tpz.ki.SOMBER_STONE)
         end
     end
+end
 
-end;
-
-function onTrigger(player,npc)
-
-    if (player:getCharVar("theSandCharmVar") == 3) then
-        player:startEvent(126,13095); -- During quest "The Sand Charm" - 3rd dialog
-    elseif (player:sendGuild(528,8,23,4)) then
-        player:showText(npc,ID.text.GOLDSMITHING_GUILD);
+function onTrigger(player, npc)
+    if player:getCharVar("theSandCharmVar") == 3 then
+        player:startEvent(126, 13095) -- During quest "The Sand Charm" - 3rd dialog
+    else
+        local guildSkillId = tpz.skill.GOLDSMITHING
+        local stock = tpz.shop.generalGuildStock[guildSkillId]
+        tpz.shop.generalGuild(player, stock, guildSkillId)
     end
+end
 
+function onEventUpdate(player, csid, option)
 end;
 
-function onEventUpdate(player,csid,option)
-end;
-
-function onEventFinish(player,csid,option)
-
-    if (csid == 126 and option == 70) then
-        player:setCharVar("theSandCharmVar",4);
+function onEventFinish(player, csid, option)
+    if csid == 126 and option == 70 then
+        player:setCharVar("theSandCharmVar", 4)
     elseif (csid == 127) then
-        player:tradeComplete();
-        player:setCharVar("theSandCharmVar",0);
-        player:setCharVar("SmallDialogByBlandine",1);
-        player:addKeyItem(tpz.ki.MAP_OF_BOSTAUNIEUX_OUBLIETTE);
-        player:messageSpecial(ID.text.KEYITEM_OBTAINED,tpz.ki.MAP_OF_BOSTAUNIEUX_OUBLIETTE);
-        player:addFame(MHAURA,30);
-        player:completeQuest(OTHER_AREAS_LOG,tpz.quest.id.otherAreas.THE_SAND_CHARM);
+        player:confirmTrade()
+        npcUtil.completeQuest(player, OTHER_AREAS_LOG, tpz.quest.id.otherAreas.THE_SAND_CHARM, {
+            keyItem = tpz.ki.MAP_OF_BOSTAUNIEUX_OUBLIETTE,
+            fame_area = MHAURA,
+            fame = 30,
+            var = theSandCharmVar
+        })
+        player:setCharVar("SmallDialogByBlandine", 1)
     end
-
-end;
+end
--- a/scripts/zones/Mhaura/npcs/Celestina.lua
+++ b/scripts/zones/Mhaura/npcs/Celestina.lua
@@ -52,8 +52,7 @@ function onEventFinish(player, csid, option)
         npcUtil.completeQuest(player, OTHER_AREAS_LOG, tpz.quest.id.otherAreas.THE_SAND_CHARM, {
             keyItem = tpz.ki.MAP_OF_BOSTAUNIEUX_OUBLIETTE,
             fame_area = MHAURA,
-            fame = 30,
-            var = theSandCharmVar
+            var = "theSandCharmVar"
         })
         player:setCharVar("SmallDialogByBlandine", 1)
     end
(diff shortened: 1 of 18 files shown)


scripts/zones/Mhaura/npcs/Celestina.lua
```diff
@@ -6,57 +6,55 @@
...
-        player:setCharVar("SmallDialogByBlandine",1);
-        player:addKeyItem(tpz.ki.MAP_OF_BOSTAUNIEUX_OUBLIETTE);
-        player:messageSpecial(ID.text.KEYITEM_OBTAINED,tpz.ki.MAP_OF_BOSTAUNIEUX_OUBLIETTE);
-        player:addFame(MHAURA,30);
-        player:completeQuest(OTHER_AREAS_LOG,tpz.quest.id.otherAreas.THE_SAND_CHARM);
+        player:confirmTrade()
+        npcUtil.completeQuest(player, OTHER_AREAS_LOG, tpz.quest.id.otherAreas.THE_SAND_CHARM, {
+            keyItem = tpz.ki.MAP_OF_BOSTAUNIEUX_OUBLIETTE,
+            fame_area = MHAURA,
+            fame = 30,
```
wrenffxi: fame defaults to 30, so this parameter isn't needed

scripts/zones/Mhaura/npcs/Celestina.lua
```diff
@@ -6,57 +6,55 @@
...
+            var = theSandCharmVar
```
wrenffxi: var = "theSandCharmVar"

scripts/zones/Southern_San_dOria/npcs/Cletae.lua
```diff
@@ -6,30 +6,28 @@
...
+require("scripts/globals/npc_util")
 -----------------------------------
 
-function onTrade(player,npc,trade)
-    if (FlyerForRegine == 1) then
-        count = trade:getItemCount()
-        MagicFlyer = trade:hasItemQty(532,1)
-        if (MagicFlyer == true and count == 1) then
+function onTrade(player, npc, trade)
+    if FlyerForRegine == 1 then
```
wrenffxi, line 18: Undefined variable: FlyerForRegine (this issue existed in the code before you edited it)

The Flyers for Regine quest got a major update on retail at some point -- see recent issue and my related PRs, that implement more retail-accurate behavior.  Now, you don't need to randomly trade flyers to everyone.  Instead, the 15 NPCs who will accept flyers give you a message when you approach them.  I'm guessing that because of this on retail, NPCs who -don't- accept flyers may no longer give refusal messages, so this code block may not be needed any longer.  However, retail confirmation is needed.

scripts/zones/Mhaura/npcs/Celestina.lua
kaincenteno: fixed thank you @wrenffxi  :cat:

scripts/zones/Mhaura/npcs/Celestina.lua
kaincenteno: fixed thank you @wrenffxi  :cat:

scripts/zones/Southern_San_dOria/npcs/Cletae.lua
kaincenteno: that's a valid point, i'm going to ask one of the testers to check out if this behavior is still needed, if so a new PR will be posted to focus solely on it. :cat:  thanks for bringing this up @wrenffxi

kaincenteno:
Can i have a review ready tag for this quest since there hasnt been any feedback about the Flyers for Regine, and nothing on it was changed for it besides using npcutils.

thank you =)

scripts/zones/Southern_San_dOria/npcs/Cletae.lua
TeoTwawki: this will still throw and error in the meantime. may as well just comment out or remove if you aren't going to define the variable you are using till a new pr.

kaincenteno: @TeoTwawki  added the var and a also a note.

src/map/lua/lua_baseentity.cpp
```diff
@@ -3792,16 +3794,28 @@ inline int32 CLuaBaseEntity::addShopItem(lua_State *L)
...
     ((CCharEntity*)m_PBaseEntity)->Container->setItem(slotID, itemID, 0, price);
 
+    if(lua_isnumber(L, 3) && lua_isnumber(L, 4))
+    {
+        uint8 guildID = (uint8)lua_tonumber(L, 3);
+        uint16 guildRank = (uint16)lua_tonumber(L, 4);
+
+        // The rank numbers in status.lua are fake. Internally, rank is really
+        // a number between 0->1600. We fix the number here but really should solve
+        // the root problem "some day" but it will need a rank system rewrite.
```
cocosolos: So, the internal rank is just the max skill level attainable at the current rank. The rank numbers in status.lua are just for our convenience and should probably be left as-is since they're used in a few different calculations so it would just shift around where these adjustments are, resulting in needing to do stuff like changing:
```lua
function getAdvImageSupportCost(player,craftID)
    local Rank = player:getSkillRank(craftID)
    return (Rank/100)*30
end
```
to:
```lua
function getAdvImageSupportCost(player,craftID)
    local Rank = player:getSkillRank(craftID)
    return (Rank+1)*30
end
```
So just rearranging where we're doing calculations instead of actually fixing/improving anything. I would just remove this comment. Everything else looks good I think.

Kreidos: My intention is to point out why the number was being modified in such a counter-intuitive manner. Ideally, I prefer not having such artificial abstractions, as they cause this kind of confusion whenever that boundary needs to be crossed. It would also be a very simple manner to just redefined the status.lua values to something representing their *real* value after the appropriate changes were made, so I don't think the comment is invalid. I will change it as you request.

cocosolos: It's not counter intuitive though, if your rank is 1, your skill cap is 20. These calculations aren't converting rank into the retail accurate version of rank, they're calculating the max skill level available at your rank, or whatever else it needs to calculate (like support pricing) which is similarly derived from your rank.

cocosolos (approved):

src/map/lua/lua_baseentity.cpp
kaincenteno: In the image attached you can see the rank being passed by the packet is at 0, 200, 300, 400, 500 (0 is not joined a guild, 100 is the first rank, i think is amateur)

Hope this helps with the confusion on why its increments of 100. (captures from retail last year)

image

thanks for the review coco :)

cocosolos: It's not actually the rank being passed though, its the skill cap which is derived from the rank. 100 = 10.0

cocosolos: Actually it might even make sense to move the
```cpp
        guildRank = (guildRank + 1) * 100;
```
from addShopItem and doing that calculation in the packet like so
```cpp
        ref<uint16>(i * 12 + 0x12) = (PChar->Container->getGuildRank(slotID) + 1) * 100;
```

Kreidos: I disagree with pushing the reality of the underlying rank value so deep into the stack that it's virtually invisible. I can't help but imagine that at some point, the obfuscating *16-value rank metaphor* we adopt would will lead to issues. That could be someone running into a problem they can't solve due to values not being what they expect, and so may believing (perhaps based on completely accurate packet captures, as Kain did) that they need to write new functions to workaround an imagined limitation.

Edit: If the issue is wanting user-friendly identifiers for the ranks, this would be a use-case for an ENUM.

cocosolos: Rank just corresponds to Amateur, Recruit, Initiate, Novice, Apprentice, Journeyman, Craftsman, Artisan, Adept, Veteran, and Expert. The other ranks are not implemented, but that's what a rank is. The value being used in the shop packet is clearly the skill cap corresponding to the rank, why would it make sense to rank up by 100 ranks every time you rank up when a rank is those things listed? Likewise the price of certain things are determined by a different calculation using the rank, and the guild turn in item also determines the item of the day by indexing with the rank. It is very possible PVLV has the [wrong label](https://github.com/ZeromusXYZ/PacketViewerLogViewer/blob/00ffbcc66a373a811255290f0d0a221eed09fdc6/helpers/PacketParsers.cs#L1517), it wouldn't be the first time. There's a lot of stuff that's still unknown or just best guess.

ibm2431 (commented):
What's preventing the value sent by retail from being the rank's skill cap?

Sure, we're seeing the "rank" words be lit up on the shop screen, but that doesn't mean the value controlling what gets lit up _is_ the rank.

If there's _other_ evidence that the game consistently uses increments of 100 to represent ranks (guild NPC params for ranking up, params for daily items, DAT file values, etc) then we should match those values.

But I'm not certain the contents of just one packet are enough to say with certainty that the internal representation of a rank is equal to those contents.

We can keep ranks at their current values while being clear about the contents of the packet:
```cpp
 ref<uint16>(i * 12 + 0x12) = someUtils::getRankCap(guildRank)
```

**Actual review:** I see no problems with the code as designed.

src/map/lua/lua_baseentity.cpp
```diff
@@ -3792,16 +3794,25 @@ inline int32 CLuaBaseEntity::addShopItem(lua_State *L)
...
     uint8 slotID = ((CCharEntity*)m_PBaseEntity)->Container->getItemsCount();
-
     ((CCharEntity*)m_PBaseEntity)->Container->setItem(slotID, itemID, 0, price);
 
+    if(lua_isnumber(L, 3) && lua_isnumber(L, 4))
+    {
+        uint8 guildID = (uint8)lua_tonumber(L, 3);
+        uint16 guildRank = (uint16)lua_tonumber(L, 4);
+
+        guildRank = (guildRank + 1) * 100;
```
ibm2431: Instead of doing the math in baseentity and calling it the "rank", can instead...

src/map/packets/shop_items.cpp
```diff
@@ -51,6 +51,8 @@ CShopItemsPacket::CShopItemsPacket(CCharEntity * PChar)
 		ref<uint32>(i * 12 + 0x08) = PChar->Container->getQuantity(slotID);
 		ref<uint16>(i * 12 + 0x0C) = PChar->Container->getItemID(slotID);
 		ref<uint8>(i * 12 + 0x0E) = slotID;
+        ref<uint8>(i * 12 + 0x10) = PChar->Container->getGuildID(slotID);
+        ref<uint16>(i * 12 + 0x12) = PChar->Container->getGuildRank(slotID);
```
ibm2431: Perform the math in the outgoing packet here instead. We don't know retail internally represents ranks with values of 100, 200, 300 based on just this packet (but if there's other indications, that's a different story). So long as we're passing the term "guild rank" around core, should probably be kept consistent across different uses (ie: a standard enum of 1-16).

If don't want to do the math each iteration: can do and set it in the binding, but just rename the "guildRank" in the container to something like "skillCap". That way it's clear that it's not to be confused with the 1-16 ranks, and is still an accurate descriptor of what the value might represent.

ibm2431 (changes requested):
Last meeting:
> ibm2431: The code seems to be doing what it's meant to do
ibm2431: I left a comment re: rank and packet values
ibm2431: But I don't know who's seen it
cocosolos: my only thing is, since i dont see that as rank, is the naming of set/getGuildRank. if we are sticking with 0-15 as rank, i think either the functions should be renamed so as not to cause confusion (getguildrank would be returning in the hundreds) even though this isnt player facing. i would prefer to keep the naming convention and just move the calculation to where its being used in the packet
cocosolos: im not gonna hold though, merging is fine because its not player facing and it does work
ibm2431: Well, difficulty is that "getGuildRank" is being invoked on the container, not the player
cocosolos: yeah, thats why i would rather just move the calculation to the packet
ibm2431: If it were a function from a player, I'd 100% hold if the value returned was the hundreds
ibm2431: Yeah, calculation in the packet sounds good to me
ibm2431: Whether through a helper function (which I'm not positive we have) or math
cocosolos: if for whatever reason someone is using the container getGuildRank for something else, they can expect it to correcpond to our enumd ranks
cocosolos: its just a 2 line change i commented
cocosolos: delete 1 line from baseentity, and move/adjust it to the line existing in the packet
ibm2431: I'll be the villain and hold~

src/map/packets/shop_items.cpp
Kreidos: Adjusted it to be done here at the packet.

One calculation per trade item isn't much so I'm not concerned there, and if we're going to push the rank metaphor down one layer to the variable, may as well push it all the way down to the packet.

ibm2431 (approved):

Issue #663, opened 2020-05-28
pet abiliities being treated as regular mobskills


TeoTwawki:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
This has been getting a lot of discussion lately in the discord so I am making an issue (didn't see an existing one) so people can find this info easily and be aware of it - it has been a longstanding known quirk that the system for making pets use their skills is riding off the mobskills system in places that retail doesn't do this. They work, except when the skill doesn't exist in the monster version of the list (Siren) and certain 3rd party addons expected the retail version of the packet.

> [2:39 PM] Teo: the person you are quoting is mostly correct that it is done wrong using monster moves
[2:40 PM] Teo: the core has this weirdness where the pet "ability" is connected with a mob move and shouldn't be (hence why ppl had problems adding siren's moveset)
[2:40 PM] Teo: real pet skills are in the player ability list in rom\181\72.dat
[2:43 PM] Teo: player ws, ability, avatar, and monstrosity stuff all there
[2:44 PM] Teo: monsters skills are in rom\24\80.dat
the way is coded right now the pet's ability had to exist in both dats for us to use it, and its the mob version firing off.

![](https://media.discordapp.net/attachments/639659267003252748/715585087072108655/0x28-2.png?width=432&height=426)

EpicTaru:
Also per Thorny in the Ashita discord, the accepted work around (until this is fixed in Topaz) is:

![bp workaround 1](https://user-images.githubusercontent.com/26195580/83174484-17c31200-a0e0-11ea-8d86-35978bfb2f7c.png)
![bp workaround 2](https://user-images.githubusercontent.com/26195580/83174495-198cd580-a0e0-11ea-9753-26017ed18710.png)
![bp workaround 3](https://user-images.githubusercontent.com/26195580/83174502-1a256c00-a0e0-11ea-806a-d33468aa82ca.png)

EpicTaru:
The above snippet is from my SMN Ashitacast profile so set names probably won't be the same for what others use in their SMN Ashitacast profile.  Adjust accordingly, ofc.

Sinnersslayer:
Thing is that if you define it like this, you won't be able to swap BP delay set before using BP dmg.
I have found another solution.
[...]

PR #804, opened 2020-07-03
ability ids and offsets

illzbee:
**_I affirm:_**


src/map/packets/char_abilities.cpp
```diff
@@ -34,7 +34,7 @@ CCharAbilitiesPacket::CCharAbilitiesPacket(CCharEntity* PChar)
 	this->size = 0x72;
 
 	memcpy(data+(0x04), PChar->m_WeaponSkills, 32);
-	memcpy(data+(0x46), PChar->m_Abilities,    62);
+	memcpy(data+(0x44), PChar->m_Abilities,    62);
```
zach2good, line 37: Nice find!

For anyone (like me) coming here and seeing this 62 and thinking something is up, it's meant to be 62 and not 64:
https://github.com/project-topaz/topaz/blob/release/src/map/entities/charentity.h#L196

zach2good (commented):
Looks good at a glance, but I wouldn't sign off on this until I had time to dig through caps of the packet (0xAC) since it's such a low-level change. 

Here is an annotated cap from PVLV for anyone passing through:
image

There isn't any information in Windower fields (https://github.com/Windower/Lua/blob/dev/addons/libs/packets/fields.lua) to back it up either :(

scripts/globals/ability.lua
```diff
@@ -10,424 +10,620 @@ tpz = tpz or {}
...
+    PERPETUANCE        = 316,
+    IMMANENCE          = 317,
+    SMITING_BREATH     = 318,
+    RESTORING_BREATH   = 319,
+    KONZEN_ITTAI       = 320,
+    BULLY              = 321,
+    MAINTENANCE        = 322,
+	BRAZEN_RUSH        = 323,
+	INNER_STRENGTH     = 324,
+	ASYLUM             = 325,
```
ibm2431: Seems to be tab characters throughout these lists~

sql/abilities.sql
```diff
@@ -38,502 +38,617 @@ CREATE TABLE `abilities` (
...
+INSERT INTO `abilities` VALUES (247,'composure',5,50,1,300,50,0,0,215,2000,0,6,20.0,0,1,80,0,0,'WOTG');
+INSERT INTO `abilities` VALUES (248,'yonin',13,40,1,180,146,0,0,218,2000,0,6,20.0,0,1,600,0,4,'WOTG');
+INSERT INTO `abilities` VALUES (249,'innin',13,40,1,180,147,0,0,219,2000,0,6,20.0,0,1,60,0,4,'WOTG');
+-- INSERT INTO `abilities` VALUES (250,'avatars_favor',22,1,1,0,300,0,0,???,2000,0,6,20.0,0,450,900,0,0,NULL);
+INSERT INTO `abilities` VALUES (251,'ready',9,25,1,0,102,0,0,83,2000,0,6,18.0,0,0,0,902,64,NULL);
+-- INSERT INTO `abilities` VALUES (252,'restraint',22,1,1,0,300,0,0,???,2000,0,6,20.0,0,450,900,0,0,NULL);
+-- INSERT INTO `abilities` VALUES (253,'perfect_counter',22,1,1,0,300,0,0,???,2000,0,6,20.0,0,450,900,0,0,NULL);
+-- INSERT INTO `abilities` VALUES (254,'mana_wall',22,1,1,0,300,0,0,???,2000,0,6,20.0,0,450,900,0,0,NULL);
+-- INSERT INTO `abilities` VALUES (255,'divine_emblem',22,1,1,0,300,0,0,???,2000,0,6,20.0,0,450,900,0,0,NULL);
+-- INSERT INTO `abilities` VALUES (256,'nether_void',22,1,1,0,300,0,0,???,2000,0,6,20.0,0,450,900,0,0,NULL);
```
ibm2431, line 278: I appreciate the placeholders 😄

src/map/ability.h
```diff
@@ -45,593 +45,622 @@ enum ADDTYPE
...
+    ABILITY_DRAGON_BREAKER     = 292,
+    ABILITY_SOUL_JUMP          = 293,
+    ABILITY_STEADY_WING        = 295,
+    ABILITY_MANA_CEDE          = 296,
+    ABILITY_EFFLUX             = 297,
+    ABILITY_UNBRIDLED_LEARNING = 298,
+    ABILITY_TRIPLE_SHOT        = 301,
+	ABILITY_ALLIES_ROLL        = 302,
+	ABILITY_MISER_S_ROLL       = 303,
+	ABILITY_COMPANION_S_ROLL   = 304,
```
ibm2431: Looks like some spacing issues in this file too

src/map/ability.h
```diff
@@ -45,593 +45,622 @@ enum ADDTYPE
...
+//	ABILITY_EXTIRPATING_SALVO  = 787,
+//	ABILITY_CLARSACH_CALL      = 960,
+//	ABILITY_WELT               = 961,
+//	ABILITY_KATABATIC_BLADES   = 962,
+//	ABILITY_LUNATIC_VOICE      = 963,
+//	ABILITY_ROUNDHOUSE         = 964,
+//	ABILITY_CHINOOK            = 965,
+//	ABILITY_BITTER_ELEGY       = 966,
+//	ABILITY_SONIC_BUFFET       = 967,
+//	ABILITY_RORNADO_II         = 968,
```
ibm2431: Should this be TORNADO?

src/map/entities/charentity.cpp
```diff
@@ -1020,7 +1020,7 @@ void CCharEntity::OnAbility(CAbilityState& state, action_t& action)
         action.id = this->id;
         action.actiontype = PAbility->getActionType();
         //#TODO: unoffset this
```
ibm2431: Can remove these `//#TODO: unoffset this` messages since the TODO has finally be fulfilled! 😉

src/map/entities/petentity.cpp
```diff
@@ -199,7 +199,7 @@ void CPetEntity::OnAbility(CAbilityState& state, action_t& action)
         action.id = this->id;
         action.actiontype = PAbility->getActionType();
         //#TODO: unoffset this
```
ibm2431: One here too.

src/map/packets/char_abilities.cpp
ibm2431: [Windower packet library has the following to say regarding these packets](https://github.com/Windower/Lua/blob/dev/addons/libs/packets/fields.lua#L3223):
> --0x0AA, 0x0AC, and 0x0AE are all bitfields where the lsb indicates whether you have index 0 of the related resource.

PVLV has the fields starting at 0x46, but Zeromus may have just copied us.

This solution pulls all the bits back by two bytes, which seems to make all the abilities line up with client expectations. But I'm not _100% positive_ if this is "proper" or if we should be messing with a LSB instead for certain action categories to deal with whatever wonky indexing the client is doing with the DATs.

@TeoTwawki might know more about DAT-indexing once the ID hits the client. I do know that all the IDs put in this PR align with what the server sends out on retail. But I don't know if our offset has just always been wrong (either an index shift, or intentionally so to align 0-based ID bits to the packet), or if it _should_ still be at 0x46 _with_ these IDs and we're missing a flag in the packet telling it which index to use. This signification may be related to the #663 family of issues; as while not part of the original quote in that issue, Thorny expounded further:
> it isn't just going to be a wrong table pull, their mapping is wrong but there are also some kind of bitflags to tell the game which table it's pulling from and they aren't documented anywhere

scripts/globals/ability.lua
illzbee: It is working this way. I used tab on the abilities that needed to be programmed, I’ll remove them if needed just let me know. Thanks!

src/map/ability.h
illzbee: It should thanks for finding that!

src/map/entities/petentity.cpp
illzbee: removed

src/map/entities/charentity.cpp
illzbee: removed

src/map/ability.h
illzbee: all set on spelling

src/map/ability.h
illzbee: took out tab keys

scripts/globals/ability.lua
illzbee: took out tab keys

sql/abilities.sql
illzbee: no problem, am also working on those abilities XD

src/map/packets/char_abilities.cpp
TeoTwawki: > TeoTwawki might know more about DAT-indexing once the ID hits the client. I do know that all the IDs put in this PR align with what the server sends out on retail. 

@ibm2431 back when darkstar was a young project the offset used to make it match a dat we were grabbing the IDs from at that time, however I believe that SE has drastically changed that dat since that time and I fidn nothing matching the 16 offset now. Some other things that will not be affected by this PR are offset _elsewhere_ to match dats currently. The only reason I know of to do so is easy client extraction.

As far as I knew, the wrong mapping was us not having pets use pet abilities proper and instead having player ability trigger a mobskill. sending out wrong packet from our pets. If that's happening with other things as well, I was not made aware of it.

ibm2431 (approved):
Looking good!

(Whoever hits the merge button, squash please~!)

cocosolos (approved):

PR #807, opened 2020-07-04
Inquartata trait

Apocarypse:
**_I affirm:_**

Added mod enum for Inquartata in modifier.h
Added entries for Inquartata in traits.sql
Applied Inquartata bonus logic to GetParryRate()

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1726,11 +1726,8 @@ namespace battleutils
                 // Inquartata grants a flat parry rate bonus.
                 if (CCharEntity* PChar = dynamic_cast<CCharEntity*>(PDefender))
                 {
-                    if (charutils::hasTrait(PChar, TRAIT_INQUARTATA))
-                    {
-                        int16 inquartataBonus = PChar->getMod(Mod::INQUARTATA);
-                        parryRate = parryRate + inquartataBonus;
-                    }
+                    int16 inquartataBonus = PChar->getMod(Mod::INQUARTATA);
+                    parryRate += inquartataBonus;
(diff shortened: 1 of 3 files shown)


Kreidos (approved):
Looks good to me!

src/map/utils/battleutils.cpp
```diff
@@ -1723,6 +1723,15 @@ namespace battleutils
                     parryRate = parryRate + issekiganBonus;
                 }
 
+                // Inquartata grants a flat parry rate bonus.
+                if (CCharEntity* PChar = dynamic_cast<CCharEntity*>(PDefender))
```
ibm2431: Sverdheid and Zurko-Bazurko are RUN mobs that players need to fight.

Kreidos: I think based on what was said on the discord, that ibm's idea was that if you aren't needing the trait check, then there's no need to cast to CCharEntity*. You can just use PDefender directly as CBattleEntity has the getMod method. :)

ibm2431 (approved):

cocosolos (approved):

TeoTwawki:
@Apocarypse @cocosolos @ibm2431 modifiers in modifier.h and statsu.lua should mirror each other until such time as lua can automagically use the core enums, even if the mod is never used in script so that future editor don't mistakenly use the same ID - this has happened multiple times in the older project history until I became its watchdog in dsp.

ibm2431:
_sees this merged PR in notifications_
_sees Teo as a participant_

"I _did_ miss one, didn't I?"

_clicks_

"Yeeeep. Whoops."

PR #809, opened 2020-07-05
Trust Balancing 2

zach2good:
**_I affirm:_**

- Fixes the recent "Trust Balancing" patch with something a little less borked.
- Fixes trust's TP decaying while they rest

TODO:
Bring trust HP/MP down to something reasonable, per "job". Further modifications can be made per-trust using Mods if needed


TeoTwawki:
> Bring trust HP/MP down to something reasonable, per "job". Further modifications can be made per-trust using Mods if needed

on normal mobs this would be group tables lv x family system value, since trusts don't have a group its just being based on lv they get from the player, but teh family system value still exists. There should be multiple usable families for these. The oens mimicking player race are already done but I don't know if their job in pool is taken into account at all currently (it probably should (?), even for regular non-trust mobs, if it does not already)

scripts/globals/spells/trust/excenmille.lua
```diff
@@ -41,9 +41,7 @@ function onMobSpawn(mob)
     mob:addSimpleGambit(ai.t.SELF, ai.c.TP_GTE, 1000,
                         ai.r.WS, ai.s.SPECIFIC, tpz.ws.DOUBLE_THRUST)
 
-    -- Excenmille is a PLD who uses a Polearm, so raise his MAIN_DMG_RATING (up from 1H Sword levels)
-    local increase_damage_by_percent = 30
-    mob:addMod(tpz.mod.MAIN_DMG_RATING, mob:getWeaponDmg() * (1.0 + (increase_damage_by_percent / 100)))
+    mob:addMod(tpz.mod.STORETP, 25)
```
TeoTwawki, line 44: If his tp per hit isn't lining up, are we sure the base delay is correct or that he doesn't secretly sub sam?

Standard low end polearm delay on nq weapons is 396, which is pretty close. The higher end slow ones are 492 and 480. A few oddball ones are 416 (Partisan) and 435 (Envy Spear) and 551 (Swartz Lance).

zach2good: I'll have to sit with him in retail to make sure his TP stats line up

zach2good: (This is based off a SE update where they said they added the "Store TP Trait" to Excenmille, rather than trying to balance him properly)

TeoTwawki: Ahh, I was unaware we had a word of god source for his storeTP. :+1:

cocosolos (approved):

ibm2431 (approved):

PR #811, opened 2020-07-05
Enhance !exec command

zach2good:
Defines "player" and "target" in exec command.

Allows: `!exec player:setHP(10)` and `!exec target:setHP(10)` etc.

On error, prints the basic string you tried to execute and the error from lua (which starts with the appended values):
image

**_I affirm:_**


Kreidos (approved):
Great idea! Looks good.

ibm2431 (approved):

cocosolos (approved):

PR #820, opened 2020-07-07
Add database tool

### LINKED ISSUE #288: Database Scripts (in /tools) should reference environment variables for MySQL Path
<a href="https://github.com/ShelbyZ"></img></a> **Issue by [ShelbyZ](https://github.com/ShelbyZ)**
_Friday Apr 26, 2019 at 04:47 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 5895_

----

**_I have:_**

- [x] _looked code way later than I should have_

In both Database_xxx.bat scripts there are direct references to path for Program Files.  Depending on the install and target system this could be either **Program Files** or **Program Files (x86)**.  To remove the need for a user to change this file we can use environment variables and check both locations.  Setting a variable when we find that option.

Database_initial_setup.bat
github/DarkstarProject/darkstar/blob/d0bc0d2cd3a6cb874c279b96811aebb692b3bbd6/tools/Database_initail_setup.batDarkstar Issue L33-L34

Database_update.bat
github/DarkstarProject/darkstar/blob/d0bc0d2cd3a6cb874c279b96811aebb692b3bbd6/tools/Database_update.batDarkstar Issue L33

Sample
```batch
if exist %ProgramFiles\MySQL\MySQL Server 5.7\bin (
    set mysql="..."
)

if not defined mysql (
    if exist %ProgramFiles(x86)\MySQL\MySQL Server 5.7\bin (
        set mysql="..."
    )
)
```

cocosolos:
-Deprecate migrations script
-Refactor give_items

Saves commit hash to version.conf when updating. On future updates, will
diff that hash against current hash to only import files that have
changed. Saves settings.

Optional args (both update options will also backup and migrate):
`python dbtool.py backup` will backup databse.
`python dbtool.py backup lite` will create a backup of only the tables marked as protected (player_data in config)
`python dbtool.py update` will check versions and only update if necessary.
`python dbtool.py update full` will do a full update.
`python dbtool.py migrate` will check and perform any needed migrations.

**_I affirm:_**


59blargedy (commented):
Tested the version before this one for both initial setup and update, adding items to ignore list and performing backups. Had one issue that was resolved in this version, everything else worked great!

cocosolos:
> Tested the version before this one for both initial setup and update, adding items to ignore list and performing backups. Had one issue that was resolved in this version, everything else worked great!

IDK how recent this was, but a lot has changed since we discussed this! Adding items to the ignore list is now saved to a file and persists, and you can even provide and change the MySQL bin path if it's not found in the system paths. I also changed how some stuff is displayed to hopefully make it easier to understand and use.

.gitignore
```diff
@@ -13,6 +13,7 @@ TestResults
 *.user
 *.sln.docstates
 *.DS_Store
+*.yaml
```
ibm2431: Dunno if this might interfere with being able to upload/edit/clone YAML files that we want shared, like Travis or containers. May want to assign the ignore to just the tools subdirectory for now.

cocosolos: Updated to ignore just this file for now.

ibm2431 (approved):

PR #823, opened 2020-07-07
Implement Eco-Warrior quests

tankfest:
**_I affirm:_**

This branch will implement the two missing Eco-Warrior quests, and refurbish Eco-Warrior Windurst.

Currently Eco-Warrior Windurst has a few bugs and incorrect dialogue/options, commit 6b5d0e6 fixes those and cleans the scripts/variables.

This is a new PR for #743 so that requested changes could be made.


Wiggo32 (approved):
I'm not familiar with how these work on retail, however these are lua scripts and only affect the mobs/npcs associated with the quest so no real chance of breaking much of anything and I trust the author has scripted events to behave accordingly with the best of their ability. I see no reason not to approve and if tweaking is needed later due to new research, that can always be added.

TeoTwawki:
@tankfest for future reference, you can use force push to overwrite existing PR's with new content so that you don't need to create new ones. this keeps the comment history intact while making the content match your current work.

TeoTwawki (approved):
Looks good enough to me. additional features need added to make this 100%.

The previous PR for these had non retail behavior (needed more than tweaking) but this one doesn't. I will open an issue for the missing feature.

TeoTwawki:
#924 as promised

PR #831, opened 2020-07-10
Add distinction between days and elements + fix elemental ordering

### LINKED ISSUE #291: Elemental ID Ordering Inconsistent
<a href="https://github.com/ibm2431"></img></a> **Issue by [ibm2431](https://github.com/ibm2431)**
_Friday May 10, 2019 at 17:22 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 6002_

----

**_I have:_**

**_Client Version_** (type `/ver` in game) **:** n/a

I am presenting the can of worms. A lot of us already know this, but I couldn't find it reported as an Issue.

[Days of the week go in an order inconsistent with retail's elemental wheel](github/DarkstarProject/darkstar/blob/master/src/map/vana_time.hDarkstar Issue L36-L46). The fault is on SE here, and our ordering of the days is correct.

However, in many locations [our ordering of elements follows the days of the week instead of retail's wheel.](github/DarkstarProject/darkstar/blob/master/scripts/globals/magic.luaDarkstar Issue L16-L28)

[It causes people to try to use the subEffect table to represent an elemental spirit instead of the magical element, using a family bridge to link family to the subEffect.](github/DarkstarProject/darkstar/blob/c8c9564ae5863d4c489567b8df7d3723086badee/scripts/globals/pets/avatar.luaDarkstar Issue L156-L171) This is because despite how the [spirit pacts spells follow the elemental wheel](github/DarkstarProject/darkstar/blob/5e04622f60dfd196dc1630d603f24d2ec8403383/sql/spell_list.sqlDarkstar Issue L345-L352) (as do [our elements for the spirits](github/DarkstarProject/darkstar/blob/master/sql/mob_family_system.sqlDarkstar Issue L164-L171)) our elemental table in Lua doesn't.

But we can't simply replace the values in the Lua table, because [our element column for spells uses the day of the week values](github/DarkstarProject/darkstar/blob/master/sql/spell_list.sqlDarkstar Issue L202-L231). This also causes our element columns to not increase in a linear order with our spell lists.

This inconsistency makes [our elemental modifiers not go in the same order](github/DarkstarProject/darkstar/blob/3e46a535c9fe3ab405a829d4de117c638c44493a/src/map/modifier.hDarkstar Issue L51-L58) as our elemental listing. (The modifiers go in the correct order, the inconsistency is with our element tables).

It causes [our Damage Types](github/DarkstarProject/darkstar/blob/2924080116c08539ab245f2a8947f1d448237bdf/src/map/entities/battleentity.hDarkstar Issue L255-L262) to not go in the same order as the [Subeffects _just 20 lines below_ the type definition](github/DarkstarProject/darkstar/blob/2924080116c08539ab245f2a8947f1d448237bdf/src/map/entities/battleentity.hDarkstar Issue L286-L296), or the [order of the elements in queries for resistances](github/DarkstarProject/darkstar/blob/09b3ddfe5f24190203a2ceb225c479bcc699d045/src/map/instance_loader.cppDarkstar Issue L105-L111).

For fun, before [seeing the order our Enspells go in, try to guess what it is](github/DarkstarProject/darkstar/blob/2d0add60f09a79bc03a8ffd0ede33d408ec77f3e/src/map/utils/battleutils.hDarkstar Issue L48-L55). Then give [COR shots](github/DarkstarProject/darkstar/blob/2d0add60f09a79bc03a8ffd0ede33d408ec77f3e/src/map/ability.hDarkstar Issue L159-L166) and [PUP Maneuvers](github/DarkstarProject/darkstar/blob/2d0add60f09a79bc03a8ffd0ede33d408ec77f3e/src/map/ability.hDarkstar Issue L175-L182) a guess. How about [Avatar perpetuation reduction](github/DarkstarProject/darkstar/blob/5e04622f60dfd196dc1630d603f24d2ec8403383/src/map/utils/charutils.cppDarkstar Issue L4404-L4414)?

It causes our [internal crystal elements](github/DarkstarProject/darkstar/blob/09b3ddfe5f24190203a2ceb225c479bcc699d045/src/map/utils/synthutils.hDarkstar Issue L39-L49) to be misaligned with the item IDs for those crystals:
```sql
INSERT INTO `item_basic` VALUES (4096,0,'fire_crystal','fire_crystal',12,516,35,0,13);
INSERT INTO `item_basic` VALUES (4097,0,'ice_crystal','ice_crystal',12,516,35,0,30);
INSERT INTO `item_basic` VALUES (4098,0,'wind_crystal','wind_crystal',12,516,35,0,14);
INSERT INTO `item_basic` VALUES (4099,0,'earth_crystal','earth_crystal',12,516,35,0,13);
INSERT INTO `item_basic` VALUES (4100,0,'lightning_crystal','lightng._crystal',12,516,35,0,30);
INSERT INTO `item_basic` VALUES (4101,0,'water_crystal','water_crystal',12,516,35,0,15);
INSERT INTO `item_basic` VALUES (4102,0,'light_crystal','light_crystal',12,516,35,0,80);
INSERT INTO `item_basic` VALUES (4103,0,'dark_crystal','dark_crystal',12,516,35,0,77);
```
Which results in [additional effort in mapping the item ID to the correct element](github/DarkstarProject/darkstar/blob/master/src/map/utils/synthutils.cppDarkstar Issue L722-L764).

It causes you to, upon reading the comment ["We are able to get the correct elemental mod here by adding the element to it since they are in the same order"](github/DarkstarProject/darkstar/blob/master/src/map/utils/synthutils.cppDarkstar Issue L630-L631) to go verify [what that order mods for "reduce fail rate when using ELEMENT crystal" go in](github/DarkstarProject/darkstar/blob/3e46a535c9fe3ab405a829d4de117c638c44493a/scripts/globals/status.luaDarkstar Issue L1490-L1497). It works out in the end, but you had to go confirm to make sure. [What about Moghancements](github/DarkstarProject/darkstar/blob/186cbafb2d091b8f1c37ea1e17f4401604cc246a/src/map/items/item_furnishing.hDarkstar Issue L35-L42)?

I'll spare a treatise about everything relating to, or depending on, weather, and/or days of the week.

I doubt it'd be _too_ controversial to say that the only thing that should follow SE's days of the week order is our definition of days of the week. _Everything else_ should be using the elemental wheel which SE uses for everything besides week days.

Now, if there were a "Hard Fix" tag, this would earn it due to how pervasive these inconsistencies are throughout the project. _I_ certainly have no plans to open the worm can and start digging through it. But if someone feels _particularly_ ambitious, here's something you can look into.

ibm2431:
They may not know it yet, but this Pull Request will make some people unhappy. But it's _long_ overdue.

---

For those unaware of the problem (skip to the next horizontal rule if you're already familiar):

A long time ago, someone decided to make enums go in the following order:
> Firesday, Earthsday, Watersday, Windsday, Iceday, Lightningsday, Lightsday, Darksday

This was the correct decision, as it follows retail. However, a later decision then came:
> When asked what element the current day is, we'll return the current day.

But on retail - with the exception of days of the week and Avatars (and their abilities) - elements follow the elemental wheel:
> Fire Ice Wind Earth Thunder Water Light Dark

Everything that retail does referencing elements follows that order: spell IDs, crafting crystal item IDs, elemental ability IDs, models, etc.

(Yes @TeoTwawki, I know spikes don't, but those are the exception~! 😉 )

The mismatch between retail's standard elemental ordering and ours has a long list of rippling effects.

**For some examples see:**
https://github.com/DarkstarProject/darkstar/issues/6002

These effects at their most benign require additional bridge tables which shouldn't be necessary. One recent example being the [mapping table that @Omnione needed for Geomancer bubble effects and models](https://github.com/project-topaz/topaz/blob/1a942bbe006d6e2b0778177b9940dbff9e33bcc2/scripts/globals/geo.lua#L14-L41) - which was the last straw. At their worst the elemental ordering mismatch causes incorrect elemental behavior when wires get crossed from using the correct order to our incorrect order.

---

So this Pull Request:
1) Creates a distinction between days of the week and magical elements
2) Adjusts all things referencing magical elements to do so in the corrected elemental order
3) Adjusts things referencing _the current day_ to check _the day_ and _not_ its element; _this distinction is important_

Forewarning: I will have missed something. I don't know what it is yet. I've done the best I can to find and address all references to days and elements that I can think of.

**I deliberately broke this up into multiple commits for ease of review. Use the commit-by-commit review feature!**

For example, I broke the SQL changes into "two steps" of commits. The "first step" commit replaces the old values with a string, so a reviewer can quickly gauge that the old values were correctly picked up and marked. The "second step" commit replaces those strings with the new, updated value. So don't fret about SQL change diff, just verify that yes, I can indeed correctly use regex~!

**I have deliberately not touched crafting due to #779 reworking the system and outright removing many of the references currently present!** Either @Xaver-DaRed or I will have to rebase and adjust our PRs accordingly depending on what gets merged when. **Any branch from this PR should _not_ go into release or canary until changes to the crafting system have been finalized and the branch adjusted to correct crafting elements (if still required).**

I'll be doing a "review" of this PR to preemptively answer questions a reviewer might have. I've been working on this over the course of a week, and may come across something that needs fixing. Will push follow-up commits if I do.

Closes #291 

**_I affirm:_**


TeoTwawki:
> (Yes @TeoTwawki, I know spikes don't, but those are the exception~! wink )

The subeffect orders, they haunt me so! WHY SE? WHY?!

scripts/commands/setweather.lua
```diff
@@ -3,7 +3,7 @@
 -- desc: Sets the current weather for the current zone.
 ---------------------------------------------------------------------------------------------------
 
-require("scripts/globals/weather")
+require("scripts/globals/world")
```
ibm2431, line 6: [Because days of the week aren't "weather"](https://github.com/ibm2431/topaz/blob/master/scripts/globals/weather.lua#L30-L40). And [time](https://github.com/ibm2431/topaz/blob/master/scripts/globals/weather.lua#L42-L52) isn't either.

Due to the importance of days and weather, and how some mods reference them, `status.lua` was modified to require this new `world` global. As such, files which required both `status` and `weather` just had `weather` removed for being redunant.

Side: We'll have to revisit our global require system in the future. `status.lua` has become a "catch-all" for "important things I don't know where to put", and a bunch of the enums which need to be required should just always be available without requiring a scripter to know and remember to include a certain global.

scripts/globals/abilities/elemental_siphon.lua
```diff
@@ -23,20 +23,8 @@ function onAbilityCheck(player,target,ability)
...
-    elseif (spiritEle == 3) then
-        spiritEle = 4
-    elseif (spiritEle == 4) then
-        spiritEle = 2
-    elseif (spiritEle == 5) then
-        spiritEle = 6
-    elseif (spiritEle == 6) then
-        spiritEle = 3
-    end
+    local spiritEle = player:getPetID() + 1 -- get the spirit's ID, it is already aligned in proper element order
```
ibm2431, line 26: "Pet IDs" were one of the things already following the true elemental order, but their listed `element` in `pet_list` did not. So this script was trying to map their true-element-order ID to the old day-of-the-week elemental ordering. Confusing, huh?

scripts/globals/abyssea.lua
```diff
@@ -234,10 +237,15 @@ function getDemiluneAbyssite(player)
...
-    if weakness < 0 then weakness = 7 elseif weakness > 7 then weakness = 0 end
-    return yellow_weakness[weakness][math.random(#yellow_weakness[weakness])]
+    local day = VanadielDayOfTheWeek()
+    local day_weakness = math.random(day-1, day+1)
+    if day_weakness < 0 then
+        day_weakness = 7
+    elseif day_weakness > 7 then
+        day_weakness = 0
+    end
+    local element = tpz.magic.dayElement[day_weakness]
```
ibm2431, line 247: `tpz.magic.dayElement` can be used to determine any given day's element; it's in `magic.lua`

While it can be used for determining the current day's element, there's a handy `VanadielDayElement()` function given to you by core _without_ need to `tpz.magic.dayElement[VanadielDayOfTheWeek()]`

scripts/globals/effects/blood_weapon.lua
```diff
@@ -7,7 +7,7 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(tpz.mod.ENSPELL,16)
+    target:addMod(tpz.mod.ENSPELL,17)
```
ibm2431, line 10: There is an enum in core for "enspell IDs", which these effect scripts need to match.

(There isn't a copy of that enum table in lua.)

When messing with enspells, I discovered that Endark II wasn't in there; so I gave it a spot, and pushed blood weapon back.

scripts/globals/effects/enaero.lua
```diff
@@ -7,7 +7,7 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(tpz.mod.ENSPELL,4)
+    target:addMod(tpz.mod.ENSPELL, tpz.magic.element.WIND)
```
ibm2431, line 10: For the first 8 "enspell IDs", the ID matches the spell's element. So I told the tier I scripts to use a relevant element in effort to reduce magic numbers in scripts.

scripts/globals/effects/enaero_ii.lua
```diff
@@ -7,7 +7,7 @@ require("scripts/globals/status")
 -----------------------------------
 
 function onEffectGain(target,effect)
-    target:addMod(tpz.mod.ENSPELL,12)
+    target:addMod(tpz.mod.ENSPELL, tpz.magic.element.WIND + 8) -- Tier IIs have higher "enspell IDs"
```
ibm2431, line 10: In core, the "enspell IDs" are aligned such as:
Elemental Tier Is
Elemental Tier IIs
Other "enspell" effects (blood weapon)

scripts/globals/icanheararainbow.lua
```diff
@@ -2,7 +2,6 @@
 
 require( "scripts/globals/status")
 require( "scripts/globals/quests")
-require( "scripts/globals/weather")
```
ibm2431, line 5: This is the first example of a script no longer needed the `weather` global due to it already requiring `status`.

Sorry if I merge conflict you on #796 , @wrenffxi 😦

scripts/globals/magic.lua
```diff
@@ -14,25 +13,24 @@ tpz.magic = tpz.magic or {}
...
-
-tpz.magic.dayWeak             = {tpz.day.WATERSDAY,             tpz.day.WINDSDAY,               tpz.day.LIGHTNINGDAY,            tpz.day.ICEDAY,                tpz.day.FIRESDAY,             tpz.day.EARTHSDAY,               tpz.day.DARKSDAY,            tpz.day.LIGHTSDAY}
-tpz.magic.singleWeatherWeak   = {tpz.weather.RAIN,              tpz.weather.WIND,               tpz.weather.THUNDER,             tpz.weather.SNOW,              tpz.weather.HOT_SPELL,        tpz.weather.DUST_STORM,          tpz.weather.GLOOM,           tpz.weather.AURORAS}
-tpz.magic.doubleWeatherWeak   = {tpz.weather.SQUALL,            tpz.weather.GALES,              tpz.weather.THUNDERSTORMS,       tpz.weather.BLIZZARDS,         tpz.weather.HEAT_WAVE,        tpz.weather.SAND_STORM,          tpz.weather.DARKNESS,        tpz.weather.STELLAR_GLARE}
-local elementalObiWeak        = {tpz.mod.FORCE_WATER_DWBONUS,   tpz.mod.FORCE_WIND_DWBONUS,     tpz.mod.FORCE_LIGHTNING_DWBONUS, tpz.mod.FORCE_ICE_DWBONUS,     tpz.mod.FORCE_FIRE_DWBONUS,   tpz.mod.FORCE_EARTH_DWBONUS,     tpz.mod.FORCE_DARK_DWBONUS,  tpz.mod.FORCE_LIGHT_DWBONUS}
```
ibm2431, line 35: Regarding the removal of `elementalObiStrong` and `elementalObiWeak`:
> 
ibm2431: So. We apparently calculate day/element/obi bonuses/penalties in three separate locations in magic.lua
ibm2431: And all of them are wrong in some fashion or another
ibm2431: For most of them, we're checking to see if the caster is wearing the obi matching the "weak" element when casting a spell on the wrong day

ibm2431: Let's assume fire, since that's the one element enum we have right:
https://github.com/project-topaz/topaz/blob/release/scripts/globals/magic.lua#L35
https://github.com/project-topaz/topaz/blob/release/scripts/globals/magic.lua#L750-L777
ibm2431: Comes out to:
```
ele = fire -- spell:getElement()
if weather == rain then -- tpz.magic.singleWeatherWeak[ele]
    if chance < 0.33 or caster:mod(tpz.mod.FORCE_WATER_DWBONUS) then -- elementalObiWeak[ele]
        dayWeatherBonus = dayWeatherBonus - 0.10
```
ibm2431: Which will weaken fire magic in rain 100% of the time if the caster is wearing Suirin Obi; not Karin Obi
aether: oh should it be FORCE_FIRE_DWBONUS?
ibm2431: That's what it should be resolving to, yeah
ibm2431: I've already addressed it in this commit I'm working on, but it'll take maybe a week to PR it
ibm2431: I simply axed both "elementalObiStrong" and "elementalObiWeak"; replacing them with just "elementalObi"

scripts/globals/magic.lua
```diff
@@ -903,7 +901,11 @@ function addBonusesAbility(caster, ele, target, dmg, params)
 
     local mab = 1
     local mdefBarBonus = 0
-    if (ele > 0 and ele <= 6 and target:hasStatusEffect(tpz.magic.barSpell[ele])) then -- bar- spell magic defense bonus
+    if
+        ele >= tpz.magic.element.FIRE and
+        ele <= tpz.magic.element.WATER and
```
ibm2431, line 906: Just standard removal of magic numbers when I found them. There were sections of code dealing with elements which I missed on my first pass because they weren't referencing the word element or any actual elemental name - just numbers.

scripts/globals/mobskills/crystal_weapon.lua
```diff
@@ -10,20 +10,23 @@
...
 ---------------------------------------------
 
 function onMobSkillCheck(target,mob,skill)
     return 0
 end
 
 function onMobWeaponSkill(target, mob, skill)
-    local element = math.random(6,9)
+    local day = math.random(0,3)
+    local damage_type = tpz.damageType.FIRE + tpz.magic.dayElement[day] - 1
```
ibm2431, line 25: The four elements which can be chosen from for Crystal Weapon (confirmed to match old wiki's description, search Discord for "crystal weapon" for discussion) just happen to be the first four days of the week.

Elements start at 1, since `ELEMENT_NONE` is 0.

Days of the week start at 0, as that's the very first day of the week when a server is set to its earliest epoch - there is no "Nonday".

The previous code started from 6 to match the beginning _damageType_ for fire (various physical types are listed in the damageType before Fire is).

scripts/zones/Phomiuna_Aqueducts/npcs/_0rn.lua
```diff
@@ -16,13 +16,13 @@ function onTrigger(player,npc)
     player:messageSpecial(ID.text.LAMP_OFFSET+2) -- water lamp
     npc:openDoor(7) -- lamp animation
 
-    local element = VanadielDayElement()
+    local day = VanadielDayOfTheWeek()
 
-    if (element == 2) then -- waterday
+    if (day == tpz.day.WATERSDAY) then
```
ibm2431, line 21: This is one such instance when behavior depending on the _current day_ should be referencing _the current day_ and _not an "element"_.

In defense of the previous scripter, `VanadielDayElement()` _was_ the only way to get the current day, because it _was_ one in the same as the current day of the week. I'm adding `VanadielDayOfTheWeek()` myself in this PR.

scripts/zones/Sacrarium/Zone.lua
```diff
@@ -42,16 +42,16 @@ function onGameDay()
...
+        [tpz.day.WINDSDAY]     = {1,1,1,1,1,0,1,1,0,0,0,1,0,0,0,0,0,0},
+        [tpz.day.ICEDAY]       = {0,0,0,0,0,1,1,1,1,1,0,0,0,1,1,0,1,0},
+        [tpz.day.LIGHTNINGDAY] = {1,0,0,1,0,1,1,0,1,1,0,0,0,1,0,0,1,0},
+        [tpz.day.LIGHTSDAY]    = {1,0,0,0,0,0,0,1,1,0,1,0,1,1,0,1,0,1},
+        [tpz.day.DARKSDAY]     = {1,1,1,1,1,0,1,1,0,0,0,1,0,0,0,0,0,0},
     };
-    local doors = labyrinthDoorsByDay[VanadielDayElement()];
+    local doors = labyrinthDoorsByDay[VanadielDayOfTheWeek()];
     for i = 0, 17 do
         GetNPCByID(ID.npc.LABYRINTH_OFFSET + i):setAnimation(tpz.anim.OPEN_DOOR + doors[i+1]);
```
ibm2431, line 56: This is another example of why the distinction between days and elements is important~!

scripts/zones/Tavnazian_Safehold/npcs/Yurim.lua
```diff
@@ -116,15 +115,8 @@ function onEventFinish(player,csid,option)
...
-
-        if itemComplete > 0 and itemInProgress == itemComplete then
-            player:setLocalVar("NAME_OF_SCIENCE_complete", 0)
-            if npcUtil.completeQuest(player, OTHER_AREAS_LOG, tpz.quest.id.otherAreas.IN_THE_NAME_OF_SCIENCE, {item=itemComplete, var={"NAME_OF_SCIENCE_target"}}) then
-                player:confirmTrade()
-            end
-        else
-            printf("%s reached end of 'In The Name of Science' in a suspicious manner.  No item rewarded.", player:getName())
+        if npcUtil.completeQuest(player, OTHER_AREAS_LOG, tpz.quest.id.otherAreas.IN_THE_NAME_OF_SCIENCE, {item=itemInProgress, var={"NAME_OF_SCIENCE_target"}}) then
+            player:confirmTrade()
```
ibm2431, line 119: I tried to avoid changing unrelated things during my journeys, but couldn't resist here.

The localVar set and printf weren't needed, as it's _impossible_ for lua to reach csid 529's eventFinish unless core started it.

There are checks built into core which set a character's current CS when `player:startEvent` is called. Even if a player tries to inject a fake event option (ie: "finish") packet, core (and by extension lua scripts) ignore it unless core itself had already marked the player as currently being on that event.

sql/item_furnishing.sql
```diff
@@ -40,480 +40,480 @@ CREATE TABLE IF NOT EXISTS `item_furnishing` (
 -- Contenu de la table `item_furnishing`
 --
 
-INSERT INTO `item_furnishing` VALUES (1, 'pile_of_chocobo_bedding', 1, 520, 7, 2);
-INSERT INTO `item_furnishing` VALUES (2, 'simple_bed', 1, 520, 5, 3);
```
ibm2431, line 44: As all furniture has elemental auras, `item_furnishing`'s element field was "base 0 == fire". 

_And_ oddly enough, unlike a lot of other things which followed "day of the week" numbering, the table already followed the true elemental order.

This combination lead to much confusion on my part when first examining this file and "1" was listed as the element for pieces of furniture which wikis were listing as ice (as opposed to "fire" or "earth").

To avoid such confusion in the future, and make sure _all_ references to elements are consistent, I incremented the "element" for all furniture by 1.

So "8" == darkness == [Chocobo Bedding's listed element](https://www.bg-wiki.com/bg/Chocobo_Bedding)
"6" == water == [Simple Bed's listed element](https://www.bg-wiki.com/bg/Simple_Bed)

sql/item_mods.sql
```diff
@@ -17359,15 +17359,15 @@ INSERT INTO `item_mods` VALUES (15434,23,5);
 INSERT INTO `item_mods` VALUES (15435,1,7);
 INSERT INTO `item_mods` VALUES (15435,531,1); -- fire obi
 INSERT INTO `item_mods` VALUES (15436,1,7);
-INSERT INTO `item_mods` VALUES (15436,535,1); -- ice obi
+INSERT INTO `item_mods` VALUES (15436,532,1); -- ice obi
```
ibm2431, line 17362: Remember to review commit-by-commit, it's easier!

Here's some links to modifiers both before and after my changing:
[Before - "current" mods](https://github.com/ibm2431/topaz/blob/ecd6264e51ce7c35b2a91cb688ee57329cfe1d24/src/map/modifier.h)
[After - some mods changed to align with new elemental order](https://github.com/ibm2431/topaz/blob/7792b82e678e29f6e8eae30fd0904ec929a397a7/src/map/modifier.h)

sql/pet_list.sql
```diff
@@ -42,21 +42,21 @@ CREATE TABLE IF NOT EXISTS `pet_list` (
 --
 
 INSERT INTO `pet_list` VALUES (0, 'FireSpirit', 5830, 1, 99, 0, 1);
-INSERT INTO `pet_list` VALUES (1, 'IceSpirit', 5831, 1, 99, 0, 5);
-INSERT INTO `pet_list` VALUES (2, 'AirSpirit', 5832, 1, 99, 0, 4);
-INSERT INTO `pet_list` VALUES (3, 'EarthSpirit', 5833, 1, 99, 0, 2);
-INSERT INTO `pet_list` VALUES (4, 'ThunderSpirit', 5834, 1, 99, 0, 6);
-INSERT INTO `pet_list` VALUES (5, 'WaterSpirit', 5835, 1, 99, 0, 3);
+INSERT INTO `pet_list` VALUES (1, 'IceSpirit', 5831, 1, 99, 0, 2);
+INSERT INTO `pet_list` VALUES (2, 'AirSpirit', 5832, 1, 99, 0, 3);
```
ibm2431, line 46: For even more confusion later: `pet_list` was using the day-of-the-week values for `element`, but `mob_family_system` was using the true-elemental ordering in its `element` field for the purposes of crystal drops!

So you had Ice Spirits pets with `element` of 5, which were members of the Ice Elemental family with `element` of 2!

Honestly, now that the elemental ordering has been corrected, this `element` column could probably be removed from `pet_list`. But I didn't want to rock the boat too much in a single PR.

sql/spell_list.sql
```diff
@@ -109,13 +109,13 @@ INSERT INTO `spell_list` VALUES (49,'shell_ii',0x0000250025002800000000000000000
...
-INSERT INTO `spell_list` VALUES (54,'stoneskin',0x00001C002200000000000000000000000000002C0037,6,0,2,0,1,34,29,7000,30000,230,0,54,2000,4,0,1.00,0,165,0,204,NULL);
-INSERT INTO `spell_list` VALUES (55,'aquaveil',0x00000A000C00000000000000000000000000000D000F,6,0,3,0,1,34,12,5000,10000,230,0,55,2000,4,0,1.00,0,165,0,0,NULL);
-INSERT INTO `spell_list` VALUES (56,'slow',0x00000D000D0000000000000000000000000000000000,6,12,2,0,4,35,15,2000,20000,230,0,56,2000,0,0,1.00,1,176,0,204,NULL);
-INSERT INTO `spell_list` VALUES (57,'haste',0x00002800300000000000000000000000000000000000,6,13,6,0,91,34,40,3000,20000,230,0,57,2000,0,0,1.00,1,165,0,204,NULL);
-INSERT INTO `spell_list` VALUES (58,'paralyze',0x00000400060000000000000000000000000000000000,6,14,5,0,4,35,6,3000,10000,84,84,58,2000,0,0,1.00,1,176,0,204,NULL);
-INSERT INTO `spell_list` VALUES (59,'silence',0x00000F00120000000000000000000000000000000000,6,0,4,0,4,35,16,3000,10000,0,0,59,2000,0,0,1.00,1,176,0,204,NULL);
+INSERT INTO `spell_list` VALUES (53,'blink',0x000013001700000000000000000000000000001D0023,6,0,3,0,1,34,20,6000,10000,230,0,53,2000,4,0,1.00,0,165,0,0,NULL);
```
ibm2431, line 112: Heads up, anyone who modifies `spell_list` with non-retail changes!

(I know this includes you, @Lynnea1320 😅 )

We're changing a bunch of element IDs for spells!

sql/spell_list.sql
```diff
@@ -628,89 +628,89 @@ INSERT INTO `spell_list` VALUES (597,'sprout_smack',0x00000000000000000000000000
...
 INSERT INTO `spell_list` VALUES (697,'amorphic_spikes',0x00000000000000000000000000000062000000000000,3,0,0,0,4,43,79,500,58250,2,0,852,2000,0,0,1.00,0,0,0,58,'ABYSSEA');
-INSERT INTO `spell_list` VALUES (698,'wind_breath',0x00000000000000000000000000000063000000000000,3,0,4,0,4,43,26,1000,29500,2,0,853,2000,2,0,1.00,0,0,0,124,'ABYSSEA');
+INSERT INTO `spell_list` VALUES (698,'wind_breath',0x00000000000000000000000000000063000000000000,3,0,3,0,4,43,26,1000,29500,2,0,853,2000,2,0,1.00,0,0,0,124,'ABYSSEA');
 INSERT INTO `spell_list` VALUES (699,'barbed_crescent',0x00000000000000000000000000000063000000000000,3,0,0,0,4,43,52,500,22000,2,0,854,2000,0,0,1.00,0,0,0,58,'ABYSSEA');
 -- INSERT INTO `spell_list` VALUES (700,'natures_meditation',0x00000000000000000000000000000063000000000000,3,1,0,1,43,38,500,60000,230,0,907,2000,6,0,1.00,0,0,0,0,'ABYSSEA');
 -- INSERT INTO `spell_list` VALUES (701,'tempestuous_upheaval',0x00000000000000000000000000000063000000000000,3,4,0,4,43,133,500,7000,2,0,908,4000,1,0,1.00,0,0,0,124,'ABYSSEA');
```
ibm2431, line 716: Elemental IDs for currently unimplemented spells were _not_ adjusted. Such line comments are _already_ out of date and are missing spell_family definitions.

sql/weapon_skills.sql
```diff
@@ -80,8 +80,8 @@ INSERT INTO `weapon_skills` VALUES (29,'pyrrhic_kleos',0x00000000000000000000000
 INSERT INTO `weapon_skills` VALUES (30,'aeolian_edge',0x00000000020200000002020002000000000002000000,2,290,16,45,2000,5,2,8,4,6,0,0);
 INSERT INTO `weapon_skills` VALUES (31,'rudras_storm',0x00000000000100000001000000000000000001000000,2,0,0,236,2000,5,1,14,10,0,1,36);
 INSERT INTO `weapon_skills` VALUES (32,'fast_blade',0x02000000020202020202020202020002020002000002,3,5,0,1,2000,5,1,4,0,0,0,0);
-INSERT INTO `weapon_skills` VALUES (33,'burning_blade',0x02000000020202020202020202020002020002000002,3,30,2,2,2000,5,1,3,0,0,0,0);
-INSERT INTO `weapon_skills` VALUES (34,'red_lotus_blade',0x02000000020002020000000000000002000000000002,3,50,2,3,2000,5,1,3,6,0,0,0);
+INSERT INTO `weapon_skills` VALUES (33,'burning_blade',0x02000000020202020202020202020002020002000002,3,30,1,2,2000,5,1,3,0,0,0,0);
+INSERT INTO `weapon_skills` VALUES (34,'red_lotus_blade',0x02000000020002020000000000000002000000000002,3,50,1,3,2000,5,1,3,6,0,0,0);
```
ibm2431, line 84: I believe I looked into this, and I wasn't able to find anywhere that actually used these elements which were being set. I think this can be removed later.

src/common/mmo.h
```diff
@@ -70,18 +70,6 @@ enum NFLAGTYPE : uint32
...
-    FIRE		= 0x01,
-    ICE			= 0x02,
-    WIND		= 0x04,
-    EARTH		= 0x08,
-    THUNDER		= 0x10,
-    WATER		= 0x20,
-	LIGHT		= 0x40,
-	DARK		= 0x80,
-};
-
```
ibm2431, line 84: In the ultimate irony, there has _always_ been a definition for Elements which go in the _correct_ order. _But it was never once used anywhere!_

src/map/entities/charentity.cpp
```diff
@@ -1746,7 +1746,7 @@ void CCharEntity::UpdateMoghancement()
                 CItemFurnishing* PFurniture = static_cast<CItemFurnishing*>(PItem);
                 if (PFurniture->isInstalled())
                 {
-                    elements[PFurniture->getElement()] += PFurniture->getAura();
+                    elements[PFurniture->getElement() - 1] += PFurniture->getAura();
```
ibm2431, line 1751: Elements for furnishings are now base 1, but the `elements` table defined up above (unchanged, you'll have to expand up) is, of course, base zero. You'll see this a couple places on the core side.

src/map/lua/luautils.cpp
```diff
@@ -675,13 +692,16 @@ namespace luautils
...
+    *   Note: THIS IS NOT THE SAME AS THE DAY OF THE WEEK                   *
+    *   Days of week: Fire Earth Water Wind Ice Lightning Light Dark        *
+    *   Elements: Fire Ice Wind Earth Thunder Water Light Dark              *
     *                                                                       *
     ************************************************************************/
 
     int32 VanadielDayElement(lua_State* L)
     {
-        lua_pushinteger(L, CVanaTime::getInstance()->getWeekday());
+        lua_pushinteger(L, battleutils::GetDayElement());
```
ibm2431, line 708: I went back and forth on this. I originally considered having elements be part of the time object, but that didn't feel entirely right:
- Elements are really only referenced in the context of combat (spell elements)
- We currently have GetWeather in `battleutils`, I wanted GetDayElement to be in there with it, and I was hesitant to create yet another header/util for "common world" functions (too much rocking the boat for one PR)

src/map/utils/battleutils.cpp
```diff
@@ -886,53 +886,13 @@ namespace battleutils
         if (PAttacker->getMod(Mod::ENSPELL) > 0 && (PAttacker->getMod(Mod::ENSPELL_CHANCE) == 0 ||
             PAttacker->getMod(Mod::ENSPELL_CHANCE) > tpzrand::GetRandomNumber(100)))
         {
-            SUBEFFECT subeffects[8] = { SUBEFFECT_LIGHT_DAMAGE, SUBEFFECT_DARKNESS_DAMAGE, SUBEFFECT_FIRE_DAMAGE, SUBEFFECT_EARTH_DAMAGE,
-                SUBEFFECT_WATER_DAMAGE, SUBEFFECT_WIND_DAMAGE, SUBEFFECT_ICE_DAMAGE, SUBEFFECT_LIGHTNING_DAMAGE };
```
ibm2431, line 893: Enspells required some restructuring, as the subeffect array had a rather confusing order with listed Light/Dark _first_ (I'm not certain why this was done) and then the normal elements (in day-of-week order).

src/map/utils/battleutils.h
```diff
@@ -44,22 +44,22 @@ enum ENSPELL
...
+    ENSPELL_II_WIND = 11,
+    ENSPELL_II_EARTH = 12,
+    ENSPELL_II_THUNDER = 13,
+    ENSPELL_II_WATER = 14,
     ENSPELL_II_LIGHT = 15,
-    ENSPELL_BLOOD_WEAPON = 16,
-    ENSPELL_ROLLING_THUNDER = 17,
+    ENSPELL_II_DARK = 16,
+    ENSPELL_BLOOD_WEAPON = 17,
     ENSPELL_AUSPICE = 18,
```
ibm2431, line 63: [Endark II](https://www.bg-wiki.com/bg/Endark_II) wasn't in this table.

"Enspell Rolling Thunder" is never used - the [ability simply uses Enthunder instead](https://github.com/ibm2431/topaz/blob/master/scripts/globals/abilities/pets/rolling_thunder.lua).

So Blood Weapon's "enspell ID" got pushed back by one.

src/map/utils/zoneutils.cpp
```diff
@@ -984,6 +984,9 @@ int GetWeatherElement(WEATHER weather)
 {
     TPZ_DEBUG_BREAK_IF(weather >= MAX_WEATHER_ID);
 
+    // TODO: Fix weather ordering; at the moment, this current fire, water, earth, wind, snow, thunder
+    // order MUST be preserved due to the weather enums going in this order. Those enums will
+    // most likely have rippling effects, such as how weather data is stored in the db
```
ibm2431, line 989: I didn't change the ordering in this array due to the fact that weather is currently in an... odd... order. You might have to expand down to see it. This is in both core [and lua](https://github.com/ibm2431/topaz/blob/master/scripts/globals/weather.lua#L6-L23).

This may be of interest to @Kreidos

src/map/vana_time.cpp
```diff
@@ -202,13 +202,13 @@ TIMETYPE CVanaTime::SyncTime()
...
         switch (m_vHour)
         {
-            case  0: m_TimeType = TIME_NIGHT;   return TIME_MIDNIGHT;
-            case  4: m_TimeType = TIME_NEWDAY;  return TIME_NEWDAY;
-            case  6: m_TimeType = TIME_DAWN;    return TIME_DAWN;
-            case  7: m_TimeType = TIME_DAY;     return TIME_DAY;
-            case 17: m_TimeType = TIME_DUSK;    return TIME_DUSK;
-            case 18: m_TimeType = TIME_EVENING; return TIME_EVENING;
-            case 20: m_TimeType = TIME_NIGHT;   return TIME_NIGHT;
+            case  0: m_TimeType = TIME_MIDNIGHT; return TIME_MIDNIGHT;
```
ibm2431, line 213: While I was investigating where to put `GetDayElement` (ie: "here in the time object?") I came across:
`case  0: m_TimeType = TIME_NIGHT`, and _assumed_ it should be `TIME_MIDNIGHT`. Other line diffs were just spacing since `MIDNIGHT` is longer than `EVENING`.

Yell at me if my assumption was wrong, @Kreidos

Kreidos: Whoops! That one wasn't me, but I did miss it when I did the TOTD fix. Good catch.

Edit: At least the returned value was correct, it looks like only the Lua was dependent on m_TimeType, and nothing is using it right now.

sql/spell_list.sql
Noikar: Okay I'll keep an eye out for this.

ibm2431:
61e492c is me finalizing a retreat from reordering skillchains. While dealing with elements, I found that they didn't match retail's order (see message IDs), and began "resorting them while I'm here".

When I realized the full depth of the reordering, I decided to shelve it and retreat from capsizing both myself and reviewers. I apparently missed a bit during my retreat.

src/map/utils/zoneutils.cpp
Kreidos: The WEATHER enum stores the actual packet/client values for weather so the order there is accurate and should probably stay unchanged. As well, nothing that changes here should affect weather in-game or in the DB as those only rely on the WEATHER enum being correct. :)

Kreidos: The WEATHER enum stores the actual packet/client values for weather so the order there is accurate and should probably stay unchanged. As well, nothing that changes here should affect weather in-game or in the DB as those only rely on the WEATHER enum being correct. :)

ibm2431: Huh. I guess [I should have checked](https://github.com/Windower/Resources/blob/master/resources_data/weather.lua#L3-L23). 

Come on SE, why do you do such things to us. 😢 

I'll amend the comment - and maybe place a few others in weather enums - so others don't make the same assumption I did.

Thanks, Kreidos~

sql/weapon_skills.sql
TeoTwawki: I think there was a plan top use them, before they got put into script instead. So its kinda redundant.

scripts/zones/Tavnazian_Safehold/npcs/Yurim.lua
TeoTwawki: short version: CS event injection is only a concern for _parameters_ and not the csid itself.

scripts/globals/mobskills/crystal_weapon.lua
TeoTwawki: >  there is no "Nonday".

irl we call it Monday. Monday tries to return us all to non existence by sucking. Vanadiel doesn't know how lucky it is to not have Monday.

scripts/globals/abilities/elemental_siphon.lua
TeoTwawki: if pet ID is arbitrary and none are capped (I thought the early ones were and then at higher IDs we through packet info out the window) we could shift them up by one having fire start at 1 and then they'd align without addition.

TeoTwawki: ..in a future pr because who wants to fix the other pr's depending on the pet ids, right?

sql/spell_list.sql
TeoTwawki: thanks for the warning. I have a bunch of replace rows (I was too lazy for update statements, so I copypasted then edited a column value) to mod enmity. I'll need to update my rows.

PR #832, opened 2020-07-10
Fix issue with not referencing text IDs

ShelbyZ:
Looks like a set of overlooked changes when tabling occurred

**_I affirm:_**


ibm2431 (approved):
Thanks, Shelby!

59blargedy (approved):

PR #833, opened 2020-07-10
Trusts - Consider Trusts for Samba Dazes

zach2good:
Now that I've had some sleep, see updated comment below

**_I affirm:_**

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -991,8 +991,7 @@ namespace battleutils
             // Generic drain for anyone able to do melee damage to a dazed target
             // TODO: ignore dazes from dancers outside party
             int16 delay = PAttacker->GetWeaponDelay(false) / 10;
-
-            if (PAttacker->PMaster == nullptr)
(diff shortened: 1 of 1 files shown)


kaincenteno:
awesome, samba needed some debuffs :dancers:

zach2good:
So, it was reported that trusts weren't getting any drain from a monster that was under the effect of drain daze (from drain samba). 

I dug in, and there is a section that skips over anything that has a master: (`PAttacker->PMaster == nullptr`). This was fine, according to the wiki:

```
Characters outside of the Dancer's party will not benefit from this effect, nor will non-party members of the Dancer's alliance. Likewise, Puppetmaster automatons, Dragoon wyverns, Beastmaster pets (Jugs or Natural), and NPC fellows will not benefit from a Drain Daze effect.
```

But I went and tested on retail and Trusts _DO_ get the benefit from drain daze.

This PR cleans up the relevant section in core by extracting the daze checking into its own lambda and then checking the relevant configurations of party/master's party/self for the samba's presence.

To clarify, this is just to allow Trusts to get samba effects, no balancing.

cocosolos (approved):

PR #842, opened 2020-07-13
Adding more rpi flags and updating cmake

ShelbyZ:
**_I affirm:_**


ibm2431 (approved):
Whoops, I must have missed the truncation flag for cmake!

cocosolos (approved):

cocosolos:
CMake wasn't giving me any problems on Debian, only G++-8, but it's probably best to have them both match.

PR #851, opened 2020-07-15
[Blue Mage] Fix assimilation merit

Xaver-DaRed:
Assimilation Blue Mage merit now works. It properly raises the amount of blue points per merit assigned.
From my testing, it doesnt seem to break anything.

Credits to Devi for decompiling the game and to Caelic for the actual code.
Permission to PR this was granted by Caelic.

**_I affirm:_**

### DIFF
--- a/src/map/packets/menu_merit.cpp
+++ b/src/map/packets/menu_merit.cpp
@@ -36,27 +36,18 @@ CMenuMeritPacket::CMenuMeritPacket(CCharEntity* PChar)
 	ref<uint8>(0x06) = 0x0C;
 
     ref<uint16>(0x08) = PChar->PMeritPoints->GetLimitPoints();
-    ref<uint8>(0x0A) = PChar->PMeritPoints->GetMeritPoints();
 
-	uint8 flag = 0x00;
+    uint16 bits = 0;
+    bits += PChar->PMeritPoints->GetMeritPoints() & 0b0000000001111111; // first seven bits are merit points
+    if (PChar->GetMJob() == JOB_BLU && PChar->GetMLevel() > 74)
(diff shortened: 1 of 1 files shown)


src/map/packets/menu_merit.cpp
```diff
@@ -36,27 +36,18 @@ CMenuMeritPacket::CMenuMeritPacket(CCharEntity* PChar)
 	ref<uint8>(0x06) = 0x0C;
 
     ref<uint16>(0x08) = PChar->PMeritPoints->GetLimitPoints();
-    ref<uint8>(0x0A) = PChar->PMeritPoints->GetMeritPoints();
 
-	uint8 flag = 0x00;
+    uint16 bits = 0;
+    bits += PChar->PMeritPoints->GetMeritPoints() & 0b0000000001111111; // first seven bits are merit points
+    if (PChar->GetMJob() == JOB_BLU && PChar->GetMLevel() > 74)
+        bits += (PChar->PMeritPoints->GetMeritValue(MERIT_ASSIMILATION, PChar) << 7) & 0b0001111110000000; // next six bits are assimilation points. the last three bits are the three flags below:
```
ibm2431, line 43: Does retail only set these bits when the character is on BLU, or is the merit point value _always_ sent but just not usable?

Xaver-DaRed: asimilation is a job specific merit, they can only be set while the main job is blu

ibm2431: Whoever wants to investigate, see the capture "Wings of the Goddess: WOTG 10 (Part 1)".

At 06:12:44 (system time - always viewable in the top-right corner of a character's screen) she enters a Mog House and changes from WHM to BLU. This is at around 28:52 on the video.

Here's the excerpt of her incoming 0x063 packet log around that time:
[WHM_to_BLU_0x063.log](https://github.com/project-topaz/topaz/files/4974746/WHM_to_BLU_0x063.log)

PR #854, opened 2020-07-16
Bestiary QC

wrenffxi:
**_I affirm:_**

As always, big thanks to Nyu for capturing this data from retail zones.

This fixes (or creates new, where needed) mob pool data for mobs belonging to following categories of the [bestiary](https://ffxiclopedia.fandom.com/wiki/Category:Bestiary).  I've broken the commits into families for easier reviewing.

Completed Categories:
* Beasts
* Lizards
* Vermin

Going to stop adding to this PR now, to allow for review before it starts conflicting with other PRs.


Wiggo32 (approved):

sql/mob_groups.sql
```diff
@@ -13519,7 +13513,7 @@ INSERT INTO `mob_groups` VALUES (28,5526,272,'Slinking_Slug',0,0,0,0,0,121,123,0
 INSERT INTO `mob_groups` VALUES (29,5527,272,'Bloodspattered_Fly',0,0,0,0,0,120,123,0);
 INSERT INTO `mob_groups` VALUES (30,2671,272,'Mineral_Eater',0,0,0,0,9999,120,123,0);
 INSERT INTO `mob_groups` VALUES (31,3102,272,'Passage_Crab',0,0,0,0,0,120,123,0);
-INSERT INTO `mob_groups` VALUES (32,4980,272,'Unyielding_Tarichuk',0,0,0,0,0,120,123,0);
+INSERT INTO `mob_groups` VALUES (32,5639,272,'Irascible_Tarichuk',0,0,0,0,0,120,123,0);
```
ibm2431, line 13516: Others: If you see a name change in this diff, it's because there's currently two groups with the same group name in a zone. Wren is updating the name of the duplicate group to match the mobs that are already in that group.

GitHub might refuse to properly expand surrounding lines, but in zone 272, there were two batches of Tarichuk: group 14 and group 32. The tarichuks in the zone are already divided into these separate groups, Group 32 in mob_spawn_points are already named Irascible Tarichuk. So Wren's updating the name in mob_groups to match the mobs belonging to it, and giving them a new mob_pool.

sql/mob_pools.sql
```diff
@@ -4796,24 +4796,24 @@ INSERT INTO `mob_pools` VALUES (4735,'Smierc','Smierc',227,0x0000340800000000000
...
 INSERT INTO `mob_pools` VALUES (4747,'Frondescent_Leafkin','Frondescent_Leafkin',455,0x0000E20900000000000000000000000000000000,1,1,8,240,100,0,1,0,0,0,0,0,1,129,0,0,0,1,0,455);
 INSERT INTO `mob_pools` VALUES (4748,'Shy_Heartwing','Shy_Heartwing',343,0x00000F0A00000000000000000000000000000000,4,1,11,240,100,0,0,0,0,0,0,0,1,1153,0,0,2,1,0,343);
-INSERT INTO `mob_pools` VALUES (4749,'Lavender_Twitherym','Lavender_Twitherym',338,0x0000E70943617265656E69675477697468000000,1,1,12,240,100,0,1,0,1,0,0,0,7,131,0,0,0,1,0,338);
-INSERT INTO `mob_pools` VALUES (4750,'Floodplain_Spider','Floodplain_Spider',235,0x00003801436F6C7373616C537069657200000000,1,1,7,240,100,0,1,0,1,0,0,0,7,643,0,0,0,1,0,236);
+INSERT INTO `mob_pools` VALUES (4749,'Lavender_Twitherym','Lavender_Twitherym',338,0x0000E70900000000000000000000000000000000,1,1,12,240,100,0,1,0,1,0,0,64,7,131,0,0,0,1,0,338);
+INSERT INTO `mob_pools` VALUES (4750,'Floodplain_Spider','Floodplain_Spider',235,0x0000380100000000000000000000000000000000,1,1,7,240,100,0,1,0,1,0,0,0,7,643,8,0,0,1,0,236);
```
ibm2431, line 4810: I'm intrigued how the previous model fields for these twitherym had random junk model data, when as "simple mobs" they should have all been "simple model" types.

ibm2431 (approved):
Trusting you on your pool IDs matching up~

sql/mob_groups.sql
wrenffxi: The reason this happened, btw, is because when the name column was added to the groups table, it was infilled from the pools table for whatever poolid was currently in the row, rather than from the mob_spawn_points table.

sql/mob_pools.sql
TeoTwawki: SE doesn't zero the unused bytes, so junk data gets caught in caps sometimes depending on capture methods (if you wireshark it you will get all sorts of nonsense)

ibm2431: Huh, TIL.

PR #855, opened 2020-07-16
Fix item distribution from Festive Moogles

cocosolos:
**_I affirm:_**


ibm2431 (commented):
Does order of the items matter to the CS trigger? getFestiveItems is inserting them in an arbitrary order. It's also unpacking the returned unpack table.

cocosolos:
I don't think the order matters. I tested with Mooge Rod and it all worked.

Wiggo32 (approved):

cocosolos:
Fixed the unpacking stuff and rebased to fix any conflicts with whitespace changes.

PR #857, opened 2020-07-17
[Bard] Fix TP loss on instrument swap

Xaver-DaRed:
In retail, you don't loose TP when changing instruments.
Also, style edited some commented out code, becouse it was there.

Why would you even have TP in the first place? That is of no matter.

**_I affirm:_**


ibm2431 (approved):

cocosolos:
https://www.ffxiah.com/forum/topic/46553/military-parade-a-dd-bard-guide/8/
>You only lose TP if you entirely unequip one of those two slots, or equip into one of them when both are empty, but switching between them is fine. And equipping ranged weapons like Rune Bow loses TP of course.

Does this handle this?

Xaver-DaRed:
@cocosolos 
Indeed. TP loss is only prevented when swaping from one instrument (wind or string) to another (again, be it wind or string)
TP loss does happen when unequiping instruments

PR #859, opened 2020-07-17
Remove ability offsets from trusts

ibm2431:
Prevent Ayame from using Reward.

Probably to Naji's dismay.

**_I affirm:_**
- [🤞 ] that I've _tested my code_ since the last commit in the PR, and will test after any later commits


ibm2431:
I'm not _supposed_ to do this, but I'm not gonna ping someone for a review~

PR #863, opened 2020-07-18
Job abilities enum-ed

Omnione:
Replaced cor die script magic numbers with the corresponding enums

Added every job ability missing from the enum in ability.lua,
these are just enums so they may or may not be in the sql yet.

**_I affirm:_**


Omnione:
gonna add to this and add every ability in the ability.h also

ibm2431 (approved):
Thanks, Omni! You are now at the top of my favorite contributors list!

Wiggo32 (approved):

PR #870, opened 2020-07-19
RoE System + Sparkshop + Early Records

Kreidos:
From the team that brought you [Magian](https://github.com/project-topaz/topaz/pull/802) and "[That other feature nobody cares about](https://github.com/project-topaz/topaz/pull/645)", we present Records of Eminence. This is a complete base from which the rest of the records may be built. Thanks and credit goes to @Apocarypse for his packet captures, deciphering assistance, supplying the Spark shop and the NPC/Lua helpers. 

### What's in here?
- All UI elements and interactions.
- Taking/Dropping/Progressing/Completing records
- The Spark Shop + Eminence NPC scripts / CS.
- Persistence of all eminence data per player
- C++ / Lua hooks for interacting with the Eminence system
- Early records fully complete-able
- Migrations for existing DBs

### Which records work?
All Topaz-available content from `Tutorial->Basics`, plus one repeatable "Vanquish Multiple Enemies I".
> First Step Forward
Undertake a FoV Training Regime
Heal Without Using Magic
Vanquish One Enemy
Undertake a GoV Training Regime
Vanquish Multiple Enemies I

### What's comes next?
This base is essentially, wholly complete and these records serve as an example of how the system works, as well as an appetizer for players to sink their teeth into. Next steps would be getting the Records implemented. The vast majority of records could be fully realized using only the hooks provided here, but it would be messy. The next course will be to consider a paradigm for how the nearly 1300+ records will be checked. Whether those checks will be in Core or Lua, and what shape such a system would take. Discussion on this would be welcomed in Discord channels, but is technically out of the scope of this PR.

#### Considerations:
- This adds a new column to the `chars` table for eminence data. New sql file + migration python script is provided. The migration **should** be run on all existing databases. 
- The 6 records which are done here should be considered example *teaser* records only. As detailed above, a more comprehensive handling system for records would be the next phase. There are 1300+ records, and there's no expectation that they'd all be handled like the examples here. :)
- All standard records can be taken by a player, even if no logic exists to progress them. In part due to the current lack of a comprehensive check system as well as in the interests of testing. The client controls which records are shown, and there's no way to hide unimplemented records from the server end; please refer to the above list of working records.

**_I affirm:_**


TeoTwawki (approved):
Looks ~~good~~ _freaking awesome_ to me. Only thing sticks out to me is copyright notice

Kreidos:
> Looks ~good~ _freaking awesome_ to me. Only thing sticks out to me is copyright notice

100%. They were copy-pasted without realizing it. Staff will advise regarding their replacement, I'd prefer the new source files go under AGPL if possible but either way that will be changed. :)

Xaver-DaRed:
other than tabling the rewards as we discussed, this looks awesome

Kreidos:
> other than tabling the rewards as we discussed, this looks awesome

Thanks Xaver! For sure. As we discussed, the 6 records here should be considered as functioning examples only since a proper record-handling system would be phase 2.

TeoTwawki:
there are a few things we can optimize here later, but everything is gewd :+1:

scripts/globals/npc_util.lua
```diff
@@ -396,6 +398,53 @@ function npcUtil.completeQuest(player, area, quest, params)
...
+            player:addCurrency('spark_of_eminence', params["sparks"] * bonus * SPARKS_RATE)
+            player:messageBasic(tpz.msg.basic.ROE_FIRST_TIME_SPARKS, params["sparks"] * bonus * SPARKS_RATE, player:getCurrency("spark_of_eminence"))
+        end
+    end
+
+    if params["xp"] ~= nil and type(params["xp"]) == "number" then
+        player:addExp(params["xp"] * EXP_RATE)
+    end
+
+    if params["item"] ~= nil then
```
wrenffxi: This check for whether the player can be given the item must be the first thing done in this function.  Otherwise, with full inventory space, the player could keep attempting to complete, reaping the currency and xp rewards, but never getting the KI or completion.

scripts/globals/sparkshop.lua
```diff
@@ -0,0 +1,602 @@
...
+-- TO DO: add "Exchange A.M.A.N. currencies" options and event trusts
+---------------------------------------------------------------------
+ require("scripts/globals/npc_util")
+ require("scripts/globals/zone")
+----------------------------------------------------------------------
+
+tpz = tpz or {}
+tpz.sparkshop = tpz.sparkshop or {}
+
+optionToItem = {
```
wrenffxi: local optionToItem, if it's only used in this one file

scripts/globals/npc_util.lua
Kreidos: This bug actually exists already in `completeQuest()` as well. Someone might want to PR a fix for that separately too as it would also be exploitable.

scripts/zones/Southern_San_dOria/npcs/Rolandienne.lua
```diff
@@ -0,0 +1,39 @@
...
+        tpz.sparkshop.onTrigger(player,npc,995)
+    end
+end
+
+function onEventUpdate(player,csid,option)
+    tpz.sparkshop.onEventUpdate(player,csid,option)
+end
+
+function onEventFinish(player,csid,option)
+    if csid == 24 and option == 1 then
```
cocosolos: Looks like this is wrong csid, same with Eternal Flame and Fhelm Jobeizat.

Apocarypse: woops, i copy and pasted those in from isakoth after changing a bunch of stuff. i'll fix it

src/map/packets/roe_questlog.cpp
```diff
@@ -0,0 +1,35 @@
+﻿/*
+===========================================================================
+
+  Copyright (c) 2020 - Kreidos | github.com/kreidos
```
zach2good, line 4: I like individual attribution, but does this have any possible ramifications down the road?
If you had a particularly aggressive employer who tried to claim all your spare time work etc.

src/map/packets/roe_questlog.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "../../common/socket.h"
+
+#include "roe_questlog.h"
+
+#include "../entities/charentity.h"
+
+CRoeQuestLogPacket::CRoeQuestLogPacket(CCharEntity* PChar, uint8 order)
+{
+	this->id(0x112);
```
zach2good: Couple of copy/pasted tabs in hiding in these id/size sections

src/map/packets/roe_questlog.h
```diff
@@ -0,0 +1,44 @@
...
+*/
+
+#ifndef _CROEQUESTLOGPACKET_H
+#define _CROEQUESTLOGPACKET_H
+
+#include "../../common/cbasetypes.h"
+
+#include "basic.h"
+
+/************************************************************************
```
zach2good: If we aren't putting information in these block comments, do we need them?

src/map/packets/roe_update.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+CRoeUpdatePacket::CRoeUpdatePacket(CCharEntity* PChar)
+{
+	this->id(0x111);
+	this->length(0x104);
+
+	/*  SE did some very unusual packing here, likely due to the progressive additions made to RoE.
+	    Each 4-bit nibble in the 4-byte chunk is labeled here. The second number is it's position.
+	            (0 is the lowest order. IE the right-most bits)
+	            A1A0 B0A2 B2B1 B4B3  ||  A = Record ID B = Progress                                 */
```
zach2good, line 35: Love the in-depth explanations around all of this stuff!

src/map/utils/charutils.cpp
```diff
@@ -429,10 +433,15 @@ namespace charutils
...
             memcpy(&PChar->m_campaignLog, campaign, (length > sizeof(PChar->m_campaignLog) ? sizeof(PChar->m_campaignLog) : length));
 
-            PChar->SetPlayTime(Sql_GetUIntData(SqlHandle, 24));
-            PChar->profile.campaign_allegiance = (uint8)Sql_GetIntData(SqlHandle, 25);
-            PChar->setStyleLocked(Sql_GetIntData(SqlHandle, 26) == 1 ? true : false);
-            PChar->SetMoghancement(Sql_GetUIntData(SqlHandle, 27));
+            length = 0;
+            char* eminence = nullptr;
+            Sql_GetData(SqlHandle, 24, &eminence, &length);
+            memcpy(&PChar->m_eminenceLog, eminence, (length > sizeof(PChar->m_eminenceLog) ? sizeof(PChar->m_eminenceLog) : length));
```
zach2good, line 439: Thoughts on replacing memcpy and other C-style stuff with std::copy and the like?

src/map/utils/charutils.cpp
```diff
@@ -4918,13 +4947,130 @@ namespace charutils
...
+
+        if (strcmp(type, "spark_of_eminence") == 0)
+            PChar->pushPacket(new CRoeSparkUpdatePacket(PChar));
+    }
+
+    void SetEminenceRecordCompletion(CCharEntity* PChar, uint16 recordID, bool newStatus)
+    {
+        uint8 page = recordID / 8;
+        uint8 bit = recordID % 8;
+        if(newStatus)
```
zach2good, line 5027: Gap after if
But also, thoughts on Allman-style bracing, even for single-line statements?
```
while (x == y)
{
    something();
}
```

src/map/utils/charutils.cpp
```diff
@@ -4918,13 +4947,130 @@ namespace charutils
...
+            PChar->pushPacket(new CRoeQuestLogPacket(PChar, i));
+        }
+        charutils::SaveEminenceData(PChar);
+    }
+
+    bool GetEminenceRecordCompletion(CCharEntity* PChar, uint16 recordID)
+    {
+        uint8 page = recordID / 8;
+        uint8 bit = recordID % 8;
+        return PChar->m_eminenceLog.complete[page] & (1 << bit);
```
zach2good, line 5043: Love this black magic 🧙

src/map/utils/charutils.cpp
```diff
@@ -4918,13 +4947,130 @@ namespace charutils
...
+    bool HasEminenceRecord(CCharEntity* PChar, uint16 recordID)
+    {
+        for(int i = 0; i < 31; i++)
+        {
+            if(PChar->m_eminenceLog.active[i] == recordID)
+            {
+                return true;
+            }
+        }
+        return false;
```
zach2good: I haven't dug into the logic for this, and it isn't that brutal of a loop to optimise, but: 
Could active quests be put into a hashmap and retrieved via lookup instead of search?

src/map/lua/lua_baseentity.cpp
```diff
@@ -14420,6 +14532,10 @@ Lunar<CLuaBaseEntity>::Register_t CLuaBaseEntity::methods[] =
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,completeMission),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,setMissionLogEx),
     LUNAR_DECLARE_METHOD(CLuaBaseEntity,getMissionLogEx),
+    LUNAR_DECLARE_METHOD(CLuaBaseEntity,getEminenceCompleted),
+    LUNAR_DECLARE_METHOD(CLuaBaseEntity,setEminenceCompleted),
+    LUNAR_DECLARE_METHOD(CLuaBaseEntity,getEminenceProgress),
+    LUNAR_DECLARE_METHOD(CLuaBaseEntity,setEminenceProgress),
```
zach2good, line 14544: Great API, super clean ⭐

src/map/lua/lua_baseentity.cpp
```diff
@@ -6620,6 +6621,117 @@ inline int32 CLuaBaseEntity::getMissionLogEx(lua_State *L)
...
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 2) || !lua_isnumber(L, 2));
+
+    CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
+    uint16 recordID = (uint16)lua_tointeger(L, 1);
+    uint32 progress = (uint32)lua_tointeger(L, 2);
```
zach2good, line 6698: Do we have a project consensus about static_cast over C-Style casts?

src/map/lua/lua_baseentity.cpp
```diff
@@ -6620,6 +6621,117 @@ inline int32 CLuaBaseEntity::getMissionLogEx(lua_State *L)
...
+    }
+
+    return 1;
+}
+
+/************************************************************************
+*  Function: getEminenceProgress(record)
+*  Purpose :
+*  Example : player:getEminenceProgress(19)
+*  Notes   : returns nil if player does not have the record.
```
zach2good, line 6717: I've been in Core too long, on the Lua side, nil is equivalent to False right?
```lua
if not player:getEminenceProgress(19) then
...
end
```
is valid usage?

src/map/lua/lua_baseentity.cpp
```diff
@@ -6620,6 +6621,117 @@ inline int32 CLuaBaseEntity::getMissionLogEx(lua_State *L)
...
+    charutils::DelEminenceRecord(PChar, recordID);
+    charutils::SetEminenceRecordCompletion(PChar, recordID, status);
+
+    return 0;
+}
+
+/************************************************************************
+*  Function: getEminenceCompleted()
+*  Purpose : Returns true if eminence is flagged complete for player
+*  Example : player:getEminenceCompleted(1)
```
zach2good, line 6662: I mentioned this in Discord, and I think I saw someone else mention it, is this 1 going to get tabled later?
`player:getEminenceCompleted(1)`

scripts/zones/Bastok_Markets/npcs/A.M.A.N._Reclaimer.lua
```diff
@@ -0,0 +1,17 @@
...
+end;
+
+function onTrigger(player,npc)
+end;
+
+function onEventUpdate(player,csid,option)
+end;
+
+function onEventFinish(player,csid,option)
+end;
```
zach2good: Please kill these `;`'s

scripts/globals/sparkshop.lua
```diff
@@ -0,0 +1,602 @@
...
+        [40] = { cost = 2231, id = 12387 }, -- Koenig shield
+        [41] = { cost = 2755, id = 12385 }, -- Acheron shield
+        [42] = { cost =  300,  id = 16189 }, -- Gleaming shield
+    },
+    [10] = { -- Equipment (Lv.99)
+        [ 0] = { cost =  3000, id = 27740 }, -- Outrider mask
+        [ 1] = { cost =  5000, id = 27881 }, -- Outrider mail
+        [ 2] = { cost =  3000, id = 28029 }, -- Outrider mittens
+        [ 3] = { cost =  3000, id = 28168 }, -- Outrider hose
+        [ 4] = { cost =  3000, id = 28306 }, -- Outrider greaves
```
zach2good, line 492: Not everything in this list needs to be correctly implemented, but does everything have a vaguely OK sell price?

scripts/globals/regimes.lua
```diff
@@ -1202,6 +1202,16 @@ tpz.regime.bookOnEventFinish = function(player, option, regimeType)
 
             player:showText(player, msgOffset)
             player:showText(player, msgOffset + 1)
+
+            -- Records of Eminence: Undertake a FoV Training Regime
+            if player:getEminenceProgress(3) and regimeType == tpz.regime.type.FIELDS then
+                npcUtil.completeRecord(player, 3, { sparks = 100, xp = 500 })
```
zach2good, line 1222: This is such a clean API 👨‍🍳

migrations/eminence_blob.py
```diff
@@ -0,0 +1,22 @@
...
+
+    row = cur.fetchone()
+    if row:
+        return False
+
+    return True
+
+def migrate(cur, db):
+    cur.execute("ALTER TABLE `chars` ADD COLUMN eminence blob;")
+    db.commit()
```
zach2good: Line break plz

zach2good (commented):
I haven't gone in to test this yet but it looks like all bases are covered and laid out very cleanly, very nice, would review again.

src/map/packets/roe_questlog.cpp
Kreidos: I'm not sure. (Not a lawyer) I self-attributed based on input from Zircon, and staff. Apparently the FSF deems self-attributing as recommended practice.

src/map/utils/charutils.cpp
Kreidos: I am pro modern-style, in this instance since much of the code was mixing in with existing functions I wanted to maintain the same dialect since it was performing identical behaviour. I have long wanted to move areas like this to modern options, (ie. this here and especially config file handling .We're still using char*s and strcmp for crying out loud!)

src/map/lua/lua_baseentity.cpp
Kreidos: There wasn't a real consensus on it right now; there's going to be 1300 trials, and they'll mostly be referenced algorithmically once the check system goes in so I'm not sure how much value there is there. Hunts and Regimes don't use enums, and they work similarly too. At any rate, something to keep in mind for phase 2!

src/map/utils/charutils.cpp
Kreidos: +1 This is one area I've already optimized and will be pushing a commit for later. I want HasEminenceRecord to be basically-free since it will be called innumerable times once the check systems are written. In doing performance tests, a hashmap was actually about the same speed as this loop. (I assume because of the *hash* part having some overhead, and this loop being very short) I chose to implement a std::bitset of the player's active trials instead, which checking is about 6 times faster than this loop or a map.

scripts/zones/Bastok_Markets/npcs/A.M.A.N._Reclaimer.lua
Kreidos: Since it's empty and we aren't adding any of the reclaimers for the other zones either, I'm just going to kill the file; I'm sure @Apocarypse won't mind.

src/map/lua/lua_baseentity.cpp
Kreidos: Not too rusty, I see! nil and false both eval to False but 0 in Lua is True.

That example :100: . Intention was the API having as few hooks, that do the most possible.

src/map/packets/roe_questlog.h
Kreidos: Axe em'!

scripts/globals/regimes.lua
Kreidos: Thanks! And thanks for the review. Good points brought up, particularly about coding standards which I'm sure will come up more often as we increase compiler strictness and deal with additional GCC8/9 warnings.

src/map/utils/charutils.cpp
Kreidos: Also, as an aside, since these structs are necessarily comprised of only PoD types and sizeof() is used with no magic numbers I'm less bothered by memcpy use in these cases. Just as long as we don't use any complex types which can't handle shallow copies.

PS. SQL's library also does kind of lend itself more to this memcpy method of usage.

src/map/packets/roe_questlog.cpp
TeoTwawki: becomes problematic for future maintainers (I mean project maintainers) we don't want to maintain a list of everyone who ever touched one file on a file by file basis do we? what happens when I pr a change to "your" file here later. We're basically agreeing to topaz having this anyway. I like my attribution handled by the commit log personally, which the only legal ramifications being the license text itself which are there to ensure it stays free as in freedom.

image

Kreidos:
I have chosen to rebase due to conflicts introduced by `release` commits over the last month since I opened the PR. I've also changed the base back to `release` for the time being, since the roe branch is stale; this way diffs and reviews can still be made accurately.

59blargedy:
@project-topaz/developer @project-topaz/reviewer this one could use some review love post fixes.

PR #876, opened 2020-07-21
Fix dbtool issues

cocosolos:
Issue when mysql not in paths, and when writing a backup without doing an update first (no version hash).

**_I affirm:_**


ibm2431 (approved):
I've been getting undefined variable problems too. It seems the sub functions aren't pulling the variables defined in the "global" space.

PR #877, opened 2020-07-22
Trust Stability

zach2good:
**_I affirm:_**

Working on stability before forging ahead to finish gambits and positioning. There were rare crashes on zone after BCNM and I think the rate has shot up since messaging was added.

**Absolutely** looking for help with this

### DIFF
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -47,13 +47,13 @@ CTrustController::~CTrustController()
     }
     POwner->PAI->PathFind.reset();
     POwner->allegiance = ALLEGIANCE_PLAYER;
-    POwner->PMaster = nullptr;
-
     m_LastTopEnmity = nullptr;
 }
 
 void CTrustController::Despawn()
 {
+    POwner->PMaster = nullptr;
+    POwner->status = STATUS_DISAPPEAR;
     POwner->animation = ANIMATION_DESPAWN;
     CMobController::Despawn();
 }
@@ -62,7 +62,7 @@ void CTrustController::Tick(time_point tick)
 {
     m_Tick = tick;
 
-    if (POwner->PMaster == nullptr)
+    if (!POwner->PMaster)
     {
         return;
     }
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -47,13 +47,13 @@ CTrustController::~CTrustController()
     }
     POwner->PAI->PathFind.reset();
     POwner->allegiance = ALLEGIANCE_PLAYER;
+    POwner->status = STATUS_DISAPPEAR;
     m_LastTopEnmity = nullptr;
 }
 
 void CTrustController::Despawn()
 {
     POwner->PMaster = nullptr;
-    POwner->status = STATUS_DISAPPEAR;
     POwner->animation = ANIMATION_DESPAWN;
     CMobController::Despawn();
 }
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -12419,6 +12419,28 @@ inline int32 CLuaBaseEntity::addFullGambit(lua_State* L)
     return 0;
 }
 
+inline int32 CLuaBaseEntity::trustPartyMessage(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_TRUST);
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+
+    auto PTrust = static_cast<CTrustEntity*>(m_PBaseEntity);
+
+    auto p0 = lua_tointeger(L, 1);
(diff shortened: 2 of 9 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -12395,6 +12395,28 @@ inline int32 CLuaBaseEntity::addFullGambit(lua_State* L)
...
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+
+    auto PTrust = static_cast<CTrustEntity*>(m_PBaseEntity);
+
+    auto p0 = lua_tointeger(L, 1);
+    auto p1 = 0;
+    auto p2 = 711;
+
+    auto PMaster = static_cast<CCharEntity*>(PTrust->PMaster);
+    PMaster->ForParty([&](CBattleEntity* PMember)
```
zach2good: Either the cast to PMaster, or PParty of PMaster is broken here while zoning.

zach2good: In the old lua-based system, the calls to :getParty() would fail for the same reason

zach2good:
There is another much rarer crash reported by Canaria after BCNMs with Trusts, but there are many many more since messaging came in, so the messaging crash should be fixed first. 

If you're suffering with crashes since the messaging update, you can comment out the guts of messaging inside `trust.lua`: 
```lua
tpz.trust.message = function(mob, message_offset)
    --[[
         ...comment out everything in here...
     ]]--
end
```
and reload the file with `!reloadglobal trust` without a restart (or do a restart, same effect).

zach2good:
In this case, Trusts rely on their master to get a reference to the party that they're a member of. During zoning, a Char's party is set to `nullptr` until they've entered their destination zone and the reference is restored. I've tried wrapping up sensitive-looking areas with `if (PChar->loc.zoning) {...}` but it hasn't helped. 

**Braindump**
Could the PTrust's PMaster be nullptr? Is the lua running after the Trust is deleted in core?

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -47,13 +47,13 @@ CTrustController::~CTrustController()
...
     POwner->allegiance = ALLEGIANCE_PLAYER;
-    POwner->PMaster = nullptr;
-
+    POwner->status = STATUS_DISAPPEAR;
     m_LastTopEnmity = nullptr;
 }
 
 void CTrustController::Despawn()
 {
+    POwner->PMaster = nullptr;
```
zach2good, line 56: This seems to be a big source for instability around Trust lifetimes.
While the trust is despawning, disappearing and then eventually being deleted the reference to PMaster is alive right up until the Trust's destructor. There are many ticks happening in this window. Immediately setting it to nullptr (and a few paranoid checks later down the line) seem to have cleared up a lot of problems

src/map/lua/lua_baseentity.cpp
```diff
@@ -12419,6 +12419,31 @@ inline int32 CLuaBaseEntity::addFullGambit(lua_State* L)
...
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_TRUST);
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+
+    auto PTrust = static_cast<CTrustEntity*>(m_PBaseEntity);
+
+    auto p0 = lua_tointeger(L, 1);
+    auto p1 = 0;
+    auto p2 = 711;
```
ibm2431: Uncertain how I feel about the assignment here instead of using the ints directly in the creation of the new packet down below at:
```cpp
PCharMember->pushPacket(new CMessageCombatPacket(PTrust, PMember, p0, p1, p2));
```

Are p1 and p2 going to be capable of being changed later?

ibm2431 (approved):
I'll let you and/or the second reviewer be The Decider on the param assignment.

src/map/lua/lua_baseentity.cpp
zach2good: Yeah, they're magic numbers specific to this usage, so I'll move them down 👍

zach2good:
~~Hold off on merging, found a bug where ClearTrusts() misses one trust, allowing you to sneak trusts into BCNMs and other places they shouldn't be in. Good news though didn't get any crashes around zoning.~~

OK to merge, those bugs are already present in current trust branch, stability work is unaffected and can go ahead! Fixed the bug where trusts can sneak into your BCNM, the orphaned trusts bug is still there but hasn't caused any problems since trust went into canary, so is fine to leave for a while.

Wiggo32 (approved):

src/map/ai/controllers/trust_controller.cpp
```diff
@@ -62,7 +62,7 @@ void CTrustController::Tick(time_point tick)
 {
     m_Tick = tick;
 
-    if (POwner->PMaster == nullptr)
+    if (!POwner->PMaster)
```
ShelbyZ, line 65: Is this change just a matter of style more than implementation?

zach2good:
Just style

src/map/ai/controllers/trust_controller.cpp
zach2good: Just style

PR #886, opened 2020-07-25
remove underscores from mob_spawn_points.polutils_name

wrenffxi:
**_I affirm:_**


ibm2431 (approved):
These kinds of PRs are my favorite kind!

Wiggo32 (approved):

TeoTwawki:
Just a heads up, both name columns are completely unused _except_ for instanced entities like pets, trusts, and Battlefield helpers.

The original name column only existed for human readability, and the 2nd one was added not knowing that when instanced entities needed a way to name them that didn't alter the script name (since many of them have named that need sent in the packet that are impossible to use as a file name).

So if desired, we could get rid of the 1st one that has the underscores in the future.

PR #887, opened 2020-07-25
Bestiary QC: Plantoids

wrenffxi:
**_I affirm:_**

Included in this PR:

* Added missing apostrophes to mob_spawn_points.polutils_name.

* Changed Sauromugue_Skink and Thunderclaw_Thuban family from panopt to raptor.

* Realized that wiki was missing a few mobs from its bestiary category lists, so I revisited the categories from my previous PR (beasts, lizards, vermin), cross-checked each by family, found the missing mobs, and QCed them as well.  Going forward I'll use both wiki and family-crosscheck to find as many mobs as possible.

* QCed against retail members of the "Plantoids" category.  Put each species in its own commit for easier reviewing.


Wiggo32 (approved):

sql/mob_spawn_points.sql
```diff
@@ -8261,16 +8261,16 @@ INSERT INTO `mob_spawn_points` VALUES (16933038,'Flying_Spear','Flying Spear',36
...
 INSERT INTO `mob_spawn_points` VALUES (16933047,'Barometz','Barometz',38,459.392,-0.501,13.353,185);
 INSERT INTO `mob_spawn_points` VALUES (16933048,'Borametz','Borametz',39,437.342,-0.500,23.941,57);
 INSERT INTO `mob_spawn_points` VALUES (16933049,'Barometz','Barometz',38,469.541,-0.500,-23.776,18);
 INSERT INTO `mob_spawn_points` VALUES (16933050,'Borametz','Borametz',39,480.411,-0.501,37.913,153);
 INSERT INTO `mob_spawn_points` VALUES (16933051,'Barometz','Barometz',38,470.248,-0.500,21.134,126);
 INSERT INTO `mob_spawn_points` VALUES (16933052,'Borametz','Borametz',39,486.132,-0.501,34.731,54);
-INSERT INTO `mob_spawn_points` VALUES (16933055,'Barometz','Barometz',38,467.168,-0.500,11.415,63);
-INSERT INTO `mob_spawn_points` VALUES (16933056,'Borametz','Borametz',39,450.915,-0.500,31.892,32);
+INSERT INTO `mob_spawn_points` VALUES (16933055,'Barometz','Barometz',98,467.168,-0.500,11.415,63);
+INSERT INTO `mob_spawn_points` VALUES (16933056,'Borametz','Borametz',99,450.915,-0.500,31.892,32);
```
ibm2431, line 8273: Do 16933057 and 16933058 need the same change?

ibm2431 (approved):
Approved assuming that the second batch of Barometz are either fine as they are (or get a follow-up commit if they need adjusting)~!

sql/mob_spawn_points.sql
wrenffxi: So, in each of Barometz and Borametz, there is one mob with different flags (noted here by pool 98 and 99) than the others (noted here with pools 38 and 39).  In the retail capture I have from Nyu, the ID of the different mob is 16933055 for Barometz, and 16933056 for Borametz.

It might be that the "special" mob is chosen at random, each time you start a NE Apollyon run.  But unless that's the case, these values are what match the most recent retail cap.

Per wiki,
```
Note that there is one large black Barometz and one large white Borametz. Combined with Goobbue Harvester, one will open the Vortex, one will drop the item chest and one will do nothing.
```

The ones with the different flags are probably these "large" versions, which I'm guessing are static rather than randomly selected.  Code for opening the portal doesn't yet exist; Goobbue Harvester script isn't created yet, and none of the Bora/Baro-metz scripts open the portal.  But that change is outside the scope of this PR.

ibm2431: @cocosolos , @59blargedy ^ (we'll have to look into how the "special" mob works; random or static; re: flags)

Wren: Scripts work in Limbus branch, but nailing down the exact behavior to correct mob ID is indeed outside of scope of this PR.

cocosolos: The large mobs are probably supposed to be random, but at the time I couldn't figure out a clean way to adjust flags on spawn, so just went with static for now since their functions are still random.

PR #888, opened 2020-07-25
Activate boss QMs in Dynamis dreamland zones (inactive with lua)

jarmengaud:
**_I affirm:_**

### DIFF
--- a/scripts/zones/Dynamis-Buburimu/IDs.lua
+++ b/scripts/zones/Dynamis-Buburimu/IDs.lua
@@ -142,19 +142,19 @@ zones[tpz.zone.DYNAMIS_BUBURIMU] =
     {
         QM =
         {
-            [16941676] =
-            {
-                param = {3457, 3474, 3475, 3476, 3477, 3478},
-                trade =
-                {
-                    {item = 3457,                           mob = 16941057}, -- Apocalyptic Beast
-                    {item = {3474, 3475, 3476, 3477, 3478}, mob = 16941368}, -- Arch Apocalyptic Beast
-                }
-            },
-            [16941677] = {trade = {{item = 3463, mob = 16941552}}}, -- Lost Stihi
-            [16941678] = {trade = {{item = 3464, mob = 16941520}}}, -- Lost Barong
-            [16941679] = {trade = {{item = 3465, mob = 16941576}}}, -- Lost Alklha
-            [16941680] = {trade = {{item = 3466, mob = 16941666}}}, -- Lost Aitvaras
+            -- [16941676] =
+            -- {
+            --     param = {3457, 3474, 3475, 3476, 3477, 3478},
+            --     trade =
+            --     {
+            --         {item = 3457,                           mob = 16941057}, -- Apocalyptic Beast
+            --         {item = {3474, 3475, 3476, 3477, 3478}, mob = 16941368}, -- Arch Apocalyptic Beast
+            --     }
+            -- },
+            -- [16941677] = {trade = {{item = 3463, mob = 16941552}}}, -- Lost Stihi
+            -- [16941678] = {trade = {{item = 3464, mob = 16941520}}}, -- Lost Barong
+            -- [16941679] = {trade = {{item = 3465, mob = 16941576}}}, -- Lost Alklha
+            -- [16941680] = {trade = {{item = 3466, mob = 16941666}}}, -- Lost Aitvaras
         },
     },
 }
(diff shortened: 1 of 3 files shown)


ibm2431:
@wrenffxi  Are the bosses finished?

wrenffxi:
I've done no work on Dynamis NMs.  Checking them all for retail-accurate pools and battle mechanics is still on the TODO list #311

jarmengaud:
I was able to fight all 3, and they had their nasty TP moves... so they are not different from main zones bosses according to me.
They give the KI on kill, and drop their tome chapter. They don't drop anything else though I think.
But the idea is above all to access other zones...

Wiggo32 (dismissed):
this PR is just making the ??? npc's exist which will have no function without a lua script attached. it will need to be done anyway, thanks for making a future PR just a little bit smaller =)

jarmengaud:
Well that's not true, the code was there. With this change you can pop the 3 dreamland bosses and get their KI.

ibm2431:
Was asking @wrenffxi because of https://github.com/project-topaz/topaz/commit/4e418e22 ~!

wrenffxi:
The code to pop the the NMs exists, but the code -for- the NMs doesn't.  That's probably why we have the QMs disabled.

Let's look at Valkurm, the first in order.

Cirrate Christelle:
* Doesn't use Extremely Bad Breath, Fragfrant Breath, Miasmic Breath, or Putrid Breath.
* Mechanics are not coded for using Absorbent Moss, Redolent Root and Odorless Fungus on her.

Lost Fairy Ring:
* Does not have enhanced movement speed.

Lost Nant'ina:
* Does not uses Blow 3 times, then Uppercut 3 times, then Charm.
* Does not do 2000 damage Uppercuts or hit particularly harder than a normal mob of its type.

Lost Steamcroissant:
* I don't even know what special mechanics this NM has, but whatever they are, they aren't coded.

Arch Christelle:
* Missing the same breath attacks as Cirrate Christelle, plus they're supposed to be stronger versions.
* Its melee attacks are supposed to be conal breath attacks.

In short, all dreamworlds bosses, as coded, are push-overs compared to what they're supposed to be.  They're essentially just higher-level versions of basic mobs of their family.  On my old server, a group of 6-10 75-relic-era players would waltz up to Cirrate Christelle and smash her into the sand without first killing any of the minibosses to weaken her.

So by all means, server owners who want to allow progress through Dreamlands can enable the fights, and they "work" in that you can pop, win, get KI, and move on.  But I don't think we should enable them by default, until the battle mechanics for the NMs are coded.

*edit to add: a cleaner solution to prevent NM pops may be to comment out the trades in the zone's IDs.lua, similar to what we do with Abyssea, rather than by hiding the QM NPC*

jarmengaud:
Understood. Though I was hoping that something, temporary and good-enough, could be allowed to at least let people access all Dyna zones.
Dynamis is the oldest end-game content and we still can't access Beaucedine, Xarc & Tavnazia, so we can't get a relic set.
And it seems nobody is working on those

Unfortunately I am not able to make retail-accuracte versions of those NMs.
Speaking of retail server, all dyna but [Divergent] are solo zones and all those NMs are easy solo anyway for a 99 with trusts, whatever their mechanics.

TeoTwawki:
> edit to add: a cleaner solution to prevent NM pops may be to comment out the trades in the zone's IDs.lua, similar to what we do with Abyssea, rather than by hiding the QM NPC

@jarmengaud I'd like to suggest we re-purpose this PR into doing that. its a minimal addition to what you already have here.

jarmengaud:
k done. PR title is misleading now though.

TeoTwawki:
The title of the PR can be edited. there is a button up top of this webpage for it. Its still technically true: you are activating the NPC, but disabling its script from running.

Wiggo32 (approved):
I actually prefer all npc's to be initialized if they're supposed to be by default. And if they aren't working yet we just shouldnt have a script for them. (or comment out works too like how it is now) This is just my personal preference and I approve but, you all can decide what's better or more efficient cause that's not my strong point. I just like being able to see and check npc's (model, pos, etc.) at a glance and then not have to restart server to initialize them once I build a script.

TeoTwawki:
> I actually prefer all npc's to be initialized if they're supposed to be by default. And if they aren't working yet we just shouldnt have a script for them. (or comment out works too like how it is now) This is just my personal preference and I approve but, you all can decide what's better or more efficient cause that's not my strong point. I just like being able to see and check npc's (model, pos, etc.) at a glance and then not have to restart server to initialize them once I build a script.

My own personal pref is commented out with a note so people know why its not active yet, so we do not lead people into making prs to fix the npc while the mob still isn't fixed. Like happened to this contributor.

PR #895, opened 2020-07-27
Check for sharpshot frame on automaton ranged attack.

TheWhaler:
**_I affirm:_**

I noticed that on Tantalus the harlequin frame was doing ranged attacks even when it shouldn't at pretty much every tick that it wasn't doing something earlier in the process. Added in a frame check to the TryRangedAttack function.

### DIFF
--- a/src/map/ai/controllers/automaton_controller.cpp
+++ b/src/map/ai/controllers/automaton_controller.cpp
@@ -1333,11 +1333,10 @@ bool CAutomatonController::TryTPMove()
 
 bool CAutomatonController::TryRangedAttack() // TODO: Find the animation for its ranged attack
 {
-    if (PAutomaton->getFrame() == FRAME_SHARPSHOT) {
+    if (PAutomaton->getFrame() == FRAME_SHARPSHOT) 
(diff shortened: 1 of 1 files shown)


src/map/ai/controllers/automaton_controller.cpp
```diff
@@ -1333,8 +1333,11 @@ bool CAutomatonController::TryTPMove()
 
 bool CAutomatonController::TryRangedAttack() // TODO: Find the animation for its ranged attack
 {
-    if (m_rangedCooldown > 0s && m_Tick > m_LastRangedTime + (m_rangedCooldown - std::chrono::seconds(PAutomaton->getMod(Mod::SNAP_SHOT))))
-        return MobSkill(PTarget->targid, m_RangedAbility);
+    if (PAutomaton->getFrame() == FRAME_SHARPSHOT) {
+        if (m_rangedCooldown > 0s && m_Tick > m_LastRangedTime + (m_rangedCooldown - std::chrono::seconds(PAutomaton->getMod(Mod::SNAP_SHOT))))
+            return MobSkill(PTarget->targid, m_RangedAbility);
+        return false;
```
ibm2431: Since we're returning false outside of this parent if check, we can combine the frame check into the same check as the cooldown check, and remove the inner `return false`.

TheWhaler: Doh cs101 change made ;)

PR #896, opened 2020-07-28
Remove all Lua line-ending semicolons

ibm2431:
No more extra diffs or merge conflicts due to people removing line-ending semicolons in Lua scripts.

(Will probably conflict with a lot of recently-merged PRs or PRs which are still open. But I'd be on the hook for resolving conflicts in the merged PRs _anyway_.)

Two commits, the first is literally just:
replace-all

Results:
> 2988 changed files
> Total replaced: 29694

I will freely admit to not reviewing the diff - there shouldn't be need to. Anything that might break from this _deserves_ to.

The second commit was upon a test boot, the server complaining about [missing table element separators here](https://github.com/project-topaz/topaz/blob/c52431e2c48516954aa2f5ebbb331238f43cc3d8/scripts/globals/limbus.lua#L65-L75). I resolved those then did a findall for `};\n` on un-replaced release branch and manually looked through all results for any other instances:
table-seperate

That Limbus file was the only one that attempted to separate tables within tables with semicolon-newline separators. And the server boots without errors in the log (please ignore my outdated conf):
boots

It is technically possible for some script somewhere to have done the following
```lua
my_table = {
  thinga;
  thingb;
  thingc
}
```
And those tables will now be broken until spotted.

But to be honest, I'm not particularly concerned about such tables. They'd be old, obscure ones which _deserve_ to break for using a semicolon as an element separator. I'd rather _kill all the line-ending semicolons_ now in one fell swoop and then potentially fix up any obscure scripts later. **We've put up with these semicolons _long_ enough.**

**This particular PR will miss semicolons which occur as the final character in a file.** mrhappyasthma tried to move us closer to being able to doing something like this in #426, but I'm seeing instances after my mass-replace which were missed. For the sake of keeping a PR of this diff magnitude **as simple as possible (literally all I did was replace `;\n` with `\n` inside `/scripts/`)**, those last line-ending semicolon survivors will be addressed at a future date.

**_I affirm:_**


ibm2431:
For the curious, this was the last conflict I was willing to let those evil semicolons create:
conflict

scripts/zones/Aht_Urhgan_Whitegate/Zone.lua
```diff
@@ -3,225 +3,225 @@
...
+require("scripts/globals/keyitems")
+require("scripts/globals/missions")
+require("scripts/globals/quests")
+require("scripts/globals/status")
+require("scripts/globals/titles")
 require("scripts/globals/zone")
 -----------------------------------
 
 function onInitialize(zone)
     zone:registerRegion(1,57,-1,-70,62,1,-65); -- Sets Mark for "Got It All" Quest cutscene.
```
59blargedy: maybe a follow up regedit to catch ones with comments?

ibm2431: Nice eye, pushed~

ibm2431:
The above commit is just:
comment-regex

> Matching files: 528
> Total replaced: 2294

And removes line-ending semicolons which were followed by comments.

59blargedy (approved):

cocosolos (approved):
Spot checked, looks good. Any 5 second fixes that might pop up are worth it imo.

ibm2431:
Sorry @wrenffxi , this is going to merge conflict you on #879 and #889.

I hope you of all people will understand~ 😅

PR #898, opened 2020-07-28
Remove straggling semicolons at the end of lua scripts

ibm2431:
They thought they were safe because they were hiding before an EOF! They were wrong.

eof

> Total replaced: 3253
> Matching files: 3253

**_I affirm:_**


scripts/zones/Apollyon/Zone.lua
```diff
@@ -136,73 +136,73 @@ function onRegionEnter(player, region)
         -- ///////////////////////APOLLYON SE TELEPORTER///////////////////////////////////////////
         [20] = function (x)
              -- print("SE_telporter_f1_to_f2")
-             if (IsMobDead(16932992)==true and player:getAnimation()==0) then player:startEvent(219);end
+             if (IsMobDead(16932992)==true and player:getAnimation()==0) then player:startEvent(219); end
```
ibm2431, line 139: These were the result of a preliminary:
Find `;([^\s])`
Replace `; $1`

Only Limbus had scrunched semicolons on the same line as more characters

Wiggo32 (approved):

PR #899, opened 2020-07-28
Replace lua script tabs to spaces

ibm2431:
These are new tabs that have made their way in since #424.

We'll need to set up a linter at some point.

**_I affirm:_**


ibm2431:
This PR is mainly to show which scripts/PRs snuck tabs in.

I'm going to do a follow-up PR to tackle trailing whitespace in scripts. I'm _assuming_ that no other Staff take issue with my merging this~

PR #904, opened 2020-07-28
Adds drops to the Ancient Goobie

tankfest:
**_I affirm:_**

image

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -16541,6 +16541,10 @@ INSERT INTO `mob_droplist` VALUES (3138,0,0,1000,1875,1000); -- (Nergal, Bata, A
 INSERT INTO `mob_droplist` VALUES (3139,0,0,1000,1875,1000); -- (Apollyon Demon, Taurus, Ahriman) ancient_beastcoin x 1-2
 INSERT INTO `mob_droplist` VALUES (3139,0,0,1000,1875,500);  -- (Apollyon Demon, Taurus, Ahriman) ancient_beastcoin x 1-2
 INSERT INTO `mob_droplist` VALUES (3140,0,0,1000,2826,350);
+INSERT INTO `mob_droplist` VALUES (3141,0,0,1000,16990,50);  -- Ancient Goobbue; Daihannya
+INSERT INTO `mob_droplist` VALUES (3141,0,0,1000,919,1000);  -- Ancient Goobbue; Boyahda Moss
+INSERT INTO `mob_droplist` VALUES (3141,0,0,1000,1264,1000); -- Ancient Goobbue; Grt. Boyahda Moss
+INSERT INTO `mob_droplist` VALUES (3141,0,0,1000,1237,1000); -- Ancient Goobbue; Tree Cuttings
 
 /*!40000 ALTER TABLE `mob_droplist` ENABLE KEYS */;
 UNLOCK TABLES;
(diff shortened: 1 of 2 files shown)


sql/mob_droplist.sql
```diff
@@ -16541,6 +16541,10 @@ INSERT INTO `mob_droplist` VALUES (3138,0,0,1000,1875,1000); -- (Nergal, Bata, A
 INSERT INTO `mob_droplist` VALUES (3139,0,0,1000,1875,1000); -- (Apollyon Demon, Taurus, Ahriman) ancient_beastcoin x 1-2
 INSERT INTO `mob_droplist` VALUES (3139,0,0,1000,1875,500);  -- (Apollyon Demon, Taurus, Ahriman) ancient_beastcoin x 1-2
 INSERT INTO `mob_droplist` VALUES (3140,0,0,1000,2826,350);
+INSERT INTO `mob_droplist` VALUES (3141,0,0,1000,16990,40);
```
59blargedy: i think daihanya is more in line with 100 for rate
based on http://ffxidb.com/zones/153/ancient-goobbue
https://ffxiclopedia.fandom.com/wiki/Ancient_Goobbue
these are small sample sizes, however, but 40 seems lowish

ibm2431: 40 also isn't a valid TH0 bucket~! (50 would be if it were Rare, but I'm leaning towards Uncommon on this one due to the numbers you've posted)

sql/mob_droplist.sql
```diff
@@ -16541,6 +16541,10 @@ INSERT INTO `mob_droplist` VALUES (3138,0,0,1000,1875,1000); -- (Nergal, Bata, A
...
+INSERT INTO `mob_droplist` VALUES (3141,0,0,1000,919,1000);
+INSERT INTO `mob_droplist` VALUES (3141,0,0,1000,1264,1000);
+INSERT INTO `mob_droplist` VALUES (3141,0,0,1000,1237,1000);
```
ibm2431: [FFXIDB has these above 100% rate](http://ffxidb.com/zones/153/ancient-goobbue)

But old wiki is capping it at 100%. Anyone feel like killing Acient Goobbue on retail with full TH?

sql/mob_droplist.sql
tankfest: i didn't understand ffxidb, though i looked at it to base this off of.  TH0 = 10%, but TH2 = 8%?  I put 4% to make topaz's TH2 actually even out at around 10% like it says.  There is a report of someone going 0/27 with TH on it, comments  make it seem very low drop rate.

ibm2431: Sometimes FFXIDB's different TH values will conflict with each other; I tend to lean towards going with whatever is the higher sample size. (Can see sample sizes on FFXIDB by hovering over the table cell).

In either case, now that TH's exact mechanics are known, new drop percentages should match a designated TH bucket (_which_ bucket is debatable). At the moment our TH code doesn't function like it should, but that'll need to be fixed at the code level, not the data level.

At some point we'll need to go through our existing data and "snap" it to the nearest TH bucket. This'll be a prerequisite _before_ we can properly fix the TH code - we can't implement bucket behavior if we don't have buckets!

TeoTwawki: LIES! All lies I tell you! That katana is a myth! Doesn't exist!
*cries in terribad luck*

tankfest: would 50 suffice?

sql/mob_droplist.sql
TeoTwawki: That happens when it _can_ drop a 2nd (or 3rd or 4th) of an item, but at least 1 is 100% rate (the additional ones may or may not also be - not required).

So what the person checking this will need to do, is get enough kills to reliably say how many are 100% and how many total there generally could be.

sql/mob_droplist.sql
ibm2431: Yeah, 50 would align with Rare. Whether it _should_ be Rare is another matter.

As I don't play, I'm not too vested in which it is, so long as it's backed by good evidence.

59blargedy (approved):

PR #905, opened 2020-07-28
Stops steelshells in boyahda from linking

tankfest:
**_I affirm:_**

Don't see anywhere that these are supposed to link: https://ffxiclopedia.fandom.com/wiki/Steelshell


59blargedy (approved):

megberry (approved):

wrenffxi:
Good catch.

For everyone's clarity: my recent bestiary PRs are only checking against retail captures, so they may contain incorrect data in fields not captured, such as mob jobs, immunities, aggro, or linking.

When I find mobs sharing a pool, I use the existing pool to create one for the "new" mob, then edit the retail-captured fields on both.  For example, Fortalice Bats were using Funnel Bats pool.  I made a new pool for Fortalice Bats, that inherited all the non-capturable fields from Funnel Bats.

All this to say, that once I finish checking all bestiary families again retail, we'll still need to do another sweep through all the mobs, to QC against wikis all the non-captured fields.  I expect we'll find a lot of mistakes similar to this one that Tankfest is fixing.

ibm2431 (approved):

PR #906, opened 2020-07-28
Fixes status_effect flags for various consumes-effect-on-hit buffs

tankfest:
**_I affirm:_**

These abilities have improper flags, allowing them to be removed in various ways inappropriately:
image


sql/status_effects.sql
```diff
@@ -374,10 +374,10 @@ INSERT INTO `status_effects` VALUES (336,'misers_roll',131368,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (337,'companions_roll',131368,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (338,'avengers_roll',131368,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (339,'naturalists_roll',131368,0,0,0,0,0,0,0);
-INSERT INTO `status_effects` VALUES (340,'warriors_charge',37,0,0,0,0,0,0,0);
+INSERT INTO `status_effects` VALUES (340,'warriors_charge',36,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (341,'formless_strikes',32,0,0,0,0,0,0,0);
-INSERT INTO `status_effects` VALUES (342,'assassins_charge',37,0,0,0,0,0,0,0);
-INSERT INTO `status_effects` VALUES (343,'feint',35,0,0,0,0,0,0,0);
```
ibm2431, line 380: I like how no one noticed Feint was never wearing off over multiple attacks.

ibm2431 (dismissed):

sql/status_effects.sql
```diff
@@ -374,10 +374,10 @@ INSERT INTO `status_effects` VALUES (336,'misers_roll',131368,0,0,0,0,0,0,0);
...
+INSERT INTO `status_effects` VALUES (342,'assassins_charge',36,0,0,0,0,0,0,0);
+INSERT INTO `status_effects` VALUES (343,'feint',36,0,0,0,0,0,0,0);
```
ibm2431: Feint should wear on the [next _successful_ attack](https://ffxiclopedia.fandom.com/wiki/Feint). The behavior was implemented in https://github.com/DarkstarProject/darkstar/pull/5677, but the submitter must have assumed that the lose-on-attack effect flag was 2 instead of 4.

So Feint's effect flag should just be 32 (since I'm not finding evidence it can be dispelled)

ibm2431 (changes requested):

ibm2431 (approved):

Wiggo32 (approved):

PR #914, opened 2020-07-31
Guild Reset at JP midnight

### LINKED ISSUE #750: Cooking guild NPC does not accept bowl of puls for guild points
**_I have:_**

Had a player report that NPC wasnt accepting bowl of puls for guild points even though it was asking for that. It did work on the previous item that the guild asked for. 

Looked up the item and it should be a guild item that is accepted based on:

https://ffxiclopedia.fandom.com/wiki/Puls

### LINKED ISSUE #875: Daily reset time does not honor "JP Midnight" default setting
**_I have:_**

Things that are supposed to reset daily or weekly (tally, guild items,...), are not resetting at JP midnight with default configuration, even though by default:
TIMEZONE_OFFSET = 9.0 -- Offset from UTC used to determine when "JP Midnight" is for the server.  Default is JST (+9.0). 

If you look at the time_server.cpp code, you will see things like:
    //Midnight
    if (CVanaTime::getInstance()->getSysHour() == 0 && CVanaTime::getInstance()->getSysMinute() == 0)

so it is hardcoded that reset will happen at midnight for the server local time zone. And cannot be changed by a configuration either.

Would probably fix:
https://github.com/project-topaz/topaz/issues/22

jarmengaud:
**_I affirm:_**

(this is @ Tokenr on Discord, btw!)

Should solve:
Fixes https://github.com/project-topaz/topaz/issues/750
Fixes https://github.com/project-topaz/topaz/issues/875
and the first item of:
https://github.com/project-topaz/topaz/issues/22

The way the guild rank items are set up, the reset must happen on JP midnight, independant on server timezone.

### DIFF
--- a/src/map/time_server.cpp
+++ b/src/map/time_server.cpp
@@ -32,7 +32,6 @@
 #include "lua/luautils.h"
 #include "entities/charentity.h"
 #include "latent_effect_container.h"
-#include "daily_system.h"
 
 
 int32 time_server(time_point tick,CTaskMgr::CTask* PTask)
@@ -83,7 +82,6 @@ int32 time_server(time_point tick,CTaskMgr::CTask* PTask)
     {
         if (tick > (CVanaTime::getInstance()->lastMidnight + 1h))
         {
-            daily::UpdateDailyTallyPoints();
             CVanaTime::getInstance()->lastMidnight = tick;
         }
     }
--- a/src/map/time_server.cpp
+++ b/src/map/time_server.cpp
@@ -32,6 +32,7 @@
 #include "lua/luautils.h"
 #include "entities/charentity.h"
 #include "latent_effect_container.h"
+#include "daily_system.h"
 
 
 int32 time_server(time_point tick,CTaskMgr::CTask* PTask)
@@ -77,12 +78,12 @@ int32 time_server(time_point tick,CTaskMgr::CTask* PTask)
 
     }
 
-    //Midnight
+    // Midnight
     if (CVanaTime::getInstance()->getSysHour() == 0 && CVanaTime::getInstance()->getSysMinute() == 0)
     {
         if (tick > (CVanaTime::getInstance()->lastMidnight + 1h))
         {
-            guildutils::UpdateGuildPointsPattern();
+            daily::UpdateDailyTallyPoints();
             CVanaTime::getInstance()->lastMidnight = tick;
         }
     }
@@ -91,6 +92,7 @@ int32 time_server(time_point tick,CTaskMgr::CTask* PTask)
     {
         if (tick > (CVanaTime::getInstance()->lastVDailyUpdate + 4800ms))
         {
+            // Vanadiel Time daily updates ("midnight")
 			zoneutils::ForEachZone([](CZone* PZone)
 			{
                 luautils::OnGameDay(PZone);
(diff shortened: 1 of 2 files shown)


src/map/utils/guildutils.cpp
```diff
@@ -72,6 +72,10 @@ void Initialize()
     }
     TPZ_DEBUG_BREAK_IF(g_PGuildShopList.size() != 0);
 
+    // Show in log day number used for guild rank items
+    int vanaDay = CVanaTime::getInstance()->getVanaTime() / (60 * 60 * 24);
+    ShowDebug(CL_CYAN"Guild Items initialized for Vanadiel day: %d\n" CL_RESET, vanaDay);
```
Kreidos, line 77: Not sure if we still want to show debug logs every hour once we're quite sure it works? :thinking:

Kreidos:
Looks good to me. If you want to add "Fixes" in front of the 2 issues mentioned in your description, GitHub will automatically close them when this merges. :)

Kreidos:
Also, if I'm not mistaken I think the offset feature was removed due to "doing nothing" when the vanadiel epoch feature was implemented, so perhaps #22 could be marked Fixes as well.

src/map/utils/guildutils.cpp
jarmengaud: It's every day only

TeoTwawki: a vana day is just shy of 1 earth hr.

Suggestion: -if- this is going to be kept even after we're sure it works, change it to ShowNotice, which will display the same but under a diff message type. we use that type for other "thing has happened" notifications in the log.

TeoTwawki: Suggestion: -if- this is going to be kept even after we're sure it works, change it to ShowNotice, which will display the same but under a diff message type. we use that type for other "thing has happened" notifications in the log.

cocosolos: This will only print when the server boots.

src/map/time_server.cpp
```diff
@@ -100,6 +100,7 @@ int32 time_server(time_point tick,CTaskMgr::CTask* PTask)
 				});
 			});
 
+            guildutils::UpdateGuildPointsPattern();
```
cocosolos, line 103: Moving this call to here will call the update function every vanadiel day. It checks `CVanaTime::getInstance()->getSysYearDay()` so it's still going to be using the server timezone. It also refreshes all the guilds patterns, even if it didn't change, so creating a lot of work for no real end result. I think to get the result you want you would need to do something ~~like adding `luautils::GetSettingsVariable("TIMEZONE_OFFSET")` to the midnight check, but then you'd need to also deal with `guildutils::UpdateGuildPointPattern()` since it uses `CVanaTime::getInstance()->getSysYearDay()` as the server var.~~ else.

Edit: I misunderstood how TIMEZONE_OFFSET is used.

src/map/utils/guildutils.cpp
TeoTwawki: Didn't even notice we were talking about code in an initialization function.

zach2good (approved):

zach2good (changes requested):
On closer inspection, I don't see how this fixes the problem, have you tested this? What happened?

Kreidos:
Just so I and others don't forget, I mentioned to @jarmengaud on Discord that I've already written system-timezone-agnostic JST timekeeping functions, as my upcoming RoE Phase 3 work relies on correct JST time for doing daily/timed record scheduling. I will try to find time to clean up and PR that code separately so he can use it. Should make this fix as easy as moving 1 line of code from local midnight block to the new JST midnight one. (Or if we want, ditch the whole local midnight block entirely)

Kreidos:
Superseded by #1215

PR #919, opened 2020-08-04
Implements Crafting Suggestions packet.

bluekirby0:
Also fixes topaz upstream issue #170

I tried asking around for retail accuracy checks. You can see in the code comments where there were uncertainties. I decided to push ahead with the commit due to how frustrating the associated bug it fixes is.

**_I affirm:_**


bluekirby0:
Going to resubmit off of a branch, sorry.

PR #921, opened 2020-08-04
Bestiary QC: Undead

wrenffxi:
**_I affirm:_**

Screw skeletons.  They're like "Those of us with this name are either BLMs or WARs or DRKs, except in this zone where we're all BLMs, and in these three zones we have different flags, and in these two zones we have different animation subs, and in this one zone we have an unusual model.  Also there's an NM version with the same name in one zone."


kaincenteno:
awesome audit Wren =)

ibm2431:
Your description speaks to my soul, Wren.

59blargedy (approved):

zach2good (approved):

PR #922, opened 2020-08-04
added lucid item usability

kaincenteno:
changed casting time to 1 second and added script, now caskets should be livelier

1 thing missing would be lucid_wings_i and lucid_wings_ii that are still missing. If someone sees this PR and can mimic the same for those 2 items would be awesome =)

Thank you @Omnione  for making caskets available =)

**_I affirm:_**


scripts/globals/items/lucid_elixir_i.lua
```diff
@@ -0,0 +1,20 @@
...
+        return tpz.msg.basic.ITEM_UNABLE_TO_USE_2
+    end
+    return 0
+end
+
+function onItemUse(target)
+    target:addHP((target:getMaxHP()/100)*50*ITEM_POWER)
+    target:addMP((target:getMaxMP()/100)*50*ITEM_POWER)
+    target:messageBasic(tpz.msg.basic.RECOVERS_HP_AND_MP)
+end
```
megberry: add line feed to final lines, to match project standard

scripts/globals/items/lucid_potion_iii.lua
```diff
@@ -0,0 +1,18 @@
...
+
+function onItemCheck(target)
+    if target:getHP() == target:getMaxHP() then
+        return tpz.msg.basic.ITEM_UNABLE_TO_USE
+    end
+    return 0
+end
+
+function onItemUse(target)
+    target:messageBasic(tpz.msg.basic.RECOVERS_HP, 0, target:addHP(1000*ITEM_POWER))
```
wrenffxi: restores 2000 HP

sql/item_usable.sql
```diff
@@ -1623,14 +1623,14 @@ INSERT INTO `item_usable` VALUES (5820,'darkling_bolt_quiver',1,4,55,0,0,0,0,0);
...
-INSERT INTO `item_usable` VALUES (5826,'lucid_potion_iii',1,4,31,0,0,0,0,0);
-INSERT INTO `item_usable` VALUES (5827,'lucid_ether_i',1,4,0,32,0,0,0,0);
-INSERT INTO `item_usable` VALUES (5828,'lucid_ether_ii',1,4,0,33,0,0,0,0);
-INSERT INTO `item_usable` VALUES (5829,'lucid_ether_iii',1,4,0,33,0,0,0,0);
-INSERT INTO `item_usable` VALUES (5830,'lucid_elixir_i',1,4,34,0,0,0,0,0);
-INSERT INTO `item_usable` VALUES (5831,'lucid_elixir_ii',1,4,34,0,0,0,0,0);
+INSERT INTO `item_usable` VALUES (5824,'lucid_potion_i',1,1,30,0,0,0,0,0);
+INSERT INTO `item_usable` VALUES (5825,'lucid_potion_ii',1,1,31,0,0,0,0,0);
+INSERT INTO `item_usable` VALUES (5826,'lucid_potion_iii',1,1,31,0,0,0,0,0);
+INSERT INTO `item_usable` VALUES (5827,'lucid_ether_i',1,1,0,32,0,0,0,0);
```
wrenffxi: the three inserts for ethers are missing animation (probably the value in animation time is supposed to be in the animation column)

kaincenteno: omg ty i didnt notice <3

scripts/globals/items/lucid_potion_iii.lua
kaincenteno: thank you for review <3

scripts/globals/items/lucid_elixir_i.lua
kaincenteno: was not aware, ty <3

kaincenteno:
Thank you for your review <3 I have made the changes asked, also followed teo's advise to reduce one division.

scripts/globals/items/lucid_elixir_ii.lua
```diff
@@ -0,0 +1,20 @@
...
+
+function onItemCheck(target)
+    if target:getMaxHP() == target:getHP() and target:getMaxMP() == target:getMP() then
+        return tpz.msg.basic.ITEM_UNABLE_TO_USE_2
+    end
+    return 0
+end
+
+function onItemUse(target)
+    target:addHP(target:getMaxHP()*0.75*ITEM_POWER)
```
TeoTwawki, line 17: this is one of those case where "extra" parentheses can make the order of operations clearer.
(new people unfamiliar with C++ may wonder if the if item power is applied to the whole amount or to the .75 but this comment is not intended to say you need to change it, so long as it works correctly)

wrenffxi (approved):
Tried all eight items in-game.  All work as expected.

59blargedy (approved):

PR #925, opened 2020-08-06
Aftermath application not damage dependant, applies before the weapon…

rude-jerk:
…skill damage calculation

**_I affirm:_**

* Moved aftermath application above weaponskill damage calculation
* Removed 0 damage check for aftermath application 

Aftermath currently (and has always) applies even for aftermath-applying weaponskills that do no damage. Aftermaths are also always taken into effect for the weaponskill that caused them.


59blargedy (approved):
Thanks rude-jerk - a note for the pr - resolution is still needed to apply aftermath on out of range ws

PR #930, opened 2020-08-07
Fix for null item crash introduced in PR #857

jarmengaud:
**_I affirm:_**

Regression of PR #857:
We cannot unequip a weapon (main / sub and range slots) without crashing the server

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -2186,16 +2186,11 @@ namespace charutils
         }
         if (equipSlotID == SLOT_MAIN || equipSlotID == SLOT_RANGED || equipSlotID == SLOT_SUB)
         {
-            if ((PItem != nullptr) && PItem->isType(ITEM_EQUIPMENT))
+            if (!PItem || !PItem->isType(ITEM_EQUIPMENT) ||
+                ( ((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT &&
+                  ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT ))
             {
-                // If the weapon ISN'T an wind based instrument or an string based instrument
-                if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
-                {
-                    PChar->health.tp = 0;
-                }
-            }
-            else
-            {
+                // If the weapon ISN'T a wind based instrument or a string based instrument
                 PChar->health.tp = 0;
             }
 
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -2186,8 +2186,15 @@ namespace charutils
         }
         if (equipSlotID == SLOT_MAIN || equipSlotID == SLOT_RANGED || equipSlotID == SLOT_SUB)
         {
-            // If the weapon ISN'T an wind based instrument or an string based instrument
-            if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
+            if ((PItem != nullptr) && PItem->isType(ITEM_EQUIPMENT))
+            {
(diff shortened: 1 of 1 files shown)


src/map/utils/charutils.cpp
```diff
@@ -2186,8 +2186,15 @@ namespace charutils
         }
         if (equipSlotID == SLOT_MAIN || equipSlotID == SLOT_RANGED || equipSlotID == SLOT_SUB)
         {
-            // If the weapon ISN'T an wind based instrument or an string based instrument
-            if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
+            if ((PItem != nullptr) && PItem->isType(ITEM_EQUIPMENT))
+            {
+                // If the weapon ISN'T an wind based instrument or an string based instrument
+                if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
+                {
```
Kreidos: This if just contains another if... Which does the same as the else statement below... I think we can simplify this to just one if with no else statement at all.

src/map/utils/charutils.cpp
```diff
@@ -2186,8 +2186,15 @@ namespace charutils
         }
         if (equipSlotID == SLOT_MAIN || equipSlotID == SLOT_RANGED || equipSlotID == SLOT_SUB)
         {
-            // If the weapon ISN'T an wind based instrument or an string based instrument
-            if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
+            if ((PItem != nullptr) && PItem->isType(ITEM_EQUIPMENT))
```
TeoTwawki: `if ((PItem)` is the same as `if ((PItem != nullptr)` and I think has been the format the project has been favoring (someone correct me if I am wrong)

Kreidos (changes requested):
Just a little restructuring would be nice to make the intention here a bit clearer. :)

src/map/utils/charutils.cpp
```diff
@@ -2186,8 +2186,15 @@ namespace charutils
...
+            {
+                // If the weapon ISN'T an wind based instrument or an string based instrument
+                if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
+                {
+                    PChar->health.tp = 0;
+                }
+            }
+            else
```
TeoTwawki, line 2197: Not sure under what condition you would hit this else block that would be retail accurate tp loss.

Kreidos (approved):

src/map/utils/charutils.cpp
```diff
@@ -2186,9 +2186,11 @@ namespace charutils
         }
         if (equipSlotID == SLOT_MAIN || equipSlotID == SLOT_RANGED || equipSlotID == SLOT_SUB)
         {
-            // If the weapon ISN'T an wind based instrument or an string based instrument
-            if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
+            if (!PItem || !PItem->isType(ITEM_EQUIPMENT) ||
```
TeoTwawki, line 2189:
```c++
auto PWeapon = dynamic_cast<CItemWeapon*>(PItem)
if (PWeapon && PWeapon->getSkillType() != SKILL_STRING_INSTRUMENT && PWeapon->getSkillType() != SKILL_WIND_INSTRUMENT)
```
That's what was meant on discord about casting. I haven't tested that and it was suggested to me by a 3rd party

Kreidos: This would work too; minor change on such a small fix but a bit cleaner.

zach2good: +1 on this recommendation, a rare and beautiful possible use of `dynamic_cast`

zach2good (approved):

59blargedy (approved):

Wiggo32 (approved):

PR #933, opened 2020-08-09
Bug with capped skills

### LINKED ISSUE #60: Combat/Magic skills can "cap out" earlier than they're supposed to.
<a href="https://github.com/Hozu"></img></a> **Issue by [Hozu](https://github.com/Hozu)**
_Friday Nov 06, 2015 at 20:24 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 2380_

----

If you reach the skill cap that you would have at the level of the mob (example, 153 skill for A+/- at mob level 50), the client will make the skill appear as capped. However, this can happen when your level is higher than the mob's level, which will prevent gaining skills even if you fight higher level mobs. The workaround for it is to either zone or un/re-equip the weapon to reset it.

jarmengaud:
**_I affirm:_**

Fixes https://github.com/project-topaz/topaz/issues/60

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -2725,7 +2725,7 @@ namespace charutils
         {
             uint16 CurSkill = PChar->RealSkills.skill[SkillID];
             uint16 CapSkill = battleutils::GetMaxSkill(SkillID, PChar->GetMJob(), PChar->GetMLevel());
-            // Max skill this mob level will allow.
+            // Max skill this victim level will allow.
             // Note this is no longer retail accurate, since now 'decent challenge' mobs allow to cap any skill.
             uint16 MaxSkill = battleutils::GetMaxSkill(SkillID, PChar->GetMJob(), std::min(PChar->GetMLevel(), lvl));
 
@@ -2777,7 +2777,6 @@ namespace charutils
                     SkillAmount += 1;
                 }
                 // convert to 10th units
-                MaxSkill = MaxSkill * 10;
                 CapSkill = CapSkill * 10;
 
                 // Do skill amount multiplier (Will only be applied if default setting is changed)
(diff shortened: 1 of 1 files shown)


src/map/utils/charutils.cpp
```diff
@@ -2773,7 +2776,9 @@ namespace charutils
                     tier -= 1;
                     SkillAmount += 1;
                 }
+                // convert to 10th units
                 MaxSkill = MaxSkill * 10;
```
zach2good: This isn't used any more at this point, can be removed

src/map/utils/charutils.cpp
```diff
@@ -2724,6 +2724,9 @@ namespace charutils
         if ((PChar->WorkingSkills.rank[SkillID] != 0) && !(PChar->WorkingSkills.skill[SkillID] & 0x8000))
         {
             uint16 CurSkill = PChar->RealSkills.skill[SkillID];
+            uint16 CapSkill = battleutils::GetMaxSkill(SkillID, PChar->GetMJob(), PChar->GetMLevel());
+            // Max skill this mob level will allow.
```
zach2good: Would be clearer to say "victim" or "target" instead of "mob" here, that's how it's called from `attack.cpp` and `battleutils.cpp` etc.

zach2good (changes requested):
Looks good, just needs a tiny bit of cleanup. 

Fixing a bug is more important than 100% retail accuracy, but if you're here and you understand what's going on, getting this to be retail-accurate would be great. If not, open up a relevant issue when this is merged 👍

jarmengaud:
I know that you can cap skill with decent challenge now, but I don't know the probability.
I would have difficulties trying to find the correct values from retail captures (besides i imagine it depends also on level, so would have to do several captures between 1 to 99...)
So i will just log an issue for the moment

59blargedy (approved):

PR #934, opened 2020-08-09
Allows Smoke on the Mountain to be repeated after completion

tankfest:
**_I affirm:_**

https://ffxiclopedia.fandom.com/wiki/Smoke_on_the_Mountain
Thanks to Nireya@GoldSaucer for reporting this (and identifying the issue).

### DIFF
--- a/scripts/zones/South_Gustaberg/npcs/qm2.lua
+++ b/scripts/zones/South_Gustaberg/npcs/qm2.lua
@@ -13,19 +13,16 @@ function onTrade(player, npc, trade)
     if not player:needToZone() then
         player:setCharVar("SGusta_Sausage_Timer", 0)
     end
-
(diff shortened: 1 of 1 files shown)


tankfest:
I've been told that you can actually make the galkan sausage with or without having the quest active/completed.  This needs verification so it can be corrected if needed.

TeoTwawki:
https://www.twitch.tv/videos/707047868

confirmed: campfire does not care about quest, just takes sheep meat and makes sausage

tankfest:
> 
> 
> https://www.twitch.tv/videos/707047868
> 
> confirmed: campfire does not care about quest, just takes sheep meat and makes sausage

Fixed so that the campfire doesn't check quest status at all now.  Thanks!

59blargedy (approved):

PR #943, opened 2020-08-13
Increases range of all spells from 20.4y to 21.5y as confirmed by tes…

tankfest:
…ting

**_I affirm:_**

Increases spell casting distance from 20.4y to 21.5y according to this retail video: https://www.youtube.com/watch?v=XHVqo2tSqzA&feature=youtu.be

Thanks to @59blargedy (aether) for providing this video ^.^

Thanks to Nireya@GoldSaucer for pointing out this issue.


m241dan:
I wonder if the better approach would be removing the range from the database entirely and gating it behind a feature flag so people could choose whether they want their server to be 20.4 (classic) or 21.5 (new age).

tankfest:
After further testing, turns out the self/party member target spells are 20.8 now.  Need more info for other various spells that were originally 20.4.

PR #945, opened 2020-08-14
Lowers the damage from Goblins' Bomb Toss

tankfest:
**_I affirm:_**

We all know that the suicide Bomb Toss from goblins is too high on Topaz.  On private servers running stock topaz bomb toss scripts, you'll see that no one can actually form an exp party to specifically hunt IT goblins, unlike retail.  This is because suicide bomb toss is doing 2-3x more damage than it should.  The formula developed to emulate SE's bomb toss was made arbitrarily because we don't actually have a method to determine how mob parameters exactly affect the damage.  I propose we slash the bomb toss power in half, and the only evidence I can offer to support this is this video from 2009 where a party is exping on a goblin, and it blows up at 25% for 25ish damage to each member:
https://youtu.be/9-4PGvPxuu8?t=420

On Topaz I believe that same bomb would have caused 60ish damage per party member.  
**Cast your vote to end the tyranny of bomb toss on private servers!**

### DIFF
--- a/scripts/globals/mobskills/bomb_toss.lua
+++ b/scripts/globals/mobskills/bomb_toss.lua
@@ -13,7 +13,7 @@ end
 
 function onMobWeaponSkill(target, mob, skill)
     local dmgmod = 1
-    local info = MobMagicalMove(mob, target, skill, mob:getWeaponDmg()*3, tpz.magic.ele.FIRE, dmgmod, TP_MAB_BONUS, 1)
+    local info = MobMagicalMove(mob, target, skill, mob:getWeaponDmg()*2.75, tpz.magic.ele.FIRE, dmgmod, TP_MAB_BONUS, 1)
     local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.MAGICAL, tpz.damageType.FIRE, MOBPARAM_IGNORE_SHADOWS)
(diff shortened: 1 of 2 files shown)


jarmengaud:
I know that I don't have any official role, but I 100% agree with this change...!

m241dan:
I'm not necessarily opposed to "for now" fixes, but I worry that this will fix it "enough," and then we'll never revisit this issue. So, if people are feeling passionate about it in the moment, I would like to capitalize on that passion and get a bit of research from one of our retail researchers. 

Maybe something like 25 bomb tosses to naked characters per 3-4 different level ranges. Then we could either do some sort of linear regression to figure out the correct formula or just hack something that's a little more average.

zach2good:
> [ ] that I've tested my code since the last commit in the PR, and will test after any later commits

Core experiences need to be tested. Unfortunately for players, getting their asses kicked by Gobbie Bombs is part of the experience. Theorycrafting is fine when it's backed up by a lot of data, one video and some anecdotal evidence is not enough.

> Maybe something like 25 bomb tosses to naked characters per 3-4 different level ranges. Then we could either do some sort of linear regression to figure out the correct formula or just hack something that's a little more average.

This is 10/10 the correct way to do it. As painful as it may be. This is the kind of work that went into figuring out that the damage output from `Pinecone Bomb` was essentially just random...

It sucks, but it's what's required.

PR #946, opened 2020-08-14
Corrects Ebony Pudding spell list

tankfest:
**_I affirm:_**

This PR corrects Ebony Puddings in ToAU to use proper spells.  Before they were linked to the beastmen BLM spell list, which allows them to cast improper spells.

Further research required to determine if this spell list is appropriate for all flans and/or if any additional spells need to be added.


tankfest:
> 
> 
> Is the spell list about the same for all puddings of the same level range? We could maybe use the upper level value and make one list for puddings - I think the problem with the generic blm spell list was that it gave them access to a tier higher than it should have due to the lv, right?

That's what I suspected we could do... but we'd need data from other flan types to confirm.  I just help emulate retail, I don't delve into it myself ^.^

TeoTwawki:
> 
> 
> > Is the spell list about the same for all puddings of the same level range? We could maybe use the upper level value and make one list for puddings - I think the problem with the generic blm spell list was that it gave them access to a tier higher than it should have due to the lv, right?
> 
> That's what I suspected we could do... but we'd need data from other flan types to confirm. I just help emulate retail, I don't delve into it myself ^.^

That is what the capture volunteers are there for Tank.

59blargedy (commented):
i am about 95% sure they also cast stun

TeoTwawki:
They probably have drain if they have aspir.

tankfest:
Drain/Stun added.

59blargedy (approved):

PR #950, opened 2020-08-14
Remove incorrect script for Windy mission 1-1 cracked wall that preve…

tankfest:
…nts it from opening

**_I affirm:_**

A couple of weeks ago, the name of this file was corrected to match its counterpart in our database.  Unfortunately, when that happened, it broke the cracked wall that opens to windurst mission 1-1.  Since no standard door on topaz has a script, this needs to be deleted to allow players access to the magical gizmo that lies within.


59blargedy (approved):

TeoTwawki:
It looks like that script was used by somebody for testing so that they did not have to walk all the way to the correct door for a quest they were working on. then they commented it out and left it in place.

tankfest:
> 
> 
> It looks like that script was used by somebody for testing so that they did not have to walk all the way to the correct door for a quest they were working on. then they commented it out and left it in place.

I had assumed that maybe it WAS the correct script for this name 2 years ago, but SE changed it and the script was never updated.

TeoTwawki:
> I had assumed that maybe it WAS the correct script for this name 2 years ago, but SE changed it and the script was never updated.

possibly if someone made a 2nd script instead of renaming the original. Sometimes the special named object due name change - in some cases the "name" we capture amounts to what offset the npc is, so any shift renamed it. `@05A` or `_5A` being shorthand for 90th object in zone for instance (0x05A). In most cases, our name doesn't even have to match but some object need their special name to function right, like the bcnm entrances up in sky. Or airships. if you rename an airship client go boom.

PR #953, opened 2020-08-15
Adds Kyoshu Kyahan latent effects and new item latent mod

Era-Lusiphur:
Adds Kyoshu Kyahan latent effects and new item latent mod for LATENT_STATUS_ACTIVE_NO_JOBSPECIAL for use when an item specifically does not work during SP. Core work for LATENT_STATUS_ACTIVE_NO_JOBSPECIAL by the ever helpful @TeoTwawki .

Kyoshu Kyahan:
https://ffxiclopedia.fandom.com/wiki/Kyoshu_Kyahan

> Latent Effect: Accuracy +20 Attack +20
> 
>     Active while using Footwork
>     Effect is cancelled during Hundred Fists, even if Footwork status is active.

**_I affirm:_**


m241dan:
Just a suggestion, you could turn this into two commits. One for the content and the second for the white space changes (including fixing the new white space misalignment by adding a well named but long new latent). The reason is this will help keep our history clean.

Otherwise, looks good to me (haven't tested myself, but I trust).

Era-Lusiphur:
> 
> 
> Just a suggestion, you could turn this into two commits. One for the content and the second for the white space changes (including fixing the new white space misalignment by adding a well named but long new latent). The reason is this will help keep our history clean.
> 
> Otherwise, looks good to me (haven't tested myself, but I trust).

Era-Lusiphur:
@m241dan Reduced latent mod name length to preserve original whitespacing. I think increasing the whitespace for just one line would reduce readability, so made the adjustment that way.

TeoTwawki:
@m241dan kinda my fault when I quick proof of concept'ed it last night my auto format made some adjustments which carried into the diff I handed off

src/map/latent_effect.h
```diff
@@ -71,7 +71,7 @@ enum LATENT
     LATENT_JOB_LEVEL_ODD            = 41,
     LATENT_JOB_LEVEL_EVEN           = 42,
     LATENT_WEAPON_DRAWN_HP_UNDER    = 43, //PARAM: HP PERCENT
-    LATENT_STATUS_ACTIVE_NO_JOBSPECIAL = 44, //PARAM: EFFECTID - When status effect active, but without a 1hr/2hr Job Special active
+    LATENT_STATUS_ACTIVE_NO_JOBSP   = 44, //PARAM: EFFECTID - When status effect active, but without a 1hr/2hr Job Special active
```
59blargedy, line 74: since this is currently only applicable on 100 fists, can we add that information to the definitions here and in status lua? otherwise people may try to use it to no effect.

TeoTwawki: Should we just add an effect flag to check against? I thought we had a way to tell if something was a 2hr already but I've been unable to locate it. We could nip it in the bud right now.

sql/item_latents.sql
```diff
@@ -3357,3 +3357,9 @@ INSERT INTO `item_latents` VALUES(27367, 399, 16, 52, 7);   -- Pedagogy Loafers+
 INSERT INTO `item_latents` VALUES(27367, 399, 16, 52, 8);   -- Pedagogy Loafers+1: Weather: Enhances Celerity and Alacrity Effect +16% (DARK)
 INSERT INTO `item_latents` VALUES(28235, 169, 25, 26, 2);   -- Hachiya Kyahan: Dusk to dawn: Movement speed +25%
 INSERT INTO `item_latents` VALUES(28256, 169, 25, 26, 2);   -- Hachiya Kyahan +1: Dusk to dawn: Movement speed +25%
+
+-- -------------------------------------------------------
+-- Kyoshu Kyahan
+-- -------------------------------------------------------
+INSERT INTO `item_latents` VALUES(11405, 23, 20, 44, 406);  -- +20 Att during Footwork, but not during SP
+INSERT INTO `item_latents` VALUES(11405, 25, 20, 44, 406);  -- +20 Acc during Footwork, but not during SP
```
zach2good, line 3365: Can we get a newline at the end here?

src/map/latent_effect_container.h
```diff
@@ -41,42 +41,42 @@ class CLatentEffectContainer
...
 
-	 CLatentEffectContainer(CCharEntity* PEntity);
+     CLatentEffectContainer(CCharEntity* PEntity);
 
 private:
 
-	CCharEntity* m_POwner;
-	std::vector<CLatentEffect>	m_LatentEffectList;
+    CCharEntity* m_POwner;
+    std::vector<CLatentEffect>m_LatentEffectList;
```
zach2good, line 79: Space after angle-bracket please

src/map/latent_effect_container.h
```diff
@@ -41,42 +41,42 @@ class CLatentEffectContainer
...
+    void CheckLatentsHours();
+    void CheckLatentsWeekDay();
+    void CheckLatentsPartyMembers(size_t members);
+    void CheckLatentsPartyJobs();
+    void CheckLatentsPartyAvatar();
+    void CheckLatentsJobLevel();
+    void CheckLatentsPetType();
+    void CheckLatentsTime();
+    void CheckLatentsWeaponBreak(uint8 slot);
+    void CheckLatentsZone();
```
zach2good, line 63: Thanks for aligning these!

src/map/latent_effect.h
zach2good: I think I agree with these guys; just making the all-SP change now will stop the TODO sitting there and rotting forever, especially because it isn't much extra work.

TeoTwawki: @Era-Lusiphur example of using an effectflag:
https://github.com/project-topaz/topaz/commit/d099abc12c8f682520672f7c6acb7435f12857e3

adding the effectflag used above:
https://github.com/project-topaz/topaz/commit/53398a83daf8734319e98800723a1d60a61d88a0

So we could flag 2hr statuses, and then check the flag. alternatively you can check each status, or come up with a 3rd option - effectflag is just all I could think of to avoid a bunch of `||` to check multiple effects.

59blargedy:
do we want to merge and issue, or just get this done?

TeoTwawki:
> do we want to merge and issue, or just get this done?

lets just add an effect flag, add said flag to the effects that 2hrs place on players. check for flag instead of for hundred wiffs.

Flags are as easy to add as mods. Easier..since you don't deal with collision baggage or duplication. [Example.](https://github.com/project-topaz/topaz/commit/a795e51e75d8ff344f73382522462f13af7bcd9f#diff-3207659a2e7b5decb0068fa2b4cb96bc)

59blargedy:
@Era-Lusiphur you want to take care of adding effect flag or do you need help?

PR #957, opened 2020-08-16
PR: DRG mobs now spawn wyverns when they idle

m241dan:
**_I affirm:_**

### DIFF
--- a/src/map/utils/mobutils.cpp
+++ b/src/map/utils/mobutils.cpp
@@ -680,6 +680,10 @@ void SetupJob(CMobEntity* PMob)
             PMob->defaultMobMod(MOBMOD_SPECIAL_COOL, 70);
             PMob->defaultMobMod(MOBMOD_SPECIAL_SKILL, 1017);
             break;
+        case JOB_DRG:
+            PMob->defaultMobMod(MOBMOD_SPECIAL_COOL, 70);
+            PMob->defaultMobMod(MOBMOD_SPECIAL_SKILL, 1022);
+            break;
         case JOB_PUP:
             PMob->defaultMobMod(MOBMOD_SPECIAL_SKILL, 1901);
             PMob->defaultMobMod(MOBMOD_SPECIAL_COOL, 720);


TeoTwawki:
This behavior isn't universal. I believe SE started doing it with ToAU when they moved it from being the jobs 2hr special to being just a non-subable JA. The left all the old mobs alone, but these were mostly NMs (a lot of the old spear wielding mobs were actually not drg job). We will want some way to prevent those older drg mobs/NMs from doing this if something does not already prevent it. If something already does, I think its appropriate to leave a comment in the code there.

m241dan:
> This behavior isn't universal. I believe SE started doing it with ToAU when they moved it from being the jobs 2hr special to being just a non-subable JA. The left all the old mobs alone, but these were mostly NMs (a lot of the old spear wielding mobs were actually not drg job). We will want some way to prevent those older drg mobs/NMs from doing this if something does not already prevent it. If something already does, I think its appropriate to leave a comment in the code there.

We should get some research and tackle this appropriately then? Though, I suppose we could just region lock it. If region ToAU or greater...

TeoTwawki:
Its not that its region specific just they did not update existing mobs, but started making new ones work this way. Probably need to control via mobmod and or mixin, as new mobs in old zones would get the new behavior,

Like you fight Steelbiter Gudrud, is Wyvern is still on a 2hr timer. You fight the old Orcish Dragoons in Monastic Cavern and they still have no Wyvern at all. Every mob created after SE changed DRG's 2hr pop them while idle (including new orcs). Doing a region check would probably cover 90% of them because that would have been right around ToAU's debut and there wouldn't be much being added in older zones, but I believe we'd have some newer BCNM mobs being affected still (which admittedly do not exist in Topaz yet).

59blargedy:
ix'aern drg also does not start with idle wyvern, though i know that's also a special case.

TeoTwawki:
Looks like when they updated dynamis, they kept with the old style of wyvern = 2hr. except they depop when master is dead now and didn't use to do that. No more angry pseudo pet following me around dynamis (I just ran through dyna sandy on retail for an item I needed). 

I think we might be able to do this by region after all, if we confirm newer bcnm drg mobs also follow the old behavior.

m241dan:
> ix'aern drg also does not start with idle wyvern, though i know that's also a special case.

this is a good concern, but does he not? I seem to remember retail he would pop with his whole crew, and we'd have someone kite while we zone nuked. But, we're like... 10 years removed at this point, so I'll defer to you. Perhaps we'll need a special case?

TeoTwawki:
> 
> 
> > ix'aern drg also does not start with idle wyvern, though i know that's also a special case.
> 
> this is a good concern, but does he not? I seem to remember retail he would pop with his whole crew, and we'd have someone kite while we zone nuked. But, we're like... 10 years removed at this point, so I'll defer to you. Perhaps we'll need a special case?

I totally forgot till just now how that fella worked, this is what we do currently:
```lua
function onMobFight(mob, target)
    -- Spawn the pets if they are despawned
    -- TODO: summon animations?
    local mobId = mob:getID()
    local x = mob:getXPos()
    local y = mob:getYPos()
    local z = mob:getZPos()

    for i = mobId + 1, mobId + 3 do
        local wynav = GetMobByID(i)
        if not wynav:isSpawned() then
            local repopWynavs = wynav:getLocalVar("repop") -- see Wynav script
            if mob:getBattleTime() - repopWynavs > 10 then
                wynav:setSpawn(x + math.random(1, 5), y, z + math.random(1, 5))
                wynav:spawn()
                wynav:updateEnmity(target)
            end
        end
    end
end
```

.. so every 10 seconds..

59blargedy:
Yeah, i just popped him the other day for damage testing on retail. He just sits there, and only spawned wynavs on engage (sorry im challenged on mobile and cant quote)

m241dan:
> Yeah, i just popped him the other day for damage testing on retail. He just sits there, and only spawned wynavs on engage (sorry im challenged on mobile and cant quote)

Okay, will make changes.

TeoTwawki:
iirc he will continue to call them after they die also, which is why it got done in onMobFight in the script

TeoTwawki:
*he

PR #958, opened 2020-08-16
ENM - Operation Desert Swarm

m241dan:
**_I affirm:_**

The only thing that is missing is the building sleep mechanic.
I didn't feel like we should hack this, but at some point, we need to implement building resistance mechanics for the server.


zach2good (approved):

TeoTwawki (approved):

zach2good:
Want to get this into its own feature branch with Dan's other ENMs so we can get this into canary ASAP

PR #961, opened 2020-08-16
ENM - Shell We Dance

m241dan:
**_I affirm:_**

The only thing missing is that the whole family of these mobs does not go into their shell like they are supposed to. It makes the BC a little easier than it should be but what can you do? Someone should make that mixin.


m241dan:
Rebased with Release and fixed the spell_list conflict created by adding the Puddings since submitting.

scripts/zones/Boneyard_Gully/IDs.lua
```diff
@@ -17,11 +17,32 @@ zones[tpz.zone.BONEYARD_GULLY] =
...
     mob =
-    {
+	{
     },
     npc =
     {
     },
+	shellWeDance = {
+		[1] = {
+			PARATA_ID        = 16810024,
```
TeoTwawki: them sneaky tabs gotchu

sql/mob_spell_lists.sql
```diff
@@ -3890,7 +3890,16 @@ INSERT INTO `mob_spell_lists` VALUES ('Ebony_Pudding',430,258,7,255);  -- Bind
 INSERT INTO `mob_spell_lists` VALUES ('Ebony_Pudding',430,273,31,255); -- Sleepga
 INSERT INTO `mob_spell_lists` VALUES ('Ebony_Pudding',430,274,56,255); -- Sleepga II
 
-
+INSERT INTO `mob_spell_lists` VALUES ('Bladmall',431,46,63,255);                  -- Protect IV
+INSERT INTO `mob_spell_lists` VALUES ('Bladmall',431,133,68,255);          		 -- Shellra IV
```
TeoTwawki: whitespace appears to have mixed tabs and spaces in same line here.

sql/mob_pool_mods.sql
```diff
@@ -59,6 +59,7 @@ INSERT INTO `mob_pool_mods` VALUES (276,12,1,1);
 INSERT INTO `mob_pool_mods` VALUES (289,4,4,1);
 INSERT INTO `mob_pool_mods` VALUES (370,28,5,1);
 INSERT INTO `mob_pool_mods` VALUES (410,236,20,0);
+INSERT INTO `mob_pool_mods` VALUES (444,23,23,1); -- Bladmall immune to Sleep, Bind, Silence and Gravity.
```
m241dan, line 62: Is this the correct way to do this? I think this needs to be changed.
@TeoTwawki

TeoTwawki: true immunity is a field in mob pools, don't add as a mod.

high resistance will be a mod but only put here if the mob is non-scripted and it affects all mobs using the pool (if its on mob, prolly belongs in the spawn mods if non-scripted)

you can tell apart resist from immunity by the log messages in retail. 

Regular old high resistance:
_"The \<player\> casts \<spell\>. \<target\> resists the spell."_

Immunity:
_"completely resists the spell"_

Job trait proc:
_"Resist!"_

Immunobreak (for completeness):
_"resists the spell"_
_"Immunobreak!"_

59blargedy:
not sure if the mixin should be implemented for this script now? they may have different properties, but figured i'd ask

zach2good:
There is a very handy mixin you can copy/steal/appropriate that handles mobs going into a shell:
https://github.com/project-topaz/topaz/pull/989

EDIT: On closer inspection, this fight is Uragnites, the mixin is Uragnites... sounds like a match to me!

zach2good:
Merged into `feature/enm` and shifted around all the SQL required. Gonna open an issue for adding the Urganite mixin to the mobs. Thanks for the ENM!

PR #965, opened 2020-08-17
Replace TravisCI with GitHub Workflows

zach2good:
**_I affirm:_**

I got caught out a couple of weeks ago because I was working exclusively with 32bit Windows and a casting warning-as-error snuck in on the 64-bit build. I did a little digging and Travis's support for Windows and multi-OS builds is a little lacking, so I split everything out into GitHub Workflows.

**Note:**
I wasn't paying enough attention to notice that Appveyor covers the gap I got caught by, but then, why didn't it? Hmmm.
Still probably better to have all CI in one place

**Successful run:**
https://github.com/zach2good/topaz/runs/1041773878
image

**C++ Formatting Error**
image

image
image


.github/workflows/build.yml
```diff
@@ -0,0 +1,144 @@
...
+  Linux_Build_32bit:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: Install Dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y software-properties-common g++-8 luajit-5.1-dev libzmq3-dev mysql-client-core-5.7 mysql-server-5.7 mysql-client-5.7 luarocks python3.7
```
zach2good: Mysql stuff should be 5.6

.github/workflows/tag_releases.yml
```diff
@@ -0,0 +1,71 @@
+name: tag_releases
```
zach2good: This shouldn't be in this PR, this was a mistake!

.github/workflows/build.yml
TeoTwawki: 5.7 kind of works, with a few warnings about whats depricated, and potentially a setting that may need changed to avoid issues with timestamps iirc. but if we use mariadb instead of stock mysql even the latest version has zero issues

also did you just review your own pr lol?

zach2good: Maybe I did... who knows...

zach2good:
If someone likes this I'll merge it.

zach2good:
Say no more fam.

PR #881, opened 2020-07-23
Create MOD for "Resistance to All Status Ailments"

neuromancerxi:
**_I affirm:_**
- [x] that I agree to Project Topaz's [Limited Contributor License Agreement](http://project-topaz.com/blob/release/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [x] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

Creates MOD 958 - STATUSRES for "Resistance to All Status Ailments".
Removes Death Resistance from Items with "All Status Ailments" (see: related BG Thread with justification - https://www.bluegartr.com/threads/108199-Random-Facts-Thread-Other?s=c66ef360f8f47cb9a568306781e9a045&p=6981981&viewfull=1#post6981981 )
Replaces Individual Resist MODS (240-255) on existing "All Status Ailment" Items with a single 958 with relevant value.
Adds math check in magic.lua to combine specific resistance values with the newly created All Status Ailments MOD.

PR #966, opened 2020-08-17
Implements dead aim trait

tankfest:
**_I affirm:_**

Gives dead aim trait to rangers.

Tests:
- Assigns the proper mod value to each level of ranger.
- Mod value increases ranged crit damage by the specified %.

Thanks to Eren@GoldSaucer for bringing this missing content up.


m241dan:
Looks good and bg-wiki accurate.

src/map/modifier.h
```diff
@@ -800,7 +801,6 @@ enum class Mod
 
     // The spares take care of finding the next ID to use so long as we don't forget to list IDs that have been freed up by refactoring.
     // 570 through 825 used by WS DMG mods these are not spares.
-    // SPARE = 964, // stuff
```
TeoTwawki, line 804:
```lua
    -- SPARE = 958, -- stuff
    -- SPARE = 959, -- stuff
    -- SPARE = 960, -- stuff
```
(see line 1562 of status.lua)

status.lua's mods and spares should 100% match modifier.h's mods and spares. I should not see any different numbers for mods or for spares in either file. when a mod gets refactored out its ID should go back into spares. Topaz contributors what have'yall been doing? These are out of sync again. If there's IDs used by another branch that we're working around there needs to be notes on the PR letting reviewers know this. @zircon-tpl this is a recurring problem. if the numbers do not match it leads people to accidentally use the same one for 2 different things. see previous discord conversation with ibm on the subject.


@tankfest you used 936 but removed spare 964? if we're up to 964 936 should already be in use by something, but I don't see anything using either number so I am confused. And see above general note.

m241dan: Well, what we need to do is stop loading static data from the database, and then these could be generated by the code, ie loading item stats from lua. That's outside the scope of this PR for sure but just a general comment to maybe get some creative juices flowing.

59blargedy:
will merge this after the status ailment one gets merged to release to avoid issues

src/map/modifier.h
TeoTwawki: @m241dan ibm had a plan and I don't think any contributor currently has both the time and the technical knowhow to make it happen now (by "it" I mean something better than copying what currencies do, which is do get/set with a string value..) but as a future thing I agree that mods all in one place should happen. This has zilch to do with the database atm and everything to do with duplicated redundant enums.

59blargedy: @tankfest can you resolve to match spares in status and modifier to equal the ones found here, as it was merged first though not yet in release?
https://github.com/project-topaz/topaz/pull/881 
I'll approve after that, just want to make it easier for conflict resolution

tankfest: Used the previously annotated SPARE = 964, updated the next 3 spare mods.

59blargedy:
whoever merges this will have to resolve the conflict with the other thing using 964 good luck
https://github.com/project-topaz/topaz/pull/982/commits

59blargedy (approved):

PR #968, opened 2020-08-18
Adds Tachi/+1 and Raptor Strap/+1 desynths

tankfest:
**_I affirm:_**

Thanks to Nireya@GoldSaucer for reporting these.


m241dan:
Only concern is that https://ffxiclopedia.fandom.com/wiki/Tachi claims its lv31 for the Tachi desynth (though 33 for the craft). Bg-wiki doesn't specify.

tankfest:
I saw that claim, both in ffxiah's message board and bgwiki.... but are we to believe a level 33 synth has a level 31 desynth?  what madness is this?

59blargedy (approved):

m241dan:
> I saw that claim, both in ffxiah's message board and bgwiki.... but are we to believe a level 33 synth has a level 31 desynth? what madness is this?

The most maddening madness, we should probably pick one and leave a comment in the code saying "we saw this, but this is what we chose"

TeoTwawki:
> leave a comment

this here? that is how you keep people from fixing your fix for the fix that they fixed that you fixed... something something Halver

Wiggo32 (approved):

Issue #967, opened 2020-08-18
Feature Request:  Command to add/remove mods from an entity manually


tankfest:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

We can use !getmod to see what the value of a mod is on an entity.  For testing new/changed mods functionality it'd be nice to have the ability to change mod values for ourselves and others at a whim.

wrenffxi:
Closed via #972, which is now in release

PR #972, opened 2020-08-18
Add `!setmod` command to quickly manipulate modifiers on player/mobster/pet/fellow/trust entities

TeoTwawki:
**_I affirm:_**

#967
* Note: Some modifiers expect to be on an item, not a player. This command can only apply mods directly to an entity.

### DIFF
--- a/scripts/commands/setmod.lua
+++ b/scripts/commands/setmod.lua
@@ -1,0 +1,45 @@
+---------------------------------------------------------------------------------------------------
+-- func: setmod
+-- desc: Sets the specified modifier to the specified value on the cursor target or player
+---------------------------------------------------------------------------------------------------
+
+require("scripts/globals/status")
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "sis"
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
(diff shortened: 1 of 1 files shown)


zach2good (approved):

Wiggo32 (approved):

m241dan (approved):
Only criticism is having multiple return points, but I know that's not a convention we stick to as a project.

So is good!

scripts/commands/setmod.lua
```diff
@@ -0,0 +1,45 @@
...
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "sis"
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!setmod {modID} {optional target}")
```
tankfest: !setmod {modID} **{mod_amount}** {optional target}

Missing a parameter in the error readout, it might be confusing to peeps that use it in game to see what the inputs should be.

TeoTwawki: nice catch fixing message

TeoTwawki: fixed also fixed my "the" in the header and made the variable naming a little easier on the eyes.

PR #977, opened 2020-08-19
Stops wamoura princes from aggroing

tankfest:
**_I affirm:_**

These should not aggro: https://ffxiclopedia.fandom.com/wiki/Wamoura_Prince

Thanks to Daemun@GoldSaucer for reporting this.


m241dan (approved):
good branch name

looks good.

59blargedy (approved):

PR #980, opened 2020-08-19
Implements Bartholomew's Knife mini-quest

### LINKED ISSUE #978: Bartholomew's Knife mini-quest doesn't work
**_I have:_**

Can't obtain [Bartholomew's Knife](https://ffxiclopedia.fandom.com/wiki/Bartholomew's_Knife).

tankfest:
**_I affirm:_**

Fixes #978 

https://ffxiclopedia.fandom.com/wiki/Talk:Bartholomew%27s_Knife

### DIFF
--- a/scripts/zones/VeLugannon_Palace/IDs.lua
+++ b/scripts/zones/VeLugannon_Palace/IDs.lua
@@ -17,6 +17,9 @@ zones[tpz.zone.VELUGANNON_PALACE] =
         NOTHING_OUT_OF_ORDINARY  = 6402,  -- There is nothing out of the ordinary here.
         CONQUEST_BASE            = 7049,  -- Tallying conquest results...
         CHEST_UNLOCKED           = 7216,  -- You unlock the chest!
+        EVIL_PRESENCE            = 7226,  -- You sense an evil presence lurking in the shadows...
+        KNIFE_CHANGES_SHAPE      = 7233,  -- The <item> begins to change shape.
+        NOTHING_HAPPENS          = 7234,  -- Nothing happens.
         REGIME_REGISTERED        = 10159, -- New training regime registered!
         PLAYER_OBTAINS_ITEM      = 11211, -- <name> obtains <item>!
         UNABLE_TO_OBTAIN_ITEM    = 11212, -- You were unable to obtain the item.
@@ -31,6 +34,7 @@ zones[tpz.zone.VELUGANNON_PALACE] =
     },
     npc =
     {
+        QM3             = 17502583,
         CASKET_BASE     = 17502585,
         Y_DOOR_OFFSET   = 17502608,
         B_DOOR_OFFSET   = 17502616,
--- a/scripts/zones/VeLugannon_Palace/mobs/Brigandish_Blade.lua
+++ b/scripts/zones/VeLugannon_Palace/mobs/Brigandish_Blade.lua
@@ -2,6 +2,7 @@
 -- Area: VeLugannon Palace
 --   NM: Brigandish Blade
 -----------------------------------
+local ID = require("scripts/zones/VeLugannon_Palace/IDs")
 require("scripts/globals/mobs")
 -----------------------------------
 
@@ -14,4 +15,5 @@ function onAdditionalEffect(mob, target, damage)
 end
 
 function onMobDeath(mob, player, isKiller)
+    GetNPCByID(ID.npc.QM3):setLocalVar("PillarCharged", 1)
 end
--- a/scripts/zones/VeLugannon_Palace/npcs/qm3.lua
+++ b/scripts/zones/VeLugannon_Palace/npcs/qm3.lua
@@ -1,0 +1,32 @@
+-----------------------------------
+-- Area: VeLugannon Palace
+--  NPC: qm3 (???) 17502583
+-- Note: Involved in Bartholomew's Knife mini-quest
+-- !pos 0.21 0.57 -322.4 177
+-----------------------------------
+local ID = require("scripts/zones/VeLugannon_Palace/IDs")
+require("scripts/globals/npc_util")
+-----------------------------------
+
+function onTrade(player, npc, trade)
+    if npcUtil.tradeHas(trade, 17622) then -- Buccaneer's Knife
+        if npc:getLocalVar("PillarCharged") == 1 and math.random(1000) <= 50 then -- 50/1000 chance to obtain on trade.
+            player:confirmTrade()
+            player:messageSpecial(ID.text.KNIFE_CHANGES_SHAPE, 17622)
+            npcUtil.giveItem(player, 17623) -- Btm. Knife
+        else
+            player:messageSpecial(ID.text.NOTHING_HAPPENS)
+        end
+        npc:setLocalVar("PillarCharged", 0) -- Pillar always loses charge after a morph attempt.
+    end
+end
+
+function onTrigger(player, npc)
+    player:startEvent(2)
+end
+
+function onEventUpdate(player, csid, option)
+end
+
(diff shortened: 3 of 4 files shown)


scripts/zones/VeLugannon_Palace/IDs.lua
```diff
@@ -31,6 +33,7 @@ zones[tpz.zone.VELUGANNON_PALACE] =
     },
     npc =
     {
+        QM3             = 17502583,
```
m241dan, line 37: any chance we could use better naming convention then "QM3"? I know there are some client limitations here so maybe ignore this. Teo can probably weigh in.

scripts/zones/VeLugannon_Palace/mobs/Brigandish_Blade.lua
```diff
@@ -14,4 +15,5 @@ function onAdditionalEffect(mob, target, damage)
 end
 
 function onMobDeath(mob, player, isKiller)
+    GetNPCByID(ID.npc.QM3):setLocalVar("PillarCharged", 1)
```
m241dan, line 18: I think this needs to be followed up by a timer. 

Though, I'm not finding any research, there is a limited window on how long the pillar remains "charged" for? If I'm wrong, no worries and ignore me.

scripts/zones/VeLugannon_Palace/npcs/qm3.lua
```diff
@@ -0,0 +1,32 @@
...
+-----------------------------------
+
+function onTrade(player, npc, trade)
+    if npcUtil.tradeHas(trade, 17622) then -- Buccaneer's Knife
+        if npc:getLocalVar("PillarCharged") == 1 and math.random(1000) <= 50 then -- 50/1000 chance to obtain on trade.
+            player:confirmTrade()
+            player:messageSpecial(ID.text.KNIFE_CHANGES_SHAPE, 17622)
+            npcUtil.giveItem(player, 17623) -- Btm. Knife
+        else
+            player:messageSpecial(ID.text.NOTHING_HAPPENS)
```
m241dan, line 18: My understanding is, if you trade that knife and don't get it, it essentially "breaks" like an orb and you can never trade that same knife again. You need to steal a fresh one.

This may require a core change unless you can use the same mechanic we use to break orbs?

m241dan (changes requested):

scripts/zones/VeLugannon_Palace/npcs/qm3.lua
tankfest: Testimonies clearly conflict on this.  That aside, burning these knives somehow will require some special code if that's how you guys want to do it.

scripts/zones/VeLugannon_Palace/mobs/Brigandish_Blade.lua
tankfest: Conflicting accounts on this, but I thought this had been debunked.

scripts/zones/VeLugannon_Palace/npcs/qm3.lua
m241dan: Well... it would be nice to implement this once and completely, but that's not my call to make. The "good enough" approach bugs me because we'll never come back to it, but again that's personal preference.

scripts/zones/VeLugannon_Palace/mobs/Brigandish_Blade.lua
m241dan: Well, if so, no worries then.

scripts/zones/VeLugannon_Palace/npcs/qm3.lua
tankfest: I looked at the main page for the knife again, someone said in 2020 that it is guaranteed burnt when you try to trade and it fails... Implementation of this is beyond my knowledge, but don't let this PR languish just because it's not 100% :(

scripts/zones/VeLugannon_Palace/IDs.lua
wrenffxi: Agreed.  One of my eventual TODOs is getting rid of all these numbered QMs in npc_list/scripts, and replacing them with more meaningful names.  Client NPC ID shifts sometimes add/remove/shift QMs in a way that messes up the order within a zone.

tankfest: Just need to know what name to change it to.

m241dan: I would say just leave it for now, unfortunately. That way it still matches up with the name in the database.

scripts/zones/VeLugannon_Palace/npcs/qm3.lua
tankfest: After delving into how WS points are stored for weapons, we could use the "extra" field in char_inventory for each item to store if it's burnt or not.  The exact usage of this field is a little beyond me, but with a little guidance I could add it if you guys want it.

TeoTwawki: > Well... it would be nice to implement this once and completely, but that's not my call to make. The "good enough" approach bugs me because we'll never come back to it, but again that's personal preference.

if someone can _prove it_ I'll make sure it isn't forgotten. I don't trust some random comment on wiki to be the end all be all for it though. Whichever way its implemented initially, this is still worthy of an issue until verified.

zach2good:
So, is my thinking right here?

```
Brigandish Blade dies
QM3/Pillar is "charged" with a localVar
Someone tries to trade Buccaneer's Knife to it, and it _might_ return Bart's Knife if the pillar is charged
Pillar is then uncharged
```

I suspect there might be some kind of timer here, else I could come and kill Brigandish Blade and then wander off, allowing some unrelated party to swing by an hour later and try to trade their knife?

For this "knife getting burned" thing, I think this is a good opportunity for someone (me, teo, anyone comfortable with core bindings) to write a get/setItemExtraData binding. 

Finally, I think the best options would be to put on a 1 minute timer, and add the "knfie being burned" thing as precautions (since they aren't hard) and then someone can come and debunk these claims later. My fear is that if we do something "good enough" it will stay _incomplete_ forever, until someone does the worlds most in-depth audit. Which is never.

zach2good (changes requested):
Code is all fine and clean, I'd say plop in a 1 minute timer when the pillar is charged that will uncharge it, and then wait for someone to write a binding for item m_extra.

ghost:
When checking the ???. that spawns Brigandish Blade:

INCOMING < NPC Chat (0x036):  NPC: 17502582 (???)
Message: 7226.
("You sense an evil presence lurking in the shadows...")

Two different Buccaneer's Knives in and still unable to get the Bartholomew's Knife. I'll edit once I get it (if I can).

TeoTwawki:
> and add the "knfie being burned" thing as precautions (since they aren't hard)

easy enough for those of us that disbelieve/don't want to just comment out too.

TeoTwawki:
@UynGH try to cap the item data when you get a new knife and after using that knife. the theory is that the knife gets "burned" like an orb, and if we can see that in its data that's confirmation if they did it that way (not seeing it doesn't prove its good for more trades, but it will save a lot of extra effort checking if we do happen to see it).

tankfest:
Correct dialogue now added to brigandine blade's ??? spawn, qm2.

Thanks for looking into that nyu <3

zach2good:
Since Tank said he was going to be away from the community for a while and we've been arguing trying to prove or disprove the logic here, and then doing deep dives into what mechanisms would be best to "burn" a knife, I think it would be best to just get it in and follow up with an issue.

PR #981, opened 2020-08-20
[fix] regimes always rewarding alliance members, regardless of fov_allow_alliance value

wrenffxi:
**_I affirm:_**

LUA processes `not` before `==` so this check was always failing.


Era-Lusiphur:
Tested. Works for me.

m241dan (approved):
I did notice this a while back and forgot to make an issue in my own tracker, thanks for the Fix!

wrenffxi:
regex scanned our lua for this order-of-operations issue.  found this one other case.

m241dan (approved):

59blargedy (approved):

PR #982, opened 2020-08-20
feat(ability) Paladin - Cover

### LINKED ISSUE #233: Cover ability not present.
<a href="https://github.com/Dynas13"></img></a> **Issue by [Dynas13](https://github.com/Dynas13)**
_Tuesday Jul 17, 2018 at 01:51 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 5095_

----

**_I have:_**

all

master

The status effect exists, the animation works, the necessary elements to make it happen are there but there's no development on making Cover.

jdsmall25:
**_I affirm:_**

This PR fully implements Cover to the best of my knowledge and should close #233 .  I haven't played retail in some time so I did make some assumptions so please let me know if I've made any mistakes.  I referenced the BG wiki page for details: https://www.bg-wiki.com/bg/Cover

I mirrored the logic after trick attack and introduced equipment checks in the C code because from my understanding the augments only occur if the equipment is on during the cover not on use.  Duration on the other hand is calculated on use.

Please let me know what/if corrections need to be made.


scripts/globals/abilities/cover.lua
```diff
@@ -0,0 +1,66 @@
...
+
+    if (helm == 27690 or helm == 23046) then -- Reverence Cornet +1 / +2 - duration + 9
+        bonusTime = bonusTime + 9
+    elseif (helm == 23381) then              -- Reverence Coronet +3 - duration + 10
+        bonusTime = bonusTime + 10
+    end
+
+    local duration = baseDuration + bonusTime + player:getMerit(tpz.merit.COVER_EFFECT_LENGTH)
+
+    if (body == 15093 or body == 14506) then -- Valor Surcoat & +1
```
m241dan: Would highly recommend creating a mod for COVER_TO_MP, and add that to the items instead of checking against item IDs.

scripts/globals/effects/cover.lua
```diff
@@ -10,5 +10,6 @@ end
 function onEffectTick(target, effect)
 end
 
-function onEffectLose(target, effect)
+function onEffectLose(target,effect)
```
m241dan: Please put space between (target,effect) -> (target, effect)

scripts/globals/abilities/cover.lua
```diff
@@ -0,0 +1,66 @@
...
+function onUseAbility(player,target,ability)
+    local power = 0
+    local baseDuration = 15
+    local bonusTime    = utils.clamp(math.floor((player:getStat(tpz.mod.VIT) + player:getStat(tpz.mod.MND) - target:getStat(tpz.mod.VIT) * 2) / 4), 0, 15)
+
+    local main         = player:getEquipID(tpz.slot.MAIN)
+    local helm         = player:getEquipID(tpz.slot.HEAD)
+    local body         = player:getEquipID(tpz.slot.BODY)
+
+    if (main == 16604) then                  -- Save the Queen - duration + 5
```
m241dan: Would highly recommend creating a mod for COVER_DURATION and adding it to those items instead of checking against the item IDs directly.

src/map/attack.h
```diff
@@ -71,51 +71,54 @@ class CAttack
...
-    int32                       m_trickAttackDamage {0};    // The damage from trick attack.
-    int32                       m_naturalH2hDamage {0};     // The damage from natural H2H.
-    int32                       m_baseDamage {0};           // The base damage.
+    bool                        m_isFirstSwing {false};  // Flag: Is this attack the first swing?
+    float                       m_damageRatio {false};   // The damage ratio.
+    int32                       m_damage {0};            // The damage for this attack.
+    int32                       m_trickAttackDamage {0}; // The damage from trick attack.
+    int32                       m_naturalH2hDamage {0};  // The damage from natural H2H.
+    int32                       m_baseDamage {0};        // The base damage.
 };
```
m241dan: If possible, please either under white space changes or put them in a separate commit.

src/map/attackround.h
```diff
@@ -46,28 +46,30 @@ class CAttackRound
...
+    CBattleEntity*       m_attacker;            // The attacker.
+    CBattleEntity*       m_defender;            // The defender.
+    CBattleEntity*       m_taEntity;            // The trick attack entity.
+    CBattleEntity*       m_coverEntity;         // The cover target entity
+    std::vector<CAttack> m_attackSwings;        // The list of attacks for this round.
+    bool                 m_sataOccured;         // Flag: Did SATA occur during the round?
+    bool                 m_kickAttackOccured;   // Flag: Did a kick attack occur during the round?
+    uint16               m_subWeaponType;       // The sub weapon type.
 
 };
```
m241dan: ibid, whitespace changes.

src/map/utils/battleutils.cpp
```diff
@@ -5624,4 +5634,151 @@ namespace battleutils
...
+    }
+
+    int16 GetCoverAbsorb(CCharEntity* target)
+    {
+        if (target != nullptr)
+        {
+            CItem* body  = target->getEquip(SLOT_BODY);
+            if (body != nullptr)
+            {
+                int32 bodyID = body->getID();
```
m241dan: ibid, MOD stuff.

scripts/globals/abilities/cover.lua
```diff
@@ -0,0 +1,66 @@
...
+-- Recast Time: 0:03:00
+-- Duration: 0:00:15 - 0:00:30
+-- Info from https://www.bg-wiki.com/bg/Cover
+-----------------------------------
+require("scripts/globals/status")
+require("scripts/globals/msg")
+-----------------------------------
+
+function onAbilityCheck(player,target,ability)
+    if (target == nil or target:getID() == player:getID() or not target:isPC()) then
```
m241dan: Do we need to check to see if the target is in the same party? alliance?

The client will prevent you, but that likely means you can circumvent it by manually doing "/ja Cover name"

src/map/attack.cpp
```diff
@@ -446,6 +446,24 @@ bool CAttack::CheckCounter()
...
 }
 
+bool CAttack::IsCovered()
+{
+    return m_isCovered;
+}
+
+bool CAttack::CheckCover()
+{
+    if (m_attackRound->GetCoverEntity() != nullptr && (m_attackRound->GetCoverEntity()->isAlive()))
```
m241dan: Is cover entity the character doing the cover or the one getting covered?

src/map/enmity_container.cpp
```diff
@@ -452,3 +452,14 @@ bool CEnmityContainer::IsTameable()
 {
     return m_tameable;
 }
+
+void CEnmityContainer::UpdateEnmityFromCover(CBattleEntity* PCoveree, CBattleEntity* PCoverTarget)
```
m241dan: I would rethink these naming conventions. The Cover Target and the Coveree sound like the same person to me.

src/map/entities/battleentity.cpp
```diff
@@ -1499,6 +1520,11 @@ bool CBattleEntity::OnAttack(CAttackState& state, action_t& action)
         // Set the swing animation.
         actionTarget.animation = attack.GetAnimationID();
 
+        if (attack.CheckCover()){
```
m241dan: " { " on next line instead of same line please, if you don't mind~

src/map/utils/battleutils.cpp
```diff
@@ -5624,4 +5634,151 @@ namespace battleutils
...
+    {
+        if (target != nullptr)
+        {
+            CItem* head  = target->getEquip(SLOT_HEAD);
+
+            if (head != nullptr)
+            {
+                int32 headID = head->getID();
+            
+                if (headID == 12515 || // Gallant Coronet
```
m241dan: ibid on mods instead of item IDs

m241dan (changes requested):
Overall, the general idea seems right here (I will add caveat that I did not pull to test, I trust that it works since you checked the boxes).

**Commits**
I would highly recommend breaking this into multiple commits, and trying to set them up in dependency order or execution order, it just makes easier to review. In addition, making white space changes in the same commit as logic changes or feature additions is a no-no in my book, but not sure the project enforces it.

**General Naming**
The naming for certain variables could use some improvement. I have no doubt that you fully understand your code, far better than me, but at a glance CoverAbilityUser and CoverAbilityTarget might be a better convention to stick with.

**Item Mods**
As a policy, I believe we try to avoid referencing item IDs directly. We like creating mods, like MOD::STR and then adding that to an item instead of having to check item IDs for what should add STR etc all over the place.

Any other comments I have are less about your specific work and more about the fact that you're having to work with a lot of clunky legacy code. So, given that, good job!

tankfest:
The style guide specifies that we should only use parenthesis to clarify order of operations.  ( )'s around if-statement conditions can be deleted.

jdsmall25:
Thank you for the feedback @m241dan and @tankfest.  All very good points, I'll work your feedback in. Thanks for your time.

scripts/globals/abilities/cover.lua
jdsmall25: I couldn't recall if you are able to cover someone outside of your party.  Can you cover an alliance member that isn't in your party?

m241dan: Should be able to test using the client. Or looking at the skill using polutils (though I think skills and spells is still broken in polutils)

jdsmall25: I did verify that the client will not let you cover an alliance member.  I also attempted your example and the client won't let you cover someone not in your party either, despite using their name specifically. I can add in additional logic though if you think its necessary.

jdsmall25:
@m241dan I've reworked my code with your feedback.  I found a bug in the process where I didn't re-assign the attack.m_victim which was calculating the damage ratio based on the Cover Ability Target vs the User.

Please let me know if you have any additional feedback. Thanks!

Edit: I forgot to mention that this PR doesn't include calculating the custom cover enmity bonus from MobSkills.  This is simply because I have no idea where to call my new function UpdateEnmityFromCover.  The MobSkill code jumps all over the place and I was having a tough time locating a central spot that made sense.  I'd welcome any advice on where you think this might belong. Thanks!

TeoTwawki (commented):
```lua
    -- SPARE = 972, -- stuff
    -- SPARE = 973, -- stuff
    -- SPARE = 974, -- stuff
```
```cpp
    // The spares take care of finding the next ID to use so long as we don't forget to list IDs that have been freed up by refactoring.
    // 570 through 825 used by WS DMG mods these are not spares.
    // SPARE = 972, // stuff
    // SPARE = 973, // stuff
    // SPARE = 974, // stuff
};
```
Modifier.h and status.lua's mod entries should always match until the project gets a change that stops requiring these duplicate enumerations.

Spares should be updated everytime the spare IDs are in use. There have been several recent PRs and the various contributors are likely to run into the same spares being adjusted and or same modid's being used, so be aware that depending on who gets merged first further adjustments may need to be made [(and staff may have to fix things if they throw things into branches before release)](https://discordapp.com/channels/639659267003252746/700752295826292826/735991238104317992)

Edit: so looking again and its actually got used IDs up till 972 now.. so 972, 973, 974 would be the next 3 spares. Ugh.

scripts/globals/abilities/cover.lua
m241dan: No, if the client is covering both from menu and from command line, we should not need the additional check.

src/map/utils/battleutils.cpp
```diff
@@ -4277,7 +4281,9 @@ namespace battleutils
         else
         {
             damage = HandleSevereDamage(PDefender, damage, true);
-            int16 absorbedMP = (int16)(damage * (PDefender->getMod(Mod::ABSORB_DMG_TO_MP) + PDefender->getMod(Mod::ABSORB_PHYSDMG_TO_MP)) / 100);
+            int16 coverAbsorb = 0;
+            if (IsCovered) coverAbsorb = PDefender->getMod(Mod::COVER_TO_MP); // If successful cover, add PLD cover absorb mod.
+            int16 absorbedMP = (int16)(damage * (PDefender->getMod(Mod::ABSORB_DMG_TO_MP) + PDefender->getMod(Mod::ABSORB_PHYSDMG_TO_MP) + coverAbsorb) / 100);
```
m241dan: I know you worked this into the existing code, so you can take or leave this suggestion, but it might be worth improving upon the readability of this section of code while we're here. Perhaps something like...

```
double dmgToMP = 0.0;

// If original target is being covered and has the cover to MP mod
if (IsCovered)
    dmgToMP += PDefender->getMod(Mod::COVER_TO_MP);

// Handle absorb dmg to mp
dmgToMP += PDefender->getMod(Mod::ABSORB_DMG_TO_MP);

// Handle specific absorb for physical damage only to mp
dmgToMP += PDefender->getMod(Mod::ABSORB_PSYDMG_TO_MP);

int16 absorbedMP = static_cast<int16>(damage * (dmgToMP / 100.0));

```

src/map/utils/battleutils.cpp
```diff
@@ -4312,7 +4318,9 @@ namespace battleutils
         else
         {
             damage = HandleSevereDamage(PDefender, damage, true);
-            int16 absorbedMP = (int16)(damage * (PDefender->getMod(Mod::ABSORB_DMG_TO_MP) + PDefender->getMod(Mod::ABSORB_PHYSDMG_TO_MP)) / 100);
+            int16 coverAbsorb = 0;
```
m241dan: ibid above, BUT, since it is used in two places, perhaps break this logic out into its own function.

```
int16 battleutils::GetMPFromDmg(CBattleEntity* PDefender, int32 damage, bool IsCovered)
{
    double dmgToMP = 0.0;

    // If original target is being covered and has the cover to MP mod
    if (IsCovered)
        dmgToMP += PDefender->getMod(Mod::COVER_TO_MP);

    // Handle absorb dmg to mp
    dmgToMP += PDefender->getMod(Mod::ABSORB_DMG_TO_MP);

    // Handle specific absorb for physical damage only to mp
    dmgToMP += PDefender->getMod(Mod::ABSORB_PSYDMG_TO_MP);

    return (damage * (dmgToMP / 100.0));
}
```

and then the call line would simply be

`int16 absorbedMP = GetMPFromDmg(PDefender, damage, IsCovered);`

m241dan (approved):
This is great work and response to feedback was fantastic.

While I am approving this, I will add that I am approving of the code and am trusting that the author (in checking the two boxes) has manually tested this themselves. If staff would like, I could push this PR to my test server and have my devs take a whack at it (though, there is no timeline on this).

I did leave 2 tightly related "suggestions" but they are less about author's work and more about trying to get author to fix some of our past... indiscretions.

jdsmall25:
@m241dan As requested, refactored the absorb dmg to mp logic.  I altered your approach a little bit by letting the function handle giving mp to the defender.  Let me know if you have any more feedback.

src/map/modifier.h
```diff
@@ -800,8 +803,7 @@ enum class Mod
 
     // The spares take care of finding the next ID to use so long as we don't forget to list IDs that have been freed up by refactoring.
     // 570 through 825 used by WS DMG mods these are not spares.
-    // SPARE = 964, // stuff
-    // SPARE = 965, // stuff
+    // SPARE = 967, // stuff
```
TeoTwawki: I went and checked the other PRs for what the highest IDs should be
```
    // SPARE = 972, // stuff
    // SPARE = 973, // stuff
    // SPARE = 974, // stuff
```
and status.lua should match:
```lua
    -- SPARE = 972, -- stuff
    -- SPARE = 973, -- stuff
    -- SPARE = 974, -- stuff
```

src/map/attack.cpp
```diff
@@ -446,6 +446,26 @@ bool CAttack::CheckCounter()
...
+
+bool CAttack::CheckCover()
+{
+    CBattleEntity* PCoverAbilityUser = m_attackRound->GetCoverAbilityUserEntity();
+    if (PCoverAbilityUser != nullptr && PCoverAbilityUser->isAlive())
+    {
+        m_isCovered = true;
+        m_victim = PCoverAbilityUser;
+    }
+    else {
```
TeoTwawki: style guide specifies braces on newlines. not a huge deal doubt it woudl hold up the pr, but when you fix the spares this would be a quick edit

TeoTwawki (changes requested):

src/map/modifier.h
jdsmall25: Just to be clear, you're saying I just need to update which are the new current spare IDs not change which IDs I'm using?

TeoTwawki: yeah its a matter of keeping them up to date and in sync so that people don't accidentally use twice. we have 4 broken mods right now from it that need fixed. (issue 998)

the next current free mods ID is 972 (I have edited my review comment to note this)

its a pain point for contributors when multiple prs add a mod at the same time, because whoever gets merged first the other has to update their pr and there is no avoiding it.

jdsmall25: Thanks for the clarification, as requested, I've updated them

TeoTwawki (approved):

59blargedy:
merger will just need to be mindful of spares differences.

59blargedy:
hey @jdsmall25 i was trying to resolve the conflicts with recently merged items and think maybe your mods got lost in some of this shuffle. I apologize for the delays and issues during vacations, and would like to merge this asap. Can you:
keep mod spares as 972-974, make sure your cover mods are actually in both files, and switch the duration mod to 967 (and any sql mods)?
Looking at this, I should have merged this before the one that took up your 964 mod, but it was done in order of receipt. Apologies and really excited to have working cover on topaz!

jdsmall25:
@59blargedy I've made the requested changes. I'm unsure why the style checks are failing, there are many lines I never touched flagged as issues.

59blargedy:
> 
> 
> @59blargedy I've made the requested changes. I'm unsure why the style checks are failing, there are many lines I never touched flagged as issues.

You're fine, don't worry about it. Thanks for updating and sorry about that!

59blargedy (approved):

59blargedy:
@jdsmall25  wanted to provide my testing results here. great work! going to be merging this into both release and canary shortly.
- phys damage cover works
- augmented cover with gallant coronet redirects magic, however the spell will still state it is being cast on the original party
- augmented cover with gallant coronet +1 redirects ranged damage as well
- valor surcoat converts 20% of damage to mp but does not do this to tp moves or magic damage. my guess is tp moves need to have cover added in monstertpmoves lua. (potential followup)
- correctly unable to cover beast or avatar pets, was not able to check trusts
- augmented magic cover works on nin and bard spells, as well as enfeebling magic (unsure if this is correct, noting)
- angle seems appropriate - can cover reasonably even on hilly terrain
- save the queen obtains the appropriate latent of 20 acc when cover used
- cover duration is random between 15-30, and capping out with 400 additional vit caps me at the 30 sec max
- cover duration mods work correctly
- putting 5 merits into cover (+20 seconds) consistently gives me 40 seconds, which lets me pass the max duration of merited cover (35 seconds) per old wiki. per bg wiki, merits add 1 second per merits, so im assuming that means past the cap, so will need addressing. job points are a different story down the line (followup needed)
- using cover while having not engaged the mob does not put the pld on the hate list - if you cover, then cover drops and the person you covered originally dies, the coverer has no hate (followup needed, as i assume it would put you on the hate list?)

jdsmall25:
I appreciate the testing/feedback @59blargedy!

> augmented cover with gallant coronet redirects magic, however the spell will still state it is being cast on the original party

I haven't played retail in some time so I wasn't exactly sure when the cover logic should fire.  My assumption is that the mob will start casting on the original target, if the cover ability user is in the proper cover position when the spell finishes casting, it will successfully cover.  At this point though the "Mob starts casting on original target" message has already been displayed.  If cover was evaluated at the start of casting, that means the ability user would only need to be in the proper position when the casting began and we could send the appropriate message.  Afterwards though the ability user could move where ever and still successfully cover.  It made more sense to me to be assessed OnCastFinished to ensure the positioning was correct and thus I implemented the first approach.  Can anyone confirm how retail works?

>valor surcoat converts 20% of damage to mp but does not do this to tp moves or magic damage. my guess is tp moves need to have cover added in monstertpmoves lua. (potential followup)

Along the same vein, Monster TP moves are also not effected by the cover enmity bonus because I wasn't sure where to place that logic centrally.  The only way the absorb logic fires is if the function "battleutils::TakePhysicalDamage" is called which some abilities do use that I believe?

>using cover while having not engaged the mob does not put the pld on the hate list - if you cover, then cover drops and the person you covered originally dies, the coverer has no hate (followup needed, as i assume it would put you on the hate list?)

According to the bg wiki, the job ability does not trigger the effect on use.  I haven't played retail in a long time, so if someone can confirm how this should work I can fix it.

PR #986, opened 2020-08-20
Fixed so that only player pets gain limited TP

m241dan:
**_I affirm:_**

@tankfest


TeoTwawki (approved):
I'm assuming this was confirmed since last nights discussion, so I slaps da approve button

59blargedy (approved):

PR #987, opened 2020-08-20
fov_allow_alliance (conf setting) --> REGIME_REWARD_ALLIANCE (lua setting)

wrenffxi:
**_I affirm:_**

This setting was used in the core for nothing other than being exposed to LUA via checkFovAllianceAllowed binding.

This PR:

* moves the setting into scripts/globals/settings.lua
* makes the setting affect both RoV and GoV.  Requested by Lusiphur.  I can split this into two settings if desired.

### DIFF
--- a/conf/default/map.conf
+++ b/conf/default/map.conf
@@ -74,7 +74,6 @@ ah_list_limit: 7
 exp_rate: 1.0
 exp_loss_rate: 1.0
 exp_party_gap_penalties: 1
-fov_allow_alliance: 1
 
 #Determines Vana'diel time epoch (886/1/1 Firesday)
 # current timestamp - vanadiel_time_epoch = vana'diel time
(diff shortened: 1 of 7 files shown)


TeoTwawki:
> makes the setting affect both RoV and GoV. Requested by Lusiphur. I can split this into two settings if desired.

Unfortunately retail does alliance normally on GoV but not FoV, so a combined setting would give one of them wrong default behavior. Some of the most fun I ever had on retail was mass destroying Bostaunieux Oubliette in an alliance. We had multiple runners go agro half the zone while out 3 sync taru stood near the book. Fun times.

wrenffxi:
Okay, split into two settings, and default them to GOV: true and FOV: false

TeoTwawki (approved):

m241dan (approved):

zach2good (approved):
Yeah, that didn't really have any business being a conf setting. Nice work 👍

59blargedy (approved):

PR #989, opened 2020-08-21
Uragnite mixin

### LINKED ISSUE #985: Uragnites do not go into their shell
**_I have:_**

I believe we need a family mixin for Uragnites to support their "going into shell" behaviour.

https://www.bg-wiki.com/bg/Category:Uragnite

Their TP moves will also need to be updated because they are only supposed to use certain TP moves while in shell vs out of shell.

wrenffxi:
**_I affirm:_**

Creates uragnite mixin and applies to Coralline and Nightmare uragnites.  Mixin includes:

* shell mode, wherein uragnite
    * changes AnimationSub
    * ceases attacking or moving
    * takes greatly reduced damage
    * gains regen
    * uses shell skills
* physical attacks have a chance to push uragnite into shell mode
* uragnite leaves shell mode after some time
* an optional config function that lets you set custom values for most of these things

Fixes #985

### DIFF
--- a/scripts/mixins/families/uragnite.lua
+++ b/scripts/mixins/families/uragnite.lua
@@ -1,0 +1,106 @@
+--[[
+https://ffxiclopedia.fandom.com/wiki/Category:Uragnites
+https://www.bg-wiki.com/bg/Category:Uragnite
+
+Uragnite mob can optionally be modified by calling tpz.mix.uragnite.config(mob, params) from within onMobSpawn.
+
+params is a table that can contain the following keys:
+    inShellSkillList : skill list given to mob when it enters shell (default: 250)
+    noShellSkillList : skill list given to mob when it exits shell (default: 251)
+    chanceToShell    : percent chance to enter shell when hit by a physical attack (default: 20)
+    timeInShellMin   : least time mob can stay in shell, in seconds (default: 30)
+    timeInShellMax   : most time mob can stay in shell, in seconds (default: 45)
+    inShellRegen     : amount of regen mob gets while in shell (default: 50)
+
+Example:
+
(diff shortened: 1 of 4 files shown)


TeoTwawki (approved):

m241dan (approved):
At a glance, this is very well implemented.

I am particularly fond of having the ability to configure the mixin for specific use; great work here.

I'm approving but did not take too deep of a dive in this.

zach2good (approved):
Looks good to me!

m241dan (changes requested):
Actually, I was thinking about this the other night, and when are mixins loaded? On init? On spawn?

I believe this could cause an issue because local vars are zero'd on Spawn.

zach2good:
They are indeed zero'd OnSpawn

TeoTwawki:
right but that doesn't look like a problem here. the mixin is just changing them from the zero they start life with here `mob:addListener("SPAWN", "URAGNITE_SPAWN", function(mob)` and it doesn't look like anything relies on them being nonzero at spawntime before that. it would be pretty silly for us to have a spawn hook in the core that execs before the rest of the spawn code cleans out out vars and mods.

m241dan (approved):
> right but that doesn't look like a problem here. the mixin is just changing them from the zero they start life with here `mob:addListener("SPAWN", "URAGNITE_SPAWN", function(mob)` and it doesn't look like anything relies on them being nonzero at spawntime before that. it would be pretty silly for us to have a spawn hook in the core that execs before the rest of the spawn code cleans out out vars and mods.

Ahhh, after my 5th look, I see my mistake.

Sorry for the hold up @wrenffxi !

Back to approving!

Wiggo32 (approved):

PR #990, opened 2020-08-21
Implement "Adds support job abilities to wyvern"

rude-jerk:
**_I affirm:_**

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -1560,13 +1560,13 @@ tpz.mod =
     COVER_TO_MP                     = 965, -- Converts a successful cover's phsyical damage to MP
     COVER_MAGIC_AND_RANGED          = 966, -- Redirects ranged and single target magic attacks to the cover ability user
     COVER_DURATION                  = 967, -- Increases Cover Duration
-
+    WYVERN_SUBJOB_TRAITS            = 972, -- Adds subjob traits to wyvern
 
     -- The spares take care of finding the next ID to use so long as we don't forget to list IDs that have been freed up by refactoring.
     -- 570 - 825 used by WS DMG mods these are not spares.
-    -- SPARE = 972, -- stuff
     -- SPARE = 973, -- stuff
     -- SPARE = 974, -- stuff
+    -- SPARE = 975, -- stuff
 }
 
 tpz.latent =
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5699,7 +5699,7 @@ int32 CLuaBaseEntity::addJobTraits(lua_State* L)
     
     if (PEntity != nullptr)
     {
-        battleutils::AddTraits(PEntity, traits::GetTraits(lua_tointeger(L, 1)), lua_tointeger(L, 2));
+        battleutils::AddTraits(PEntity, traits::GetTraits((uint8)lua_tointeger(L, 1)), (uint8)lua_tointeger(L, 2));
     }
 
     return 0;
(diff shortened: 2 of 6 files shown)


scripts/globals/status.lua
```diff
@@ -1554,13 +1554,12 @@ tpz.mod =
     RADIAL_ARCANA                   = 969,
     CURATIVE_RECANTATION            = 970,
     PRIMEVAL_ZEAL                   = 971,
+    WYVERN_SUBJOB_TRAITS            = 972, -- Adds subjob traits to wyvern
 
     -- The spares take care of finding the next ID to use so long as we don't forget to list IDs that have been freed up by refactoring.
     -- 570 - 825 used by WS DMG mods these are not spares.
 
     -- SPARE = 958, -- stuff
-    -- SPARE = 959, -- stuff
```
TeoTwawki: these should be updated, not removed. this is also going to conflict with the multiple other prs dealing with mods. Staff needs to sort these out and get a process going where these stay accurate at least on a pr-by-pr basis (conflicts are inevitable, but we shouldn't have inaccurate spares in branch everyone bases PRs on!)

src/map/lua/lua_baseentity.cpp
```diff
@@ -5627,6 +5627,22 @@ inline int32 CLuaBaseEntity::levelRestriction(lua_State* L)
...
+*  Example : player:addJobTraits(tpz.job.WHM, 75)
+************************************************************************/
+
+int32 CLuaBaseEntity::addJobTraits(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 1) || !lua_isnumber(L, 1));
+    TPZ_DEBUG_BREAK_IF(lua_isnil(L, 2) || !lua_isnumber(L, 2));
+
+    battleutils::AddTraits(static_cast<CCharEntity*>(m_PBaseEntity), traits::GetTraits(lua_tointeger(L, 1)), lua_tointeger(L, 2));
```
TeoTwawki: should be battleentity, not charentity. Also unsure if the static cast prevents the possibility that the base entity had a legit pointer but the battleentity still was invalid (we should probably explicitly check the battleentity exists after casting it, this also saves us from checking if it was an NPC someone mistakenlyaimed this at)

Side note: we have mixed of using static cast and dynamic cast all over core, I'm no expert and both seem to work just fine but I thought we're supposed to use dynamic casts for derived classes..and all the other entity types are derived from baseentity ¯\\\_(ツ)\_/¯ can somebody clarify what we -should- be doing and why?

scripts/globals/status.lua
rude-jerk: I removed only the two that are actually not spares. I agree with you that the current situation is not ideal, but I'm not sure why you'd advocate against removing a comment that something that is not a spare is one. I feel like auditing the list in order to find more spares to update two of the three spare comments is outside of the scope of this, though I'm willing to look into it separately.

TeoTwawki: we maintain 3 spares. we don't remove them we just update the numbers.

src/map/lua/lua_baseentity.cpp
rude-jerk: CharEntity was a goof when I rearranged my original bodged test code, whoops.

I think you're probably right on dynamic vs static cast here. Dynamic with a nullptr check should be the safest option.

59blargedy:
looks like this one just needs an additional pass @project-topaz/developer  @project-topaz/reviewer

zach2good (approved):
Very clean, nice binding, looks like it does what it says on the tin! Ready to merge by whoever is willing to be brave and handle shifting IDs etc.

59blargedy:
@rude-jerk do you mind resolving the conflicts in mods and rebasing? got the other ones prior to yours merged in and now there are conflicts. almost done! (and thank you)

rude-jerk:
@59blargedy done!

PR #992, opened 2020-08-21
Cutscene NPC ID and Windurst 3-2 mission

NiQ1:
**_I affirm:_**

1. Changed CS packet first argument to the NPC triggering the CS, which is how retail behaves (before it would set it to the player's char id, which would cause issues in some cases according to Teo). Kept a fallback to player char id if no NPC or NPC ID cannot be determined.

2. Fixed Windurst mission 3-2 (Written in the Stars) - When repeating it will now give the correct mission of obtaining 3 rusty daggers (before it would repeat the initial mission). Gate guards will also give the correct dialog.


scripts/zones/Windurst_Walls/npcs/Zokima-Rokima.lua
```diff
@@ -3,6 +3,7 @@
 --  NPC: Zokima-Rokima
 -- Starts Windurst Missions
 -- !pos 0 -16 124 239
+-- Note: Includes Topaz modifications which may be subject to Topaz license
```
TeoTwawki, line 6: > -- Note: Includes Topaz modifications which may be subject to Topaz license

since that's true for the entire project (anything not stuck under dsp's gpl is) is it really needed to say it in each file?

NiQ1: Well, I mainly wanted to mark the files. Some downstream servers apparently pull without thinking. This was mainly my way of pointing them to actual code which is not regular GPL.

TeoTwawki: maybe we do need to draw more attention to it, esp for servers that still don't pull but instead cherry pick or copy paste code (both of which they should not ever do but I digress). I'd prefer we somehow make it obvious at the commit level though.

TeoTwawki (approved):
looks legit
(we're newline brace ppl in this project but thats a trivial change to make)

zach2good (changes requested):
Logic all looks good, but I have some style requests:
C++: Allman style braces, please
Lua: Check your indents 

Then the real question is how to deal with:
```lua
-- Note: Includes Topaz modifications which may be subject to Topaz license
```

Currently, I don't really know at what level the TPZL applies and what the ramifications are for pullers and cherry-pickers of this code. I'm actively reading and trying to figure it out, but it's going to take some time still (even though you have patiently tried to explain to me at least 3 times). 

After some reformatting, I'm happy for a staff to pull the trigger on this with these warnings in place and we'll figure it out later...

NiQ1:
> Logic all looks good, but I have some style requests:
> C++: Allman style braces, please
> Lua: Check your indents
> 
> Then the real question is how to deal with:
> 
> ```lua
> -- Note: Includes Topaz modifications which may be subject to Topaz license
> ```
> 
> Currently, I don't really know at what level the TPZL applies and what the ramifications are for pullers and cherry-pickers of this code. I'm actively reading and trying to figure it out, but it's going to take some time still (even though you have patiently tried to explain to me at least 3 times).
> 
> After some reformatting, I'm happy for a staff to pull the trigger on this with these warnings in place and we'll figure it out later...

Good old Notepad++ tabs/spaces issue. Please let me know if there's something I missed.
About that comment - Note that it says "may" and does not clarify any further, as you've says that's still undecided. The comment is there just to mark the file so downstream servers know it may be subject to a license different than DSP's GPL. Nothing more.

zach2good (approved):

59blargedy (approved):

scripts/globals/missions.lua
```diff
@@ -804,7 +804,7 @@ function getMissionMask(player)
             if (player:hasCompletedMission(WINDURST, tpz.mission.id.windurst.TO_EACH_HIS_OWN_RIGHT) == false and getMissionRankPoints(player, 10) == 1) then
                 first_mission = first_mission + 1024
             else
-                if (player:hasCompletedMission(WINDURST, tpz.mission.id.windurst.WRITTEN_IN_THE_STARS) == false and getMissionRankPoints(player, 11) == 1) then
+                if (player:hasCompletedMission(WINDURST, tpz.mission.id.windurst.WRITTEN_IN_THE_STARS) == true and getMissionRankPoints(player, 11) == 1) then
```
kamajii7, line 807: I am pretty sure you don't need to add '== true' here. Lua assess the value is true and will continue. Also don't think you need the == 1 since the function getMissionRankPoints() already returns 1 or 0 which evaluates to true / false respectively. I could be wrong and I will let someone else correct me if I am. 

Nothing technically wrong here, just style things.

NiQ1: > I am pretty sure you don't need to add '== true' here. Lua assess the value is true and will continue. Also don't think you need the == 1 since the function getMissionRankPoints() already returns 1 or 0 which evaluates to true / false respectively. I could be wrong and I will let someone else correct me if I am.
> 
> Nothing technically wrong here, just style things.

Regarding the ==true, you're correct of course. I preferred to write it this way just to be constant with other similar lines in the same files.
Regarding your second comment, I did not change the getMissionrankPoints() logic. I added a call to hasCompletedMission(), which returns a Boolean. I'm not very familiar with LUA to be honest so I wasn't even sure if casting Boolean to integer is allowed but I doubt it's not discouraged anyway.

PR #997, opened 2020-08-23
Bashe placeholder

kaincenteno:
**_I affirm:_**

This was noticed by Hiro (Canaria player) unable to proceed with Axe trial.

Thank you Xastranh (for helping with SQL values) and @jarmengaud  (for testing)


jarmengaud:
I have pulled this change on my test server, and verified that you could pop the NM by killing the placeholder lizard.
I also verified that during the NM only uses Baleful Gaze TP move

-Tokenr

Wiggo32 (approved):

sql/mob_pools.sql
```diff
@@ -6625,6 +6625,7 @@ INSERT INTO `mob_pools` VALUES (6579,'Deserter_Draugar','Deserter_Draugar',89,0x
 INSERT INTO `mob_pools` VALUES (6580,'Deserter_Draugar','Deserter_Draugar',89,0x0000E60600000000000000000000000000000000,8,8,3,240,100,0,1,0,1,0,0,0,1,131,0,0,5,1,0,89);
 INSERT INTO `mob_pools` VALUES (6581,'Varkolak','Varkolak',309,0x0000350700000000000000000000000000000000,15,15,7,290,100,0,1,1,0,2,0,0,3992,157,0,0,293,1,0,309);
 INSERT INTO `mob_pools` VALUES (6582,'Wurdalak','Wurdalak',309,0x0000350700000000000000000000000000000000,15,15,7,290,100,0,1,1,0,2,0,0,3992,131,0,0,293,1,0,309);
+INSERT INTO `mob_pools` VALUES (6583,'Bashe','Bashe',174,0x0000480100000000000000000000000000000000,1,1,7,240,100,0,0,0,1,2,0,0,0,3,0,0,0,1,0,1142);
```
59blargedy, line 6628: can grab flags and flag data from here, also is it "the bashe" or "bashe" if just "bashe", then put 32 in the name_prefix field
https://github.com/KnowOne134/NPC-MOB-Logger/blob/557707325c4a9c8a25c8253ced75ca0ca100a120/npclogger/Tables/Sauromugue%20Champaign.lua#L50

59blargedy: do we not need a mob group to set its level? what level is it spawning as? 
neat to know that it is spawning without setting a spawn_type in mob_groups though,honestly!

kaincenteno: @59blargedy  it has a mob_group.sql  lol

it spawns from 54 to 55

```INSERT INTO `mob_groups` VALUES (49,6583,120,'Bashe',0,32,3142,3300,0,54,55,0);```

kaincenteno:
changes made to flag by @Xastranh  =)

sql/mob_pools.sql
Xastranh: @59blargedy It should be "The Bashe" when being targeted, based on your LUA pull (which has the flag at 0) and youtube videos found of the fight that supports this ( https://www.youtube.com/watch?v=j0gqdVlKFCU&start=53 ) 

With the the LUA documentation you provided, we've updated EntityFlags (3->139) and animationsub (0->8) Flag itself is still at 0, I'm unable to pull it, even looking at the packet provided in the LUA through PVLV. Any assistance if this needs fixing, or documentation on what exactly 'flag' does (it's a pretty generic name tbh) would be appreciated.

kaincenteno:
thank you for rebasing and fixing the conflict @Kreidos

PR #1005, opened 2020-08-26
Reworks Eagle Eye Shot calculations

tankfest:
**_I affirm:_**

This is a WIP, and I'd like input or suggestions.

Some of this might only be properly handled in source.  
Still todo:

- add TP per weapon delay - confirmed by rudejerk with two chars on retail: rudejerk08/25/2020
"TP return is based on weapon delay"
- adjust enmity value to "mimimum amount": Nyu08/25/2020 https://forum.square-enix.com/ffxi/threads/23295-May-16-2012-%28JST%29-Version-Update Eagle Eye Shot: Enmity incurred by ranged attacks made with this ability has been lowered to the minimum possible value.
- fix calculation?  it should be 5x a ranged attack (from wiki: Damage of Eagle Eye Shot = Damage of 1 ranged attack * 5.00.), not based specifically on your ranged weapon damage.
- factor in crit hit rate and +% crit damage -- EES can crit, from wikipedia as well... assume it can crit as often as a normal ranged attack after factoring in mods.

### DIFF
--- a/scripts/globals/abilities/eagle_eye_shot.lua
+++ b/scripts/globals/abilities/eagle_eye_shot.lua
@@ -28,30 +28,23 @@ function onAbilityCheck(player, target, ability)
 end
 
 function onUseAbility(player, target, ability, action)
+    local accBonus = 100
+    local damage = 0
+
     if (player:getWeaponSkillType(tpz.slot.RANGED) == tpz.skill.MARKSMANSHIP) then
         action:animation(target:getID(), action:animation(target:getID()) + 1)
     end
-    local params = {}
-    params.numHits = 1
-    local ftp = 5
-    params.ftp100 = ftp params.ftp200 = ftp params.ftp300 = ftp
-    params.str_wsc = 0.0 params.dex_wsc = 0.0 params.vit_wsc = 0.0 params.agi_wsc = 0.0 params.int_wsc = 0.0 params.mnd_wsc = 0.0 params.chr_wsc = 0.0
-    params.crit100 = 0.0 params.crit200 = 0.0 params.crit300 = 0.0
-    params.canCrit = true
-    params.acc100 = 0.0 params.acc200= 0.0 params.acc300= 0.0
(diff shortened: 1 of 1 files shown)


m241dan:
can you cite sources?

my feedback would be to get the math down, this mechanic will hopefully be changing in the next few months.

tankfest:
> 
> 
> can you cite sources?
> 
> my feedback would be to get the math down, this mechanic will hopefully be changing in the next few months.

sources cited

zach2good:
The sources you've cited:

```
Eagle Eye Shot
Enmity incurred by ranged attacks made with this ability has been lowered to the minimum possible value.

[dev1119] The following job abilities have had their effective range extended:
Eagle Eye Shot / Shadowbind / Bounty Shot / Quick Draw
```

Any updates on the ideas behind these changes?

Also, @TeoTwawki ?

zach2good:
Tank has stepped away from the project and we don't have any evidence that this is the correct way to go for damage calcs, so closing the PR

TeoTwawki:
should be fairly easy for a retail ranger to sort out the actual math. if its really 5x the final damage and not 5x some step along the way, then the final damage can only ever be a multiple of 5. no 2001 or 1503 dmg amounts for example would ever be possible. "5x a ranged attack" is pretty ambiguous as to how the multiplier is applied. SE's own phrasing says "five-fold" which they usually reserve for multihit, leading me to think _maybe_ they calculated acc once but calculated 5 damages and added them, rather than a straight 5x(original value) or _maybe_ they did a 5x multiplier on a step prior to the final dmg. I could locate no official sources on the 1,5x multiple being reference either. In teh event we do need to just take a standard ranged attack and 5x the final damage value, we should _just bind the function that already does ranged attacks_  and make use of that.

zach2good:
I've always wanted to level ranger... Guess now I have an excuse

ibm2431:
Is there a difference between ranged attack damage * 5 and a single-hit ranged weaponskill with a ftp multiplier of 5?

PR #1010, opened 2020-08-27
Implements several ring desynths

Aeshur:
**_I affirm:_**

Implements missing desynthesis recipes for multiple level 95 platinum tier goldsmithing rings. I have requested the exact capture data for this in the captures discord without any luck, however someone was able to verify the recipe exists and that ffxiclopedia was incorrectly missing the recipes as I had suspected (hence why it was never added before most likely).

https://i.gyazo.com/f25423b64d284fccf906ef62ed442315.mp4

"Synthesis canceled. That formula is beyond your current craft skill level." indicates that there is indeed a recipe,.
"Combination of materials cannot be synthesized" Would be what it would say if these recipes did not exist.

I believe with 95%+ certainty that these ring recipes will follow the "pattern" all the other high level ring  (gold/orichalcum) recipes have which is as follows:

NQ: x1 ingot 
HQ1: x1 ingot 
HQ2: x2 ingot 
HQ3: Gem returned

Example of a gold ring with this pattern:
https://ffxiclopedia.fandom.com/wiki/Moon_Ring

Example of an orichalcum ring with this pattern:
https://ffxiclopedia.fandom.com/wiki/Flame_Ring


TeoTwawki:
sounds plausible, can always be corrected later if its not the case.

zach2good (approved):
I don't know much about crafting or desynth, but this all seems pretty reasonable

tankfest (approved):
Good job on your first PR Nireya

Wiggo32 (approved):

PR #1012, opened 2020-08-27
Fix Dancing Chains to apply drown only in a 10y radius

tankfest:
**_I affirm:_**

Dancing chains is not supposed to damage players, but it is supposed to apply a high drown effect to players in a 10y radius.

To be fixed: CURRENTLY our elemental hp/tic down debuffs get their stat debuff AND hp/tic debuff from the effect power.  Because the power for this is set at 15 hp/tic, the strength stat down is -33.  I doubt it was ever that strong, and this needs to be fixed eventually.

ALSO, the duration was a complete guess.  Needs verification if anyone can snag that... might as well grab the -str debuff amount too ^.^

### DIFF
--- a/scripts/globals/mobskills/dancing_chains.lua
+++ b/scripts/globals/mobskills/dancing_chains.lua
@@ -1,34 +1,20 @@
 ---------------------------------------------
 --  Dancing Chains
---
---  Description: Additional effect: Drown
---  Type: Magical
---  Utsusemi/Blink absorb: Ignores shadows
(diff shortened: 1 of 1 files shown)


59blargedy (approved):

zach2good (approved):

TeoTwawki:
+1 for that “to be fixed”. We should set dot by power and stats down by subpower just defaulting to current behavior when no subpower is sent. Lets make an issue so we don’t forget it.

PR #1015, opened 2020-08-27
local those variables..

TeoTwawki:
**_I affirm:_**

just doing a bit of cleanup while looking at the probs with avatar skills


m241dan (approved):

Kreidos (approved):
Good stuff

59blargedy (approved):

PR #1020, opened 2020-08-28
Fixes outpost teleport conquest point cost to be retail accurate

tankfest:
**_I affirm:_**

Tested, works well.

conquest.lua had a variable called cp in the local outposts table that was unused.  Retail functions by charging cp equivalent to 1/10 the gil cost for using outposts, so this was an easy cleanup.

image

Thanks to several people on Gold Saucer for bringing this up, and Nireya@GS for targeting the issue.

### DIFF
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -60,25 +60,25 @@ end
 
 local outposts =
 {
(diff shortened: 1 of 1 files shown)


zach2good (approved):
Looks pretty straightforward 👍

PR #1021, opened 2020-08-28
deoffset jug pet abilities

illzbee:
**_I affirm:_**


scripts/globals/abilities/azure_lore.lua
```diff
@@ -0,0 +1,20 @@
+-----------------------------------
```
59blargedy: do you mind putting this in a separate pr? is unrelated.

PR #1031, opened 2020-08-29
Change mobskill blood drain to remove 1 shadow instead of ignoring them

tankfest:
**_I affirm:_**

It appears that 99.9% of giant bats using blood drain should remove a shadow instead of ignoring shadows and hitting:
https://ffxiclopedia.fandom.com/wiki/Asanbosam
"Blood Drain: Modified version that ignores Utsusemi. Single target Drain."


zach2good (approved):

TeoTwawki (approved):
approved but needs to be an issue raised for the other version the nm used

tankfest:
> 
> 
> approved but needs to be an issue raised for the other version the nm used

just added that functionality to this PR

TeoTwawki (approved):

PR #1035, opened 2020-08-29
Add Provoker (Promyvion - Vahzl) behavior and DB changes

mlariat:
**_I affirm:_**

All implemented behavior for Provoker is based on the capture and noted behavior by Jimmayus: https://youtu.be/0fSKlDz5bWU
There didn't seem to be any apparent way to add a dummy ability for its elemental affinity change aside from using the existing "vulture_1" mob skill, so I put those changes in as well.

For the most part, Provoker's behavior matches the elemental affinity change of Kam'lanaut, so its script is largely adapted from Kam'lanaut's.

Furthermore, though Provoker has Enspells currently in Topaz, the capture indicates that Provoker's enspell effect comes from its current elemental affinity. 

Lastly, the double attack and accuracy boost are based on observations that Provoker had a fairly high accuracy against a 75 RDM and double attacked fairly often.


scripts/zones/Promyvion-Vahzl/mobs/Provoker.lua
```diff
@@ -0,0 +1,67 @@
...
+    end
+end
+
+function onAdditionalEffect(mob, target, damage)
+
+    local element = mob:getLocalVar("element")
+
+    if (element ~= 0) then
+        if (element == 1) then
+            return tpz.mob.onAddEffect(mob, target, damage, tpz.mob.ae.ENFIRE, {chance = 1000})
```
TeoTwawki: suggestion:
```lua
function onAdditionalEffect(mob, target, damage)
    local element = mob:getLocalVar("element")
    local index =
    {
        [1] = tpz.mob.ae.ENFIRE,
        [2] = tpz.mob.ae.ENSTONE,
        [3] = tpz.mob.ae.ENWATER,
        [4] = tpz.mob.ae.ENAERO,
        [5] = tpz.mob.ae.ENBLIZZARD,
        [6] = tpz.mob.ae.ENTHUNDER,
        [7] = tpz.mob.ae.ENLIGHT,
        [8] = tpz.mob.ae.ENDARK
    }
    if index[element] then
        return tpz.mob.onAddEffect(mob, target, damage, index[element], {chance = 1000})
    else
        return 0, 0, 0 -- Just in case its somehow not got a variable set
    end
end
```
then we don't have to do all that if/elseif. The else in my sample above can be ripped out if the mob spawns in with the en-effect already active for its element. I do not know if retail does that I have no looked at the capture. if it does we probably want to set a random element at spawn time just so there is no tiny milliseconds gap between its AI trying to swing at you and the variable being set by onMobFight to throw an explainable nil that seemingly can't  be reproduced when you try to reproduce it later :)

scripts/zones/Promyvion-Vahzl/mobs/Provoker.lua
```diff
@@ -0,0 +1,67 @@
+-----------------------------------
+-- Area: Promyvion - Vahzl
+--   NM: Provoker
+-----------------------------------
+require("scripts/globals/status")
+require("scripts/globals/magic")
+-----------------------------------
+
+function onMobInitialize(mob)
+    mob:addMod(tpz.mod.DOUBLE_ATTACK, 25)
```
TeoTwawki: considering wiki claims it hits hard as well as having double attack I'm wondering if maybe one of its jobs should be war, but I'll leave well enough alone for now and say this is fine if we have indications besides the additional effect that its a rdm (wiki will often see one thing that looks like what  job has, and then just declare a mob to be that job.. its often wrong or forgets that mobs can have more than 1)

mlariat: There doesn't seem to be any substantial proof that its job is RDM aside from the misconception that it casts its enspell effects. Unless someone more knowledgeable than I am about mob jobs wants to chime in I think I'll go ahead and change its job to WAR.

59blargedy: https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit#gid=0
I'd go with war/pld with a family bonus of 10% defp per jimmayus/meloetta's data. that would give the double attack.

TeoTwawki: With the job change remember to update the onMobInitialize section. you'd be adding 25 on top of what warrior already gets at this level (setMod instead of addmod would enforce it having only 25, while removing the line would rely on what warrior has native at this lv)

59blargedy (commented):
not sure if you missed teo's comment, but should remove that double attack mod now that war is set as their job. Can then mark this as good to go!

mlariat:
It absolutely slipped my mind! Thanks for letting me know!

scripts/zones/Promyvion-Vahzl/mobs/Provoker.lua
```diff
@@ -0,0 +1,61 @@
...
+    mob:setMobMod(tpz.mobMod.IDLE_DESPAWN, 180)
+    mob:setMobMod(tpz.mobMod.ADD_EFFECT, 1)
+end
+
+function onMobFight(mob, target)
+
+    local changeTime = mob:getLocalVar("changeTime")
+    local element = mob:getLocalVar("element")
+
+    if (changeTime == 0) then
```
zach2good: If there's only simple logic going on, please drop the brackets
```lua
if changeTime == 0 then
    ...
```

Same for the rest of the file etc.

scripts/zones/Promyvion-Vahzl/mobs/Provoker.lua
```diff
@@ -0,0 +1,61 @@
...
+        mob:setLocalVar("changeTime", math.random(2, 3)*15)
+        return
+    end
+    if (mob:getBattleTime() >= changeTime) then
+        local newelement = element
```
zach2good: newelement -> newElement for clarity please

scripts/globals/mobskills/vulture_1.lua
```diff
@@ -0,0 +1,17 @@
+---------------------------------------------
+-- (Vulture 1)
```
zach2good, line 2: I know you inherited this from what was already in the db, but could we change this to be something a little more descriptive? `provoker_2hr` or something would probably work. That change would need to be reflected here, in this file's name and the entry in mob_skills

zach2good (approved):
Welcome to the project! Everything looks pretty good and neat, I've just flagged a few style issues for you to address and then we should be ready to merge. Thanks for the contribution!

scripts/globals/mobskills/vulture_1.lua
59blargedy: @zach2good it's used by other mobs as well and that is what reports out from actionview in captures. changing the name will make finding and using it potentially difficult.

zach2good: Aha, makes sense! Disregard this comment!

scripts/zones/Promyvion-Vahzl/mobs/Provoker.lua
zach2good: You're almost there, there are still some simple logic blocks (like that while loop) that don't need brackets. Then we're good to go

mlariat: Sorry but are there actually any more unnecessary brackets aside from the while loop? I'm not too familiar with which other brackets may be optional in Lua quite yet, and I don't want to make a bunch of separate commits because I missed something else!

scripts/zones/Promyvion-Vahzl/mobs/Provoker.lua
```diff
@@ -0,0 +1,61 @@
...
+    local changeTime = mob:getLocalVar("changeTime")
+    local element = mob:getLocalVar("element")
+
+    if changeTime == 0 then
+        mob:setLocalVar("changeTime", math.random(2, 3)*15)
+        return
+    end
+    if mob:getBattleTime() >= changeTime then
+        local newElement = element
+        while (newElement == element) do
```
zach2good: just me :)

zach2good:
Merging this into it's own feature branch means that it'll get merged into `canary` in the next update (in a few days maybe). It'll then get battle tested for _some length of time_ and then snuck into `release`. Congrats on your first PR!

PR #1036, opened 2020-08-29
Changes to allow Soulsoother automaton cast Healing and buffs on trusts

TheWhaler:
**_I affirm:_**


zach2good:
@TheWhaler I appear to have ruined your day by re-targetting this to `trust` instead of `canary`...

I had a look through and it all seems to be good, thanks for all the cleanup you did! If you just confirm that this has been running OK on your server then I'll clean it up and merge it when I have time.

TheWhaler:
@zach2good all good on retargetting and Its been behaving so far as I would expect and has not caused a crash yet :)

PR #1038, opened 2020-08-30
Correct critical damage multiplier

Aeshur:
**_I affirm:_**

Currently critical damage caps at 2.25 which is incorrect as it should cap at a multiplier of 3. Right now it has the same cap as a maximum pDIF hit, so you can run into weird scenarios where your critical hits will do the same damage or less than a non-critical hit when fighting mobs much lower level than you because of this.

Sources:
https://ffxiclopedia.fandom.com/wiki/Level_Correction_Function_and_pDIF
"For both one- and two-handed melee weapons, a Critical Hit adds 1.0 to intermediate pDIFa value and capped at 3.0 before the x1.00-1.05 randomization is applied."

89826797-c8a84d00-db24-11ea-998d-34c9545b4ddf
(the only thing incorrect on this chart is that the pDIF high should be capping at 2.25 as per the updated information in the top link)

Taken from:
http://ffxiclopedia.wikia.com/wiki/Level_Correction_Function_and_pDIF?oldid=85991

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2335,7 +2335,7 @@ namespace battleutils
 
         if (PAttacker->objtype == TYPE_PC)
         {
-            ratioCap = 2.25f;
+            ratioCap = isCritical ? 3 : 2.25f;
         }
         if (PAttacker->objtype == TYPE_MOB)
         {


59blargedy (approved):
the increase should occur at pdif calc, rather than cratio. the problem is the pdif calc isnt factoring in crit its calculation, not that the cratio cap is wrong. 

I will approve this, but it really should be placed as an issue that pdif calcs are not utilizing crit, arent factored for 1h/2h (they use different pdif calcs) and arent factoring in q and U calcs.

PR #1043, opened 2020-08-30
Fix Seismostomp Mob ability

Aeshur:
**_I affirm:_**

Previously this ability was wiping all shadows if the mob classified as an NM but that's not how it should work. Time Extensions also classify as "Notorious Monsters" so they were wiping all shadows with their Seismostomps. This fix corrects this and makes Seismostomp absorb either 1 or 2 shadows as shown in the video evidence below.

https://ffxiclopedia.fandom.com/wiki/Category:Simulacra
Screenshot_184

https://www.bg-wiki.com/bg/Mu%27Sha_Effigy
"Advisable for DDs to sub nin to absorb Seismostomp with Utsusemi as back to back use of this move can result in death" is on
most of the Dynamis statue NM pages. As well as the NM fought in the video evidence below.

Source video to show it absorbs 1-2 shadows randomly (watch the battle mod log with the NIN tank's shadows):
https://youtu.be/gEyhDflh_2g?t=1517

Thanks to Tankfest for help with this.

### DIFF
--- a/scripts/globals/mobskills/seismostomp.lua
+++ b/scripts/globals/mobskills/seismostomp.lua
@@ -15,30 +15,20 @@ function onMobSkillCheck(target, mob, skill)
 end
 
 function onMobWeaponSkill(target, mob, skill)
-
     local numhits = 1
     local accmod = 1
     local dmgmod = 2.3
 
-    if (mob:isMobType(MOBTYPE_NOTORIOUS)) then
+    if mob:isMobType(MOBTYPE_NOTORIOUS) then
         dmgmod = dmgmod + math.random()
     end
 
     local info = MobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, TP_NO_EFFECT)
-
-    local shadows = info.hitslanded
-
-    -- wipe shadows
-    if (mob:isMobType(MOBTYPE_NOTORIOUS)) then
-        shadows = MOBPARAM_WIPE_SHADOWS
-    end
-
-    local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT, shadows)
-
+    local shadows_removed = math.random(2)
+    local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT, shadows_removed)
     local typeEffect = tpz.effect.STUN
 
     MobPhysicalStatusEffectMove(mob, target, skill, typeEffect, 1, 0, 4)
-
     target:takeDamage(dmg, mob, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT)
     return dmg
 end


scripts/globals/mobskills/seismostomp.lua
```diff
@@ -25,20 +24,16 @@ function onMobWeaponSkill(target, mob, skill)
...
-    local shadows = info.hitslanded
-
-    -- wipe shadows
-    if (mob:isMobType(MOBTYPE_NOTORIOUS)) then
-        shadows = MOBPARAM_WIPE_SHADOWS
+    
+    if math.random(2) == 1 then
+    local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT, MOBPARAM_1_SHADOW)
+    
+    else
```
TeoTwawki: indents were off and I dislike seeing the same var local'd twice so I was about to suggest we just us a variable for the shadows in a conditional instead of copying the function call into conditionals, but then I realized the shadow values are actually 1 and 2 for the MOBPARAM https://github.com/project-topaz/topaz/blob/27b97b7bf30b8336ca5a0a7b9117ce7eed205f18/scripts/globals/monstertpmoves.lua#L27

So this does the same work, but slightly cleaner to evaluate:
```lua
    local numhits = 1
    local accmod = 1
    local dmgmod = 2.3

    if mob:isMobType(MOBTYPE_NOTORIOUS) then
        dmgmod = dmgmod + math.random()
    end

    local info = MobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, TP_NO_EFFECT)
    local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT, math.random(2))
    local typeEffect = tpz.effect.STUN

    MobPhysicalStatusEffectMove(mob, target, skill, typeEffect, 1, 0, 4)
    target:takeDamage(dmg, mob, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT)
    return dmg
end
```

zach2good (approved):

zach2good:
Code looks good, I would maybe have left the shadows var so it's clear what's being passed into the damage calculation, but the end result is the same.

I'll let another staff look this over and confirm its OK

**EDIT:**

My point being is that it's much clearer to lay things out like this, even if it's a little redundant:
```lua
local shadows_removed = math.random(2)
local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT, shadows_removed)
```

Aeshur:
I agree with you zach. I'll likely be looking over other mob abilities in the future and rewriting their shadow behavior to be more retail accurate like this one is now.

TeoTwawki:
> 
> 
> Code looks good, I would maybe have left the shadows var so it's clear what's being passed into the damage calculation, but the end result is the same.
> 
> I'll let another staff look this over and confirm its OK
> 
> **EDIT:**
> 
> My point being is that it's much clearer to lay things out like this, even if it's a little redundant:
> 
> ```lua
> local shadows_removed = math.random(2)
> local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT, shadows_removed)
> ```

@zach2good that was almost what I initially suggested before seeing the global variables for the shadows were literally just 1 and 2 and thought "if I say to do that someone is just going to go 'we don't need a variable for that why even do that Teo' "
which happens pretty much everytime I use a variable like that in chat on discord in my [example scripts](https://discordapp.com/channels/639659267003252746/639670779386134548/745785506633220216)

tankfest:
> 
> 
> Code looks good, I would maybe have left the shadows var so it's clear what's being passed into the damage calculation, but the end result is the same.
> 
> I'll let another staff look this over and confirm its OK
> 
> **EDIT:**
> 
> My point being is that it's much clearer to lay things out like this, even if it's a little redundant:
> 
> ```lua
> local shadows_removed = math.random(2)
> local dmg = MobFinalAdjustments(info.dmg, mob, skill, target, tpz.attackType.PHYSICAL, tpz.damageType.BLUNT, shadows_removed)
> ```

"Hey that's something everyone can enjoy!"

PR #1065, opened 2020-09-05
Bestiary: Dragons

wrenffxi:
**_I affirm:_**

Compared our data against retail captures for mob species within the "Dragons" category of the bestiary. Used both the wiki, and also cross-referencing the database by familyId, to find as many mobs within each species as I could. Did one species per commit for easier reviewing.

According to retails caps, Hydra's AnimationSub varies between 4 and 6, rather than 0 and 2.  The animations look the same, e.g. animation 0 and 4 are both three-headed.

Fixed Feldrautte_I_Rouhent's groupid.


Arklus:
only the linking table was changed?

zach2good (approved):

TeoTwawki:
> According to retails caps, Hydra's AnimationSub varies between 4 and 6, rather than 0 and 2. The animations look the same, e.g. animation 0 and 4 are both three-headed.

That makes me wonder what SE is using the extra bits on the mask for. Most models have 2 states, witha  few having 3. rarely you'll see more like that flying snake in abyssea that does wind wall animation. Especially weird that there are more values being used that it has visuals for.

TeoTwawki (approved):

PR #1067, opened 2020-09-05
Fix conditional logic in doesToolBreak and pickItem

tankfest:
**_I affirm:_**

I was just doing some HELM testing on my test server to ensure the mods were properly scripted for break rates, and I noticed a small issue with comparing the random roll to the player's break rate.

If you have break rate on the server set to 100%, there's a 1/100 chance that you would have NOT broken with the old script.  Similarly, for determining if the player got an item if item chance was set to 100%, there was a 1/100 chance the player would not receive an item.

Now at all setting possibilities, the break rate and item pick rate will function as intended.

### DIFF
--- a/scripts/globals/helm.lua
+++ b/scripts/globals/helm.lua
@@ -1321,7 +1321,7 @@ local function doesToolBreak(player, info)
         roll = roll + (player:getMod(mod) / 10)
     end
 
-    if roll < _G[info.settingBreak] then
+    if roll <= _G[info.settingBreak] then
         player:tradeComplete()
         return true
     end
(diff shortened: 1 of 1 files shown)


TeoTwawki:
this probably happened from this code having been the replacement for code that did 0-99 instead of 1-100 in its random rolls back then. thanks for catching and fixing it :+1:

tankfest:
> 
> 
> this probably happened from this code having been the replacement for code that did 0-99 instead of 1-100 in its random rolls back then. thanks for catching and fixing it +1

yeah that was another way i was looking at "fixing" this issue, but this way seemed cleaner.

TeoTwawki:
> 
> 
> > this probably happened from this code having been the replacement for code that did 0-99 instead of 1-100 in its random rolls back then. thanks for catching and fixing it :+1:
> 
> yeah that was another way i was looking at "fixing" this issue, but this way seemed cleaner.

definitely. I dislike seeing `math.random(0, 99)` when we mean 1-100 for a percentage anyway!

zach2good (approved):
LGTM!

PR #1068, opened 2020-09-05
Increase Serket rage timer to 30 minutes

tankfest:
**_I affirm:_**

Serket rage timer currently defaults at 20 minutes, but it should be 30.
https://ffxiclopedia.fandom.com/wiki/Serket

Thanks to Eren@GoldSaucer for reporting.


zach2good (approved):
This is the pattern used in a bunch of other places, looks good 👍

59blargedy:
where did the 20 minutes come from? is it a mob mod in sql to be removed?

zach2good:
@59blargedy The default in the rage mixin:
https://github.com/project-topaz/topaz/blob/release/scripts/mixins/rage.lua#L19

PR #1070, opened 2020-09-06
Adds mixin for eruca crawlers sleeping habits and regain mod

### LINKED ISSUE #1069: Eruca Crawlers should sleep during the night, and regain TP during firesday
**_I have:_**

From [wikipedia](https://ffxiclopedia.fandom.com/wiki/Category:Crawlers):
"Eruca, with red and black coloring can be found in certain regions on the Aradjiah continent, these Crawlers have special attributes not native to standard Crawlers. They sleep during the night and also regain TP during Firesday."

Thanks to Daemun@GoldSaucer for reporting.

tankfest:
**_I affirm:_**

Fixes #1069 

Thanks to Teo and especially Wren for his urganite mixin work which I used as the skeleton for this mixin.

The regain portion was actually untested because I didn't fancy waiting around for firesday.  The regain amount is a complete guess too.  This might vary from crawler to crawler and the mixin might need another variable for it.

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -2267,7 +2267,8 @@ tpz.mobMod =
     NO_MOVE             = 65, -- Mob will not be able to move
     MULTI_HIT           = 66, -- Mob will have as many swings as defined.
     NO_AGGRO            = 67, -- If set, mob cannot aggro until unset.
-    ALLI_HATE           = 68  -- Range around target to add alliance member to enmity list.
+    ALLI_HATE           = 68, -- Range around target to add alliance member to enmity list.
+    NO_LINK             = 69  -- If set, mob cannot link until unset.
 }
 
 ------------------------------------
--- a/scripts/mixins/families/eruca.lua
+++ b/scripts/mixins/families/eruca.lua
@@ -1,0 +1,97 @@
+--[[
+https://ffxiclopedia.fandom.com/wiki/Category:Crawlers
+
+Eruca mobs can optionally be modified by calling tpz.mix.eruca.config(mob, params) from within onMobSpawn.
+
+params is a table that can contain the following keys:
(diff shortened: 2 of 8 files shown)


tankfest:
Also had to add the NO_LINK mob mod to this PR to disable linking temporarily for the crawlers while asleep.

tankfest:
Last order of business to make this behavior 100% is to determine what the eruca crawlers do when it's sleepy time and they have been pulled to the zone or something.  Right now, they will just fall asleep immediately wherever it was they disengaged and stay there until sun up or re-engage, but they should likely head to their respawn point before sleepy time.

If someone wants to recommend a fix for that I'll include it, otherwise it's good to go as is imo.

ghost:
From Discord (thanks to Apo for checking on retail!):
> if you move them even a couple yalms outside of their spawn point, they despawn and respawn already asleep
if you leave them at their spawn point, logout, come back in without aggroing them, they go back to sleep after exactly 2 mins
until that 2 mins is reached, they wander around their spawn like normal

tankfest:
Waiting on teo's distance function PR so that we can get easy distance from home determination for the crawlers.

tankfest:
It's finished!  Tested with teo's checkDistance fixes as well.

Eruca crawlers will now behave in a retail-like manner when a player disengages them during sleep hours:  

- For servers that allow mobs to roam home if disengaged, the crawlers will roam like normal mobs until they are within 25y of their spawn, at which point a 2 minute timer will put them to sleep again if they remain undisturbed, and it's still night time.
- For servers that despawn mobs that roam too far from spawn, they will depop like normal, and spawn asleep if it's still night.
- For either server, if the crawler is disengaged within 25y of its spawn position, it will idle for 2 minutes and then sleep, if it's still night.

Thanks for testing this on retail ^.^

PR #1072, opened 2020-09-06
Crafting: Various Fixes

Xaver-DaRed:
- Corrected Highest skill checked, so it cannot be the same skill we are currently leveling.
- Added skill down messages.
- Renamed a variable, to make its function more clear
- Changed 0.2+ skill ups level limit to 60 since we got confirmation from retail.
- Removed some magic numbers.

Also, thank you Kreidos for the Git help.

**_I affirm:_**

### DIFF
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -117,7 +117,7 @@ bool isRightRecipe(CCharEntity* PChar)
             uint16 skillValue   = 0;
             uint16 currentSkill = 0;
 
-            for (uint8 skillID = 49; skillID < 57; ++skillID) // range for all 8 synth skills
+            for (uint8 skillID = SKILL_WOODWORKING; skillID <= SKILL_COOKING; ++skillID) // range for all 8 synth skills
             {
                 skillValue   = (uint16)Sql_GetUIntData(SqlHandle,(skillID-49+2));
                 currentSkill = PChar->RealSkills.skill[skillID];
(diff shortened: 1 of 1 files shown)


src/map/utils/synthutils.cpp
```diff
@@ -235,7 +235,7 @@ uint8 calcSynthResult(CCharEntity* PChar)
     double chance  = 0;
     double random = tpzrand::GetRandomNumber(1.);
 
-    for(uint8 skillID = 49; skillID < 57; ++skillID)
+    for (uint8 skillID = SKILL_WOODWORKING; skillID <= SKILL_COOKING; ++skillID)
```
Kreidos, line 238: I like it. :+1:

Kreidos (approved):
Looks good to me. 

I can't say I know everything about the subtleties of removing the configurable 0.1 threshold but otherwise, no objections here. :+1:

Xaver-DaRed:
If ppl want me to make a new configuration just for it, I'll add it, no problem.

Just to be clear, skill ups over 0.1 can still happen is you configure the already existing skill up amount multiplier.

src/map/utils/synthutils.cpp
kaincenteno: that's better, begone magic numbers. (for some reason I'm getting flashbacks about this)

zach2good (approved):
> I can't say I know everything about the subtleties of removing the configurable 0.1 threshold but otherwise, no objections here. 👍

_Same._

Regardless, code looks good and clean. Since I'm still trying to get the new CI to behave I'll run the builds on Win/Linux/32-bit/64-bit and report back, I don't expect any problems.

zach2good:
RE: A new config option: I'm always in favor of more customization options, I'd accept this PR with or without one - your call

Xaver-DaRed:
@zach2good 
I can add customization options in a later PR. This isn't the last crafting related PR I'm going to make anyway, there are still many changes left pending, wich will need customization options to switch between era and current retail.
I'd rather leave this PR as it is and merged whenever possible, becuse this is mainly a bugfix wich was affecting one player from canaria.

zach2good:
Since this is going into your feature branch, I'll merge this right away

PR #1073, opened 2020-09-07
Markdownify and add badges to README.md

zach2good:
**_I affirm:_**

Pretty much just for kicks. Avoided adjusting any license text or adding license badges because I don't want to kick off that discussion right now. That's for another time.

**Change visible here:**
https://github.com/project-topaz/topaz/blob/6e862f6ce2b9ba2d9c1c755b686462bebf8e012d/README.md

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,24 @@
-Project Topaz
-========
+# Project Topaz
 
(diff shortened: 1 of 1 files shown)


Kreidos (approved):
Slick! :+1:

TeoTwawki (approved):

Issue #1062, opened 2020-09-05
Crash on Trust death/despawn in BCNM/Battlefields


kaincenteno:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

from Canaria players, it happens when releasing trusts in BCNM. Tagging @jarmengaud  for more info
```
[02/Sep] [08:51:12][Debug] CZone:: Monastic_Cavern IncreaseZoneCounter <1> Brain 
[02/Sep] [08:51:12][Info] parse: 00D | 0143 0141 04 from user: Yinnyboy
[02/Sep] [08:51:12][Debug] CZone:: Bastok_Markets DecreaseZoneCounter <0> Yinnyboy
[02/Sep] [08:51:12][Info] parse: 011 | 0002 0001 04 from user: Brain
```
```
Thread 1 "topaz_game" received signal SIGSEGV, Segmentation fault.
CStatusEffect::GetFlag (this=0x7573656b01043780)
    at src/map/status_effect.cpp:94
94              return m_Flag;
89          return m_Tier;
90      }
91
92      uint32 CStatusEffect::GetFlag()
---Type <return> to continue, or q <return> to quit---
93      {
94              return m_Flag;
95      }
96
97      uint32 CStatusEffect::GetTickTime()
98      {
#0  CStatusEffect::GetFlag (this=0x7573656b01043780) at src/map/status_effect.cpp:94
#1  0x000055555566abab in CStatusEffectContainer::DelStatusEffectsByFlag (this=0x555586903770, flag=flag@entry=524288,
    silent=silent@entry=true) at src/map/status_effect_container.cpp:648
#2  0x00005555555dce4c in CBattlefield::RemoveEntity (this=0x5555867284f0, PEntity=PEntity@entry=0x555586906d50,
    leavecode=leavecode@entry=3 '\003') at src/map/battlefield.cpp:488
#3  0x00005555555c6010 in CBaseEntity::~CBaseEntity (this=0x555586906d50, __in_chrg=<optimized out>)
    at src/map/entities/baseentity.cpp:55
#4  0x00005555555dbbf3 in CTrustEntity::~CTrustEntity (this=0x555586906d50, __in_chrg=<optimized out>)
    at src/map/entities/trustentity.h:33
[...]

PR #1074, opened 2020-09-07
Trust BCNM death/despawn fix (again)

zach2good:
**_I affirm:_**

Fixes https://github.com/project-topaz/topaz/issues/1062

From discussion in Discord and a bunch of reading and clarifying, we have discovered that `static_cast` always succeeds, leading to bad pointers which will explode when used. It looks like that's whats been going on in this section when trusts are being deleted:

```cpp
    auto entity = static_cast<CBattleEntity*>(PEntity);
    entity->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_CONFRONTATION, true);
    // ^-- entity will be a bad pointer if PEntity is a trust that's in the middle of being deleted....
```

The cast to `CBattleEntity*` fails, but then the resulting pointer is used.

This PR cleans up that section with a `dynamic_cast`, which will return `nullptr` if it fails, skipping the sections that would rely on it.

**I ASSUME** that the `if in battlefield, remove from battlefield` logic is in `CBaseEntity::~CBaseEntity` is there because NPC's (like treasure chests) need it... so it can't be moved up into BattleEntity.

Please sanity check my logic!


zach2good:
Going into my feature branch, _yeet_

PR #1076, opened 2020-09-08
upgrade checkDistance

TeoTwawki:
Now accepts -any- type of entity or positional data as input, instead of being limited to mobs/players.

**_I affirm:_**
- [x] that Tank has _tested my code_ since the last commit in the PR, and I/we/he will test after any later commits

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -1909,21 +1909,64 @@ inline int32 CLuaBaseEntity::clearPath(lua_State* L)
 /************************************************************************
 *  Function: checkDistance()
 *  Purpose : Returns the yalm distance between entities
-*  Example : if (player:checkDistance(pet) <= 25) then
-*  Notes   :
+*  Example1: if player:checkDistance(target) <= 25 then
+*  Example2: if player:checkDistance(pos) <= 25 then
+*  Example3: if player:checkDistance(posX, posY, PosZ) <= 25 then
+*  Notes   : Example1 is an entity, the others are coordinate point inputs
 ************************************************************************/
 
 inline int32 CLuaBaseEntity::checkDistance(lua_State *L)
 {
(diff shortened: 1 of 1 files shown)


TeoTwawki:
@tankfest wanna give this a shot? I may not have time to test all 3 intended uses personally till the weekend, so its a draft until all tested thoroughly with every combination.

src/map/lua/lua_baseentity.cpp
```diff
@@ -1909,21 +1909,64 @@ inline int32 CLuaBaseEntity::clearPath(lua_State* L)
...
+        position_t point;
+
+        if (lua_istable(L, 1))
+        {
+            lua_getfield(L, 1, "x");
+            posX = (float)lua_tonumber(L, -1);
+            lua_getfield(L, 1, "y");
+            posY = (float)lua_tonumber(L, -1);
+            lua_getfield(L, 1, "z");
+            point.z = (float)lua_tonumber(L, -1);
```
tankfest: small issue here, just needs to be swapped to posZ instead of ``point.z``

tankfest (commented):
I fixed the variable to posZ on my test server and all three types of information (an entity, a table, 3 individual variables) being passed through this function worked as intended.  Awesome job ^.^

src/map/lua/lua_baseentity.cpp
TeoTwawki: missed one when I undid something I was trying.

zach2good:
Seems clear enough, would like confirmation that a couple of instances that use it aren't broken, just to be sure.

https://github.com/project-topaz/topaz/search?q=%3AcheckDistance&unscoped_q=%3AcheckDistance

tankfest:
Tested Catapult, Fight, and Old Sabertooth, all worked well.

zach2good (approved):
LGTM after some testing 👍

PR #1079, opened 2020-09-08
Downgrade clang-format rules to v7 from v10

zach2good:
**_I affirm:_**

@Kreidos made a very good point that not everyone has access to bleeding-edge v10 clang-format and the associated rules, so it's safer to stick to an older version.

Tested locally with v10, still works.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,117 +1,12 @@
----
-Language:        Cpp
-AccessModifierOffset: -4
-AlignAfterOpenBracket: DontAlign
-AlignConsecutiveAssignments: false
-AlignConsecutiveDeclarations: false
-AlignEscapedNewlines: Right
-AlignOperands:   false
-AlignTrailingComments: false
-AllowAllParametersOfDeclarationOnNextLine: true
-AllowShortBlocksOnASingleLine: false
-AllowShortCaseLabelsOnASingleLine: false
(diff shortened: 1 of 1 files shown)


Kreidos (approved):
I like it! I do note that it looks like some of the brace-wrapping rules changed between this and the previous, unsure if that was desired.

v7 is available on Debian Stable (buster).
Ubuntu 18.04 LTS only offers v6, but still has support for a few more years. Would be cool if someone running that (maybe @kaincenteno ?) would be able to test, too.

zach2good:
Ah, so thats why v6 is in my mind

kaincenteno:
what choo talking about @Kreidos 

18.04 LTS is running v7
```
bananarama@topaz:~$ gcc --version
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Kreidos:
> what choo talking about @Kreidos
> 
> 18.04 LTS is running v7

Sorry, Clang-format, not GCC. :)
https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=clang-format&searchon=names

Issue #513, opened 2020-04-22
Limbus Tracker


cocosolos:
# QUESTIONS
### **Temenos North**
- Floor 1:
  - Does only the upper Slaughterman need to be killed for chests to spawn, or both, or random? **both**
  - Do chests always spawn at the same spot? If you kill the upper Slaughterman then the lower Slaughterman, do they spawn by the lower one? **spawn by last Slaughterman killed**
  - Do only Dustman open the door to the next floor, or can a Slaughterman? **only Dustman**

- Floor 3:
  - Does only the BRD need to be killed for chests to spawn, or BRD and WHM, or random? **same as floor 1**
  - Do chests always spawn at the same spot? If you kill the BRD then the WHM, do they spawn by the WHM? **same as floor 1**
  - Do only MNK and DRG open the door to the next floor, or can a BRD/WHM? **MNK and DRG**

- Floor 5:
  - Does the door always open on the first Antica killed? **yes**

- Floor 6:
  - Does only the Skulker need to be killed for chests to spawn, or the Skulker and the Charmer, or random? **same as floor 1**
  - Do chests always spawn at the same spot? If you kill the Charmer then the Skulker, do they spawn by the Charmer? **same as floor 1**
  - Does the first Designator or Abductor killed always open the door? **yes**

### **Temenos West**
- Is it possible to get more than 3 chests on 1 floor or multiples of the same chest? **no**

### **Temenos Central 2**
- How does killing the elementals/avatars effect Carbuncles damage taken/resistance? **[link](https://github.com/project-topaz/topaz/issues/513#issuecomment-650861989)**

### **Temenos Central 4 (Ultima)**
- How does Nuclear Waste Elemental Resist Down effect work? Does it last a certain amount of time, or wear off when hit by the next ability, or some combination? **random duration 1-3 mins?**

### **Temenos Central Basement**
- How are coin drops actually determined? (wiki says 1 coin per RR, I experienced 1->2->4 drops.) **it appears to be 1->2->3 for each reraise, with the rare possibility of one or possibly all 3 adding a +1, e.g. 2->2->3, 1->3->3, 1->2->4**
- Which Aerns are actually the time extensions? Is it completely random, random within set groups, or set mobs? **[link](https://github.com/project-topaz/topaz/issues/513#issuecomment-652149187)**

### **NW Apollyon**
- How does killing the normal mobs effect the floor boss? **[link](https://github.com/project-topaz/topaz/issues/513#issuecomment-653620469)**

[...]

PR #1081, opened 2020-09-09
Temenos Aerns

wrenffxi:
**_I affirm:_**

Assigned new jobs to Temenos Aerns, and adjusted their pools to match retail.

Thanks to Nyu for retail captures and determining which job each Aerns is on retail.

Spreadsheet we're using to fix Aerns is [here](https://docs.google.com/spreadsheets/d/1miQYoxtuk20sjWEKB-u97q1Fcvg6l55fjN_LQgnjHTQ/edit#gid=0).


zach2good (approved):
Looks sane to me, AND it won't cause conflicts with anyone else's work. Thats a win-win 🎉

sql/mob_pets.sql
```diff
@@ -1135,18 +1135,21 @@ INSERT INTO `mob_pets` VALUES (16928827, 1, 15); -- Tonberry_s_Elemental
...
--- INSERT INTO `mob_pets` VALUES (16929102, 1, 9); -- Aern_s_Wynav
--- INSERT INTO `mob_pets` VALUES (16929114, 1, 9); -- Aern_s_Elemental
--- INSERT INTO `mob_pets` VALUES (16929115, 1, 9); -- Aern_s_Avatar
+
+-- ------------------------------------------------------------
+-- Temenos Aerns (Zone 37)
+-- ------------------------------------------------------------
+
+INSERT INTO `mob_pets` VALUES (16929055, 1, 14); -- Temenos Aern DRG
+INSERT INTO `mob_pets` VALUES (16929058, 1, 9); -- Temenos Aern BST
```
59blargedy, line 1144: my records show this 58 is a rdm, and 60 is the bst, after 2 runs. 
https://github.com/project-topaz/topaz/issues/513#issuecomment-652149187

sql/mob_spawn_points.sql
```diff
@@ -8031,8 +8031,8 @@ INSERT INTO `mob_spawn_points` VALUES (16929059,'Aerns_Euvhi','Aern\'s Euvhi',97
 INSERT INTO `mob_spawn_points` VALUES (16929060,'Temenos_Aern','Temenos Aern',119,568.000,-0.500,-500.000,0);
```
59blargedy, line 8031: this was the bst on my runs, with 16929058  being the rdm

sql/mob_spawn_points.sql
```diff
@@ -8057,30 +8057,30 @@ INSERT INTO `mob_spawn_points` VALUES (16929085,'Temenos_Aern','Temenos Aern',12
 INSERT INTO `mob_spawn_points` VALUES (16929086,'Temenos_Aern','Temenos Aern',108,560.000,4.500,-404.000,64);
```
59blargedy, line 8057: i had 16929082 as another rdm on my runs

sql/mob_pets.sql
wrenffxi: Hm.  I see that your list in issue 513 says that 16929058 is a RDM, but

1. The NPClogger data attached to 513 agrees with Nyu's cap that 16929058 has model 00005905, which is the axe model used by WAR/RNG/BST; and
2. The Aern's Euvhi with id 16929059 immediately follows this Aern, and can't belong to any other.  Pets always come after their master in the extracts (they don't always immediately follow, but they always come after; never before)

Similarly, 16929060 has model 00005A05, in both Nyu's cap and yours, which is the sword model used by RDM/PLD/DRK.  And no pet follows it in the data.

So I'm inclined to believe the jobs currently in this PR are correct.

sql/mob_spawn_points.sql
wrenffxi: 16929082 is a RDM in this PR too.  (It already was one; this PR didn't change it.)

sql/mob_pets.sql
59blargedy: fair enough, data doesnt lie. thanks for taking a second look!

sql/mob_spawn_points.sql
59blargedy: i did not see it here and still don't, but i'm sure i'm just missing something - was looking where it jumps from 16929081 to 16929086. not a comment on the pr, but the seemingly missing data. where is that spawn??? ill look around.

wrenffxi: I believe this is just a side effect of viewing the diff of a large file on github.  It's only showing you the edited lines plus a few on either side, and leaving out many of the unchanged lines between the edits.

PR #1083, opened 2020-09-09
Fix Curilla's conditions for giving her trust spell

tankfest:
**_I affirm:_**

Fixes a bug reported in discord a bit ago which caused Curilla to give you her spell regardless of whether or not the player has the proper rank or says yes/no to her question.  Now, she will only give the spell at the end of the event if the player has proper rank and says "yes".

Thanks to Nyu for bringing this up.


tankfest:
Nyu informed me that two other NPCs suffer from the same (or similar) problem.  They will be dealt with tomorrow.

zach2good:
Yeah, people are enjoying early access to Ayame, Curilla and Nanaa... so lets ruin their fun 😈

PR #1087, opened 2020-09-09
Fix conditions for earning Nanaa Mihgo and Ayame trusts

tankfest:
**_I affirm:_**

Fixes conditional logic issue pointed out and identified by zach, nyu, and residentevil.

Also makes Nanaa's dialogue make sense depending on player rank/quest completion.

If there are any more, just point them out to me and I'll PR the fix.


zach2good:
I guess the reason I thought it nanaas condition was 1+1+1 was because there are 4 possible responses in that Cs (0, 1, 2, 3). I didn't stop to consider that you could do the amigo quest and the rank 3 mission out of order to get the different combinations.

Anyway, looks good.

zach2good (approved):
Looks good, will merge tomorrow. Or anyone else can hit the button if they are so able.

PR #1089, opened 2020-09-09
Fix Windurst mission 8-2 and improve Rukususu's script

tankfest:
**_I affirm:_**

Adds windurst 8-2 to Rukususu's script and some minor style fixes for him.  The other ring NPCs need their logic (their ring cutscene is repeatable after you get the ring) and script updated to be in line with topaz's style guidelines, will do in a follow up PR.


zach2good (approved):
Man, clang-format really doesn't like your changes, I'll look into this soon I promise!

The changes themselves look good, nice work

Issue #1090, opened 2020-09-09
Expansion zones attempt to perform updates when setting.lua has the expansion disabled.


Dirk-Dieters:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

When loading topaz_game with settings described above, the server reports hundreds of errors as it attempts to implement each survival guide in the world. 

Not all of them, but you get the idea:
![image](https://user-images.githubusercontent.com/35855037/92665808-b99ff080-f2c4-11ea-9cc0-3dca3c000ba5.png)

The guides appear in-game, but when interacting with them, nothing happens on the client end, however, the server reports an additional error:
![image](https://user-images.githubusercontent.com/35855037/92666011-4f3b8000-f2c5-11ea-8e92-9dc8ccd6ecfb.png)

TO REPRODUCE:

in Settings.lua, set the following:
`ENABLE_COP = 0`
...
`RESTRICT_CONTENT = 1`
Launch server

Expected behavior:
Survival guides should either not appear entirely (if they are properly categorized as part of COP), or appear and be functional. (preferably with an additional setting to disable them separately)

Kreidos:
It could be possible to disable manuals with a flag... I can't claim to be an expert, however I would suggest that it seems pretty difficult to imagine turning off COP and leaving all the later stuff on. Content from later expansions would expect and *rely* on NPCs and content from the earlier ones at least "existing". This appears, in fact, to be what is happening here.

Dirk-Dieters:
The same error occurs with all expansions turned off, so it's not something in the later expansions that are expecting the guides to exist, and the guides are even failing to load in vanilla zones, so the problem is deeper than that.

cocosolos:
survival_guide.lua references `tpz.teleport` but doesnt require teleports.lua. My guess is that teleports.lua is loaded by some zone/npc that is flagged as a certain expansion, and if it is blocked from loading, survival_guide.lua can't take advantage of that load and fails to load, then the chain reaction.

tldr; `require("scripts/globals/teleports")`

Kreidos:
Indeed. Adding that require fixes some of the errors, but there are still other files with errors due to missing require statements as well when all expansions are disabled.

Dirk-Dieters:
[...]

PR #1091, opened 2020-09-10
Fix for load errors when expansions are disabled (issue #1090)

Dirk-Dieters:
**_I affirm:_**


Kreidos:
Welcome @Dirk-Dieters ! Thanks for PRing this fix.

If possible, I'd like if we please hold this PR as work-in-progress until the remaining errors/requires are remedied so that we can have a single complete fix. :) Would you be willing to add the necessary commits to your branch, and thus this PR, to facilitate that?

Dirk-Dieters:
Perfectly reasonable, I'll take care of that as soon as I can.

Dirk-Dieters:
All load errors are now fixed in my fork. Summary of changes:
Added require line to survival_guide.lua
Added checks for all failing load attempts for disabled expansion zones

scripts/globals/promyvion.lua
```diff
@@ -13,39 +14,45 @@ tpz.promyvion = tpz.promyvion or {}
...
-    local m = 0
+	if ENABLE_COP == 1 then
+		local m = 0
 
-    for _, v in pairs(ID.mob.MEMORY_RECEPTACLES) do
-        m = math.max(m, v[1])
-    end
+		for _, v in pairs(ID.mob.MEMORY_RECEPTACLES) do
+			m = math.max(m, v[1])
+		end
```
tankfest, line 22: Lot o' tab indents being used in these changed files.  All tabs need to be replaced with 4 spaces.

Notepad++ has this as a settings option so that when you press tab, it gives you 4 spaces instead.

Kreidos:
I'm not sure selectively putting ifs around so many pieces of disparate code is the right approach to solving a problem like this; Yes, the startup errors went away but I feel there's unintended collateral damage here. For one, it still leaves stub functions which do nothing and that could be affecting other functions elsewhere which rely on them without anyone knowing until the script tries to run.

If there are certain script files which are **confirmed** to contain _no viable code_ when a certain expansion is disabled and can be safely ignored then perhaps they can either be removed from the startup, or a single `if` could be used to return immediately and not execute the rest of the file.

It also appears that some files are being inadvertently removed (.gitignore)

scripts/globals/promyvion.lua
Dirk-Dieters: My mistake, I'll fix that.

Dirk-Dieters:
I agree, the kludge is strong. I'll keep testing and watching for any aberrant behavior.

cocosolos:
Instead of touching promyvion.lua, you should just check if COP is enabled in the promyvion Zone.lua files where they call the promyvion init function.

ibm2431:
If an expansion isn't enabled, the zone shouldn't be loaded at all.

conf/login.conf
```diff
@@ -0,0 +1,70 @@
...
+login_auth_port: 54231
+
+#--------------------------------
+#SQL parameters
+#--------------------------------
+
+mysql_host:      127.0.0.1
+mysql_port:      3306
+mysql_login:     root
+mysql_password:  wolfkin
```
TeoTwawki, line 53: yer gonna wanna change your password now. also amend / rebase your commits to not include your confs

TeoTwawki:
> If an expansion isn't enabled, the zone shouldn't be loaded at all.

I almost agreed with this but it creates other problems I don't want to deal with. I need zones to exist unless I explicitly disable them in my zone_settings table.

Dirk-Dieters:
I'm going to start fresh and try again.

PR #1108, opened 2020-09-12
Make finishing move checks less tall

TeoTwawki:
Alternate title: Satisfy Teo OCD

Todo: helper/utils function? The other times we check this we're also modifying the number right in the ability check, which we probably shouldn't be doing.. see Violent Flourish for an example.

**_I affirm:_**

### DIFF
--- a/scripts/globals/abilities/animated_flourish.lua
+++ b/scripts/globals/abilities/animated_flourish.lua
@@ -11,20 +11,13 @@ require("scripts/globals/msg")
 -----------------------------------
 
 function onAbilityCheck(player, target, ability)
-
-    if (player:hasStatusEffect(tpz.effect.FINISHING_MOVE_1)) then
-        return 0, 0
-
-    elseif (player:hasStatusEffect(tpz.effect.FINISHING_MOVE_2)) then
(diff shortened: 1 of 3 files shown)


TeoTwawki:
on one hand, there is probably a tiny performance saving on returning out before hitting another hasStatsusEffect call. On the other hand..Just look at it. It looks like a sin doesn't it? I'm just saying..

zach2good (approved):

zach2good:
I'm having a small argument with myself about this. On one hand - it is optimised and does no additional work than it has to.

On the other hand, it isn't being called _that much_, and we do much more horrifying things per-tick that make this look insignificant (looking at enmity distribution when someone cures... good lord.)

zach2good:
That being said, if you just ramble a couple of lines at me about performance impact (or lack thereof) I'll just merge it 👍

TeoTwawki:
I wasn’t 100% on this myself figured I’d ross noodle and see if it sticks. I thing the way it was is technically more performant by an possibly immeasurably small amount. I just hate if/else towers.

zach2good (approved):
Sold!

PR #1109, opened 2020-09-12
Must zone before obntaining additional sunsand

TeoTwawki:
No more infinity sunsands without zonage: this behavior is retail accurate.

**_I affirm:_**
- [:thinking:] that I've _tested my code_ since the last commit in the PR, and will test after any later commits


zach2good (approved):
Yup, looks good

PR #1111, opened 2020-09-13
self target howl

59blargedy:
changing the use type of howl from enemy to self, hoping to fix recent crashes

**_I affirm:_**


TeoTwawki:
it also need aoe field and distance corrected:
`,1,20.0,2000,1500,1,0,0,0,0,0,0);`

Howl buffs nearby mobs as well.

TeoTwawki:
orc one appears to have not gotten its animation ID as well? I thought cap said 354 when it was checked

PR #1114, opened 2020-09-13
Fix missing require, provide on/off for survival guides

TeoTwawki:
**_I affirm:_**

**Solves module loading errors depicted in #1090**


zach2good (approved):
Looks sane to me. I didn't know we had extended content tags (now that I see VOIDWATCH), I thought they were just for expansions 🤷

TeoTwawki:
it was originally just expansions, but the way @starlightknight wrote it allows for this so we've used it that way several times (hope you don't mind the tag friend)

zach2good (approved):

PR #1115, opened 2020-09-13
Improve Cover performance

zach2good:
**_I affirm:_**

Cover partners are being re-calculated and checked on every attack round, for every entity in a battle - including mobs. 
Mobs can exist in giant parties, so if they're checking for cover all the time things are gonna get bad.

This is a quick fix to limit that re-calculation to just PCs, but there is plenty of room for improvement here.

**Ideally**
- Only recalculate when: the party changes, someone's cover status changes

If anyone sees any more **quick wins** for performance here, let me know so we can ninja this into `release` and `canary`

### DIFF
--- a/src/map/attackround.cpp
+++ b/src/map/attackround.cpp
@@ -48,7 +48,14 @@ CAttackRound::CAttackRound(CBattleEntity* attacker, CBattleEntity* defender)
     m_taEntity = battleutils::getAvailableTrickAttackChar(attacker, attacker->GetBattleTarget());
 
     // Get cover partner
-    m_coverAbilityUserEntity = battleutils::GetCoverAbilityUser(attacker->GetBattleTarget(), attacker);
+    if (attacker->GetBattleTarget()->objtype == TYPE_PC)
+    {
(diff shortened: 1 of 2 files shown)


Kreidos:
I don't care what clang-format says, I still think it looks good. :laughing: 

Simple enough remediation for now. I mentioned on the Discord that this may be causing the server hangs reported on Canaria; due to the cover code using uint8's for its loop counters, and some areas having more than 255 mobs in a party. Understandably, `i` would loop infinitely in such a case. Presumably this fix should obviate that issue, can't have that many players in a party from what I understand.

Kreidos (approved):

zach2good:
```
Simple enough remediation for now. I mentioned on the Discord that this may be causing the server hangs reported on Canaria; due to the cover code using uint8's for its loop counters, and some areas having more than 255 mobs in a party. Understandably, i would loop infinitely in such a case. Presumably this fix should obviate that issue, can't have that many players in a party from what I understand.
```
That is a very interesting bug!

zach2good:
Gonna merge and start proliferating things around, if more fixes come up we can deal with them normally

Issue #1096, opened 2020-09-11
Client version 30200904_2 may have changed behavior of text delivery between client and server


neuromancerxi:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

With client version 30200904_2 it appears that text sent from client to server (and potentially other areas) has changed.

while on client version 30200904_2:

- Attempting to create new character, will fail with error 3322 when entering character name.
- Text sent via client will appear to be correct client side e.g. /l hi will echo `<charactername> hi` on local client side. 
- Messages received from the server will appear on other receiving clients as `<charactername> `.

This may extend to other forms of text (/s /p SERVERMESSAGE, etc).

fishytheuppity:
I am having a similar issue.  players on the 30200904_2 cannot receive the text of /tell.

Players on a previous version work correctly.

PR #1117, opened 2020-09-13
Incoming chat fix

zach2good:
**_I affirm:_**

Entirely thanks to the work of Windower devs.
https://github.com/Windower/Lua/pull/1936/files

Adjust data locations and sizes after the version update

@wrenffxi if you're working on a version update PR, feel free to nab this

**Testing**
- I wasn't getting correct system messages with additem after the update
- Now I am

### DIFF
--- a/src/map/packets/chat_message.cpp
+++ b/src/map/packets/chat_message.cpp
@@ -47,6 +47,6 @@ CChatMessagePacket::CChatMessagePacket(CCharEntity* PChar, CHAT_MESSAGE_TYPE Mes
 
     ref<uint16>(0x06) = PChar->getZone();
 
-    memcpy(data + (0x08), &name[0], name.size());
-    memcpy(data + (0x18), &message[0], buffSize);
+    memcpy(data + (0x08), &name[0], std::min(name.size(), (size_t)0xF));
+    memcpy(data + (0x17), &message[0], buffSize);
 }


TeoTwawki (approved):

TeoTwawki:
looks right to me.

Kreidos:
Nice work :+1: 

There was word on the Discord that character creation was also broken because of this (presumably same reason and fix). If this is true, any chance of getting that tested and rolled in as well? It would be nice to have one commit to fix it all up in one fell swoop. :smiley: 

Edit: disclaimer I haven't tested whether this was true. :detective:

ghost:
~~Confirming that I get an "Error code: POL-0001" when creating a new character (without and with the fix). Message pops up right before logging in (after chosing a nation). Character will still be present in the list but the same error pops up if you try to log in with it afterwards. I don't know if it's clearly related but here it is. Logging with characters created before that is fine.~~

Chat is working again with it tho! Thank you, zach.

[edit]

Thank you, EpicTaru, and sorry, zach!

zach2good:
I'm not sure what packets are doing in login, but I don't think they'll use the packet I changed here. The fix will be the same, just in the packets in login. I'll take a look later

EpicTaru:
That issue with getting the POL-0001 error is because the connection between the lobby server and the game server timing out or something. If you speed thru the character creation process you won't run into that issue. I haven't created enough characters to know exactly how long it takes to not have the connection time out, but I have created enough to know that that error is due to that.

EpicTaru (approved):

zach2good:
Can anyone with an older version test this and make sure it doesn't destroy the server for you? Don't want to merge this and _force_ an update

zach2good:
Wait... by its very nature it isn't breaking anything... never mind!

zach2good:
I forced an update, I think it's better to have a slightly wonky game than a game that doesn't work at all (depending on what version you have / or are stuck with). Will be working hard the next few days to sort out the login issues now that we've figured out chat

hannahpi:
Maybe this is a stupid question but just for clarification: is this related to issue #1096  ?

ghost:
This is not a stupid question and yes it's related to said issue!

PR #1118, opened 2020-09-13
Whats with all the howling? (fixing howl sql fields)

TeoTwawki:
**_I affirm:_**
- [-] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

Compared with the previously posted retail caps, and with the previously working howls.

Set:
aoe = 1
range = 20 yalm
target = self
anim time and prep time = as they were, [which is probably still wrong like 99% of every mobskill in the entire list and has been for years](https://github.com/project-topaz/topaz/issues/309)


zach2good (approved):
Looks good

PR #1119, opened 2020-09-13
Bestiary QC: Luminians, Luminions

wrenffxi:
**_I affirm:_**

Compared our data against retail captures for mob species within the "Luminians" and "Luminions" categories of the bestiary. Used both the wiki, and also cross-referencing the database by familyId, to find as many mobs within each species as I could. Did one species per commit for easier reviewing.


zach2good (approved):
I love all your PRs, but this one is the best because it doesn't add any new mob_pools 😆

PR #1129, opened 2020-09-13
Populate Guild Shops

### LINKED ISSUE #201: Guild Vendors "Buy window" Error Tone or Selling nothing.
<a href="https://github.com/Rigginit"></img></a> **Issue by [Rigginit](https://github.com/Rigginit)**
_Sunday Dec 17, 2017 at 20:01 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 4395_

----

**_I have:_**

- [] checked the commit log to see if my issue has been resolved since my server was last updated
^Fresh install less than 12 hours ago

**_Client Version_** (type `/ver` in game) **.**   30171103_1

**_Server Version_** (type `!revision` in game) **:** f643501963854f1b763aec94c9ee730d1812b1c7

Try to buy something from Guild vendors and "Buy" button gives error tone or a blank sales window. Not all vendors, and I could only narrow it down to find the problematic NPCS have the 4 digit guild ID in the script/zone/****/NPC.  Not all 4 digit NPCs have errors (Cooking Guild NPCs work for example) 

NPCs Ive found (not a complete list, just the few Ive checked):
Northern San d'Oria    - Woodworking Guild    - Chaupire
Windurst Woods         - Clothcraft Guild            -Kuzah Hpirohpon
Windurst Woods         - Bonecraft Guild            -Retto_Marutto

jarmengaud:
**_I affirm:_**

Fixes #201

Remove guild holidays as per retail
Uncomment guild shops items that were removed because an old 30-item limit

There is still work to try to find exact prices from retail (wiki data are very approximative!)
However note that the Mhaura/Selbina/Al zhabi guilds already had most of the restocked items, not commented out

### DIFF
--- a/sql/guild_shops.sql
+++ b/sql/guild_shops.sql
@@ -2,7 +2,7 @@
 --
 -- Host: localhost    Database: tpzdb
 -- ------------------------------------------------------
--- Server version	5.5.5-10.1.22-MariaDB
+-- Server version   5.5.5-10.1.22-MariaDB
 
 /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
 /*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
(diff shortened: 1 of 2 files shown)


kaincenteno:
Can this be instead a setting that can be handled as a global in settings.lua?

I ask cuz oldschool servers might not want to have this current QoL change, right @TeoTwawki ?

TeoTwawki:
Old school servers would have already had to customize this coz it was never retail accurate even for the old system

zach2good:
I know _nothing_ about guilds and crafting, so I can only comment on code and style etc. All looks pretty sane, this is one of the rare cases where I'm in favor of leaving the old code around with a comment saying "this was removed in 2014" in case people want to revert it.

We can make _some_ allowances and settings to help people that want to run era servers, but we can't be expected to do that for every possible thing that's changed between 2006 and now. If _yet another person_ wants to spin up a TOAU-era server, then they can expect to do at least a little bit of work.

Happy to sign off on this if it gets a couple of "yes" votes.

kaincenteno:
Yes

jarmengaud:
Let me clear some confusion: this is not at all a QoL...!
The understanding of what happened, as I understand it:
Before the big crafting change from SE (2014?), there were 2 guild shops per guild in main cities. Selling the same thing.

DSP had an issue with more than 30 items per shop, so they split the big list between those 2 NPCs.

When SE changed the crafting, the left hand side NPC turned into a normal shop, selling the kits + materials depending on your rank.
But for DSP/Topaz it meant there were only 1 guild NPC left. And all the restocked items were missing as they usually were on the left hand side NPC! (thats why we have empty shops at the moment).

The PR consists of putting back the complete original list on the remaining NPC.

If you would like an era-like system, you would have to revert also the left hand side NPC.

zach2good:
It seems like this is fixes something that wasn't era - so it doesn't serve era servers, and isn't retail - so it doesn't serve the goal of matching retail. Seems like a win to me 🤷

PR #1132, opened 2020-09-13
[build-fix] Fix build issue with newer g++

pacampbell:
Fixed an issue where world_pass.gcc failed to compile
on newer gcc due to -Werror=format-truncation=.

**_I affirm:_**

### DIFF
--- a/src/map/packets/world_pass.cpp
+++ b/src/map/packets/world_pass.cpp
@@ -21,13 +21,14 @@
 
 #include "../../common/socket.h"
 
+#include "../../common/fmt/printf.h"
 #include "world_pass.h"
-
+#include <cinttypes>
 
 CWorldPassPacket::CWorldPassPacket(uint32 WorldPass)
 {
-	this->type = 0x59;
-	this->size = 0x12;
+    this->type = 0x59;
+    this->size = 0x12;
 
     ref<uint32>(0x0C) = 10000;       // price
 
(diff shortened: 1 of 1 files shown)


zach2good:
Hello @pacampbell! 
Welcome to the project! Our build system is a horrible Frankenstein's monster of scripts and legacy voodoo, so sadly you'll have to include `fmt` using a relative path from where you're including it. We also have our own assert as `TPZ_DEBUG_BREAK_IF`, please swap to using that.

Otherwise looks sane, thanks for the contribution!

zach2good (changes requested):
Left requests as a regular comment

zach2good (approved):
Looks pretty sane to me, nice work 👍 

As a side note, there is no need to assume uint32_MAX is 10 digit, _it is_: 4294967295
and std::string isn't null-terminated, so it's all good

EDIT: Also, don't worry about the clang-format check, it's been playing up recently :(

pacampbell:
Hopefully got it right this time. Make sure to double check the assert condition. I also added a memset to the data before copying the string into it. It's unclear to me currently if this memory would always be zeroed so I was being safe.

pacampbell:
> As a side note, there is no need to assume uint32_MAX is 10 digit, it is: 4294967295

I feel kinda dirty about this, even though it should be true on all sane platforms. I just wanted to be explicit for someone looking at the code later since it took me a few moments to figure out what was happening here. I can remove the comment if you think it was redundant.

> EDIT: Also, don't worry about the clang-format check, it's been playing up recently :(

Ah, I was wondering what was happening with this. I ran the command a few times and it wasn't changing anything :)

PR #1134, opened 2020-09-13
mount speed conf

TeoTwawki:
**_I affirm:_**

Also included: addition to `!speed` command to report values on use, and extra commenting of source.

**I thoroughly tested that the additions I made work, and that I did not break anything that wasn't already broken. Of things that WERE broken _before I got here(!!)_ I will be opening a detailed issue for in the near future.**

Note worthy: I verified on retail that chocobos/mounts are NOT simply double your speed. They are base 40 to the players 50, but the client doubles their "stride" effectively moving them at 80. 1 point of speed is worth 2 when mounted. Likewise flee is not "double speed" as was always thought, but I will go over that in my issue later.

### DIFF
--- a/scripts/commands/speed.lua
+++ b/scripts/commands/speed.lua
@@ -2,6 +2,7 @@
 -- func: speed
 -- desc: Sets the players movement speed.
 ---------------------------------------------------------------------------------------------------
+require("scripts/globals/status");
 
 cmdprops =
 {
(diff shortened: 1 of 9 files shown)


TeoTwawki:
Realized I said modifier instead of conf, so amended both pr title and commit message.

Modifier was something I tried to do (for that mount quest) but couldn't get working.

scripts/commands/speed.lua
```diff
@@ -15,14 +16,20 @@ function error(player, msg)
...
         return
     end
 
-    -- set speed
-    player:speed( speed )
-
+    -- Inform player and set speed
+    player:PrintToPlayer(string.format("Old Speed: %u", player:speed()))
+    player:PrintToPlayer(string.format("New Speed: %u", speed))
+    player:speed(speed)
```
zach2good, line 34: I... can't believe this is the original binding... would you mind swooping in and changing this to `setSpeed()` or something better?

If it's in 10 million places don't worry about it, it isn't _that_ awful 🙏

zach2good (approved):
Looks sane to me

scripts/commands/speed.lua
TeoTwawki: There were actually a lot of the old bindings that "report current" if no input was given, with no get/set. Pretty much anytime it was a member value instead of a function call..

Do we want 2 bindings? They would both be acting on `m_baseentity->speed`
https://github.com/project-topaz/topaz/blob/release/src/map/lua/lua_baseentity.cpp#L5255

A lot about speed need reworked and I thought I'd leave this alone until we do that. This is the tip of the iceberg and it is all ugly imo.

zach2good: Yeah, just leave it be for now then 👍

PR #1136, opened 2020-09-14
Fix for errors regarding unloaded assets

Dirk-Dieters:
**_I affirm:_**

I tried to make my changes as small as possible, using existing if statements whenever possible.


Dirk-Dieters:
#1090

scripts/zones/Grauberg_[S]/Zone.lua
```diff
@@ -26,9 +27,9 @@ end
 
 function onZoneWeatherChange(weather)
     local npc = GetNPCByID(ID.npc.INDESCRIPT_MARKINGS)
-    if (weather == tpz.weather.WIND or weather == tpz.weather.GALES) then
+    if (weather == tpz.weather.WIND or weather == tpz.weather.GALES) and ENABLE_WOTG = 1 then
```
Comment (tankfest) [inline scripts/zones/Grauberg_[S]/Zone.lua]: all of these = conditional statements need to be =='s

     ``if (weather == tpz.weather.WIND or weather == tpz.weather.GALES) and ENABLE_WOTG == 1 then``

zach2good:
Hi @Dirk-Dieters! 

I'm not sure what you're trying to achieve here? I saw you make a similar PR a little while ago but didn't get around to looking at it. You have this issue open:

https://github.com/project-topaz/topaz/issues/1090

@TeoTwawki is looking in the right direction:

> it occurs to me that servers that don't allow these zones/expansions/npc's should just be setting that zone sport to zero.

Rather than modifying scripts to not explode when a strange expansion combination is in use, we should not be loading or ticking those zones. I would be inclined to put these checks inside core. 

Perhaps on boot, set flags for each zone that's active/inactive, and at runtime check for if that zone is active and should be ticked here:
https://github.com/project-topaz/topaz/blob/release/src/map/zone.cpp#L783

... or Teo might be right in doing something databasey

zach2good:
I spoke to Teo about this a little bit: a combination of disabling expansions through settings flags, setting the restrict content flag and setting the ~~zoneip~~ zoneport to 0 for zones you don't want accessible or being ticked will achieve what you want. (we think)

TeoTwawki:
zoneport, not zoneip. https://github.com/project-topaz/topaz/blob/release/sql/zone_settings.sql#L29

anything not on a port your current mapserver is set to handle never gets loaded so the scripts never execute there.

zach2good:
Seems pretty resolved to me - please reach out on #customisation in Discord if you need any more advice about zones etc. (I didn't know this was possible either!)

PR #1140, opened 2020-09-14
Fix for few RoE subjugation + spoils records not counting

Kreidos:
**_I affirm:_**
- [myeh] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

### DIFF
--- a/scripts/globals/roe.lua
+++ b/scripts/globals/roe.lua
@@ -33,7 +33,7 @@ end
 
 
 checks.mobID = function(self, player, params)    -- Mob ID check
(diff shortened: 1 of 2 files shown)


TeoTwawki (approved):

TeoTwawki:
unrelated to this pr, but this is the line the build is complaining on https://github.com/project-topaz/topaz/blob/889dae56079039f84ce6b4fb627826f7b79e6ad8/src/map/lua/lua_baseentity.cpp#L6817

zach2good:
> Removes C-Style Casts
image

PR #1146, opened 2020-09-17
Resolve shop and NPC trade duplication exploits

ibm2431:
_Reported_ exploits will be investigated and resolved. Fixes will be shared with the entire community - we're not the type to hoard exploit knowledge/fixes.

Two exploits, two resolutions, two commits. I'll post a summary, exact replication steps, and videos later.

**_I affirm:_**

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -3902,9 +3902,14 @@ inline int32 CLuaBaseEntity::addShopItem(lua_State *L)
     uint16 itemID = (uint16)lua_tonumber(L, -2);
     uint32 price = (uint32)lua_tonumber(L, -1);
 
-    uint8 slotID = ((CCharEntity*)m_PBaseEntity)->Container->getItemsCount();
+    CCharEntity* PChar = (CCharEntity*)m_PBaseEntity;
+    uint8 slotID = PChar->Container->getItemsCount();
 
-    ((CCharEntity*)m_PBaseEntity)->Container->setItem(slotID, itemID, 0, price);
+    PChar->Container->setItem(slotID, itemID, 0, price);
+
+    // Players can add items to the shop container during shop interaction,
+    // so track the shop's number of items separately from the container's size.
+    PChar->Container->setExSize(PChar->Container->getExSize() + 1);
 
     return 0;
 }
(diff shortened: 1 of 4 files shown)


ibm2431:
Some people may have gotten an earlier version of the NPC trade fix. You'll _probably_ want this version which is a _touch_ more permissive and lets players use the remaining stack of items after using it in a trade. 😉

zach2good:
Yeah, I was pretty hasty in getting the earlier version into people's hands, that's on me 😅

ibm2431:
**Shop Duplication exploit:**
When an item is appraised by an NPC vendor to potentially be sold, it's added as the last item to the trade container (I'm guessing this was due to difficulties in how the shop appraise/sell packet flow works). Back when shops were limited to 16 items, this item was added in a hardcoded slot (slot 16, so the 17th item), and there was a hardcoded check to prevent purchases from slots > 15. But when shops were expanded to allow more than 16 items, the check was changed to disallow purchase of items in slots higher than the container's size. This unfortunately no longer prevented the exploiter from injecting a packet to buy the items they've placed in the container (by appraising them), as the size of the container was dynamic. Items purchased in this manner would be "bought from" the shop for 1 gil.

<https://www.youtube.com/watch?v=ZojTUBQKIPw>

I attacked this by adding setExSize and getExSize. When vendor shops are generated from lua, the shop will save its original size to the container. Purchases from slots higher than this size are ignored.

**NPC Trade Replication exploit:**
Unlike player-to-player trade windows, the "trade to NPC" window doesn't send intermediary packets when each item is placed in a trade slot. The only time the client sends any packet regarding a "to NPC" trade is when the entire trade window is sent. This packet contains 8 fields for inventory slot IDs, and 8 fields for the quantity being traded. Previously, the only check was to see if the quantity of the item in that inventory slot was enough for the quantity the player claimed to be trading.

The exploit worked by injecting a packet with the same inventory slot ID across multiple trade slots, producing "shadow clones" of the single item you're replicating. This could be used to make NPCs think you're giving them more than you actually are, so an exploiter could complete quests with less than the required items, or inflate the amount of beastmen seals/currency they're trading in.

<https://www.youtube.com/watch?v=wUL6qHc-rg4>

The original solution (in this PR) was to place a lock on items being traded when the trade packet was being processed (when stepping through the trade slots), and then reject the trade if it detected a locked item was being traded (ie: someone trying to use the same slot twice). This turned out to be too extreme (mainly with how some NPCs gave items _back_ to the player), so it was changed to set and check reserves in #1196

ghost:
Why surely you are mistaken. I saw you ask for people to actually report bugs first, which is obviously the same as saying there are no bugs. So you fixing this is imagined because topaz is bug free. I saw it said on the internets so it must be true.

ibm2431:
Fuck off, Wuku.

zircon-tpl:
ibm2431 has said all there is to say on this subject.

The discussion is now closed.

PR #1147, opened 2020-09-17
Crafting Expert limit break (110)

neuromancerxi:
Adds a check in lua_baseentity to check signature of traded item.
General crafting.lua cleanup
Logic clean-up in Guildmaster scripts
	- Uses actual skill level instead of a generic value of 100
	- Compares against cap for rank instead of arbitrary value
	- Now presents Artisan specific text during test/rank up due to proper values
	- Wires up checks for Expert "quests" for each guild
	- Wires up checks for Guildpoint Key Item
	- Wires up trade check for Rank + Quest active + Guild Key Item + Signed Test item
	- Wires up fail condition for trading unsigned item for Expert test.

Tested all 10 Guild NPCs:
	- Each rank at set intervals (8, 18, 28, etc) for test and rank up
	- Tested failures with unsigned Expert Rank item
	- Tested arbitary signed and unsigned items at various stages

What doesn't work (TODO)
	- The cut scene when you have met the conditions to craft the expert test (Quest Flagged + KI)
	- The CS hangs the client when it gets to (presumably) presenting the recipe to the client
	- As a stand in, I've attached the raw text for each event so the player can get the important details
		- Needs to be signed, name of the craft, ec.

With only one item outstanding I didn't want to hold back the rest of this PR since the rest is fully functional.

Thanks to DHoffryn of Nocturnal Souls for his assistance with the cpp bits.

**_I affirm:_**


scripts/globals/crafting.lua
```diff
@@ -26,16 +26,16 @@ guild =
...
+local TI_Alchemy =      {  937,  4157,  4163,   947, 16543,  4116, 16479,  4120, 16609, 10792}
+local TI_Bonecraft =    {13442, 13441, 13323, 13459, 13091, 17299, 16420, 12508, 13987, 11058}
+local TI_Clothcraft =   {13583, 13584, 13204, 13075, 12723, 13586, 13752, 12612, 14253, 11000}
+local TI_Cooking =      { 4355,  4416,  4489,  4381,  4413,  4558,  4546,  4440,  4561,  5930}
+local TI_Fishing =      { 4401,  4379,  4469,  4480,  4462,  4479,  4471,  4478,  4474,  5817}
+local TI_Goldsmithing = {12496, 12497, 12495, 13082, 13446, 13084, 12545, 13125, 16515, 11060}
+local TI_Leathercraft = {13594, 16386, 13588, 13195, 12571, 12572, 12980, 12702, 12447, 10577}
+local TI_Smithing =     {16530, 12299, 16512, 16650, 16651, 16559, 12427, 16577, 12428, 19788}
+local TI_Woodworking =  {   22,    23, 17354, 17348, 17053, 17156, 17054,    56, 17101, 18884}
+local TI_Synergy =      {}
```
zach2good, line 38: Delicious formatting 😍

zach2good (changes requested):
So first off, this is formatted beautifully. Real nice.

What would you think about replacing `signedByTrader` in `lua_baseentity` with `getSignature` inside `lua_item`, and then doing the string comparison back up in the lua? Seems like a saner binding to me.

I'd like someone to weigh in on this:
```
        --[[  Feeding the proper parameter currently hangs the client in cutscene. This may
              possibly be due to an unimplemented packet or function (display recipe?) Work
              around to present dialog to player to let them know the trade is ready to be
              received by triggering with lower rank up parameters.  ]]--
``` 

I don't really know how best to deal with it.

scripts/globals/crafting.lua
TeoTwawki: I think you mean "proper alignment" Zach. :wink:

scripts/globals/crafting.lua
```diff
@@ -115,21 +114,55 @@ function getTestItem(player, npc, craftID)
...
+        ["Abd-al-Raziq"] =      function (x)    TItemList = TI_Alchemy end,
+        ["Peshi_Yohnts"] =      function (x)    TItemList = TI_Bonecraft end,
+        ["Ponono"] =            function (x)    TItemList = TI_Clothcraft end,
+        ["Piketo-Puketo"] =     function (x)    TItemList = TI_Cooking end,
+        ["Thubu_Parohren"] =    function (x)    TItemList = TI_Fishing end,
+        ["Reinberta"] =         function (x)    TItemList = TI_Goldsmithing end,
+        ["Faulpie"] =           function (x)    TItemList = TI_Leathercraft end,
+        ["Mevreauche"] =        function (x)    TItemList = TI_Smithing end,
+        ["Ghemp"] =             function (x)    TItemList = TI_Smithing end,
+        ["Cheupirudaux"] =      function (x)    TItemList = TI_Woodworking end,
```
TeoTwawki, line 126: Just look at that beautiful alignment. :heart:

scripts/globals/crafting.lua
```diff
@@ -115,21 +114,55 @@ function getTestItem(player, npc, craftID)
...
+
+-----------------------------------
+-- getRankKeyItem by craftID
+-----------------------------------
+
+function getRankKeyItem(player,craftID)
+
+    local RankKI = 0
+
+    if craftID == 48 then -- fishing
```
TeoTwawki: could this be a table instead of if/else?

TeoTwawki:
> 
> So first off, this is formatted beautifully. Real nice.
> 
> What would you think about replacing `signedByTrader` in `lua_baseentity` with `getSignature` inside `lua_item`, and then doing the string comparison back up in the lua? Seems like a saner binding to me.
> 
> I'd like someone to weigh in on this:
> 
> ```
>         --[[  Feeding the proper parameter currently hangs the client in cutscene. This may
>               possibly be due to an unimplemented packet or function (display recipe?) Work
>               around to present dialog to player to let them know the trade is ready to be
>               received by triggering with lower rank up parameters.  ]]--
> ```
> 
> I don't really know how best to deal with it.

We're gonna need a retail cap to check out I think. @zach2good 

Edit: and I agree about `getSignature()`

scripts/globals/crafting.lua
neuromancerxi: Removed the if/elseif stack replaced with ENUM (which is all this was doing anyway).

scripts/globals/crafting.lua
Kreidos: Indeed. Very nice.

I am curious however if this really needs to be a switch? There's no actual function code happening here, and although we C++ types like to think of switches as fast, Lua has no actual switch and there's a fair bit of boilerplate that has to happen to set this 'simulacrum of a switch' up for what could be an easy table index.

Edit: I do realize, technically not Tarumancer's code, he just prettied. But it just caught my eye. :sweat_smile:

zach2good:
I've made that binding for you here: https://github.com/project-topaz/topaz/pull/1153
It should make it into `release` fairly soon. You'll then need to update the instances of `local signed = player:signedByTrader(player,0)` with:
```lua
-- find the item in the trade (or for all items in trade)
local signed = item:getSignature() == player:getName() and 1 or 0
-- remember that Lua true is 0, so "and 1 or 0" converts it to true = 1, false = 0
-- or change your logic later in the function, whichever you prefer
```

scripts/globals/crafting.lua
TeoTwawki: clicky resolvy on conversationy~

scripts/globals/crafting.lua
TeoTwawki: :smiley:

scripts/globals/crafting.lua
neuromancerxi: Migrated signature checks to getSignature (thanks again for that @zach2good), purged signedByTrader. As for the TODO item, I'll leave this conversation open (unresolved) until an agreement is reached on the unimplemented portion.

scripts/zones/Bastok_Markets/npcs/Reinberta.lua
```diff
@@ -22,12 +22,14 @@ function onTrade(player, npc, trade)
             player:setSkillRank(tpz.skill.GOLDSMITHING, newRank)
             player:startEvent(301, 0, 0, 0, 0, newRank, 1)
             player:setCharVar("GoldsmithingExpertQuest",2)
+            player:tradeComplete()
```
TeoTwawki: when using tradeComplete or confirming a trade, that should _always_ happen in the events finish code, not inside the trade code where the event started. Here you have the trade complete while the event is running and if the player is disconnected the event code never completed but the item is gone.

neuromancerxi: Moved `tradeComplete()` from `onTrade` to `onEventComplete`

zach2good:
You're almost there, you want to be using `player:setLocalVar` instead of `player:setCharVar`! CharVars end up in the database and you don't need that here :)

zach2good:
I want to get this in for testing so I'm going to merge this into it's own feature branch and then do those changes myself since they're so small and quick. Thanks for the wonderful contribution! :)

zach2good:
So, I had a quick look to see if I could upgrade parts of this to use LocalVars. I don't fully understand what's meant to be going on with these tracking vars.

**Using Reinberta (Goldsmithing) as an example:**
Is your intent after these quests to leave `GoldsmithingExpertQuest` as 2 forever? Is there a quest that could be completed instead?

The `player:setCharVar("GoldsmithingTraded",1)` logic is OK, it just needs to be a LocalVar so it doesn't live forever if the player somehow bails out in the middle of the CS.

Any further changes now need to be pointed at `feature/crafting_expert_lb`, once they're done I can do a midweek merge into `canary` 👍

PR #1150, opened 2020-09-17
Add RoE headers to VC filter for Windowsers

Kreidos:
:crossed_fingers: I hope!

**_I affirm:_**
- [x] that I can't _test my code_ since the last commit.


zach2good:
Will boot up and check tomorrow, but looks sane at a glance, thanks!

zach2good:
Ah, you also need entries for your `Source Files`

zach2good:
I've tried to make this work nicely... but it jus _wont_. It breaks all the other filters and doesn't produce any errors...

So I'd say just hold on until I can finish the transition to CMake and then you won't have to worry about it 🤷

TeoTwawki:
whaaaaaaaaaaat?

I'll take a look tonight. this normally isn't difficult.

PR #1153, opened 2020-09-18
Add getSignature for lua_item

zach2good:
**_I affirm:_**

We mentioned this in the PR and on discord, this will help out @neuromancerxi with https://github.com/project-topaz/topaz/pull/1147

**EDITED TO BE CORRECT**

**Testing Script**
```lua

function onTrigger(player, str)
    local ring = player:getEquippedItem(13)
    printf("sig %s.", ring:getSignature()) -- <-- Note the full stop. Clean string.
end
```

**Steps**
- Craft a bone ring with a _Cyclone Crystal_ (to sign it)
- Run script to get signature

**Results:**
```
[18/Sep] [09:35:54][LUA Script] sig Test.
```

### DIFF
--- a/src/map/lua/lua_item.cpp
+++ b/src/map/lua/lua_item.cpp
@@ -319,6 +319,25 @@ inline int32 CLuaItem::isShield(lua_State* L)
 
     return 1;
 }
+
+inline int32 CLuaItem::getSignature(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PLuaItem == nullptr);
+
+    lua_pushstring(L, (const char*)m_PLuaItem->getSignature());
+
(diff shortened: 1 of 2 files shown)


zach2good:
Now I see that signature is it's own field in the db, so it isn't held in m_extra at all?
EDIT:
I forgot what coco said yesterday:
```
if anyone is wondering about signatures, we store them as strings in the db, then when we load them into memory, we copy that string into the items ex data, since the client uses the ex data to display to the client
```

TeoTwawki:
> getEquippedItem(13)

WRY U NO USE ENUM IN EXAMPLE

:stuck_out_tongue:

TeoTwawki (approved):

PR #1154, opened 2020-09-18
Style Guide 2: Electric Boogaloo

zach2good:
**_I affirm:_**

- Comprehensive instructions and examples for Core and Lua in docs.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,11 +1,18 @@
 ﻿---
-BasedOnStyle: WebKit
+# We base things of WebKit + Allman braces
+# BasedOnStyle: WebKit
 BreakBeforeBraces: Allman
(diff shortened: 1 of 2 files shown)


zach2good:
An update on this:
The clang-format file is somewhat sane, but I've disabled the check in CI because it straight up doesn't work if people have branched off of a point that doesn't have the latest changes. This is why _I_ never had any problems - because I was always bleeding-edge up-to-date.

I'll try and write up the points in the clang-format file as a guide with examples, also some important developer workflow stuff.

ibm2431 (commented):
I would like to say that I approved this Style Guide over Discord. I failed to do so publicly. I apologize if that led to anyone thinking that it was decided by just zach.

@Kreidos

PR #1156, opened 2020-09-18
Ensure no Lua stack mis-management can occur in RoE.

Kreidos:
I've seen no issues in testing, but just so nobody can say it's RoE blowing the Lua stack (ala Canaria reports).

**_I affirm:_**

### DIFF
--- a/src/map/roe.cpp
+++ b/src/map/roe.cpp
@@ -129,6 +129,7 @@ bool event(ROE_EVENT eventID, CCharEntity* PChar, RoeDatagramList payload)
         return 0;
 
     lua_State* L = luautils::LuaHandle;
+    uint32 stackTop = lua_gettop(L);
     lua_getglobal(L, "tpz");
     lua_getfield(L, -1, "roe");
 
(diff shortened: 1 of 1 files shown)


zach2good:
I should probably also do this for trust stuff 👀

Kreidos:
It might be good in future reviews if we suggest that all Lua functions which return no values to Lua clear the stack before returning for good measure? :shrug:

TeoTwawki:
> 
> 
> It might be good in future reviews if we suggest that all Lua functions which return no values to Lua clear the stack before returning for good measure? shrug

we could..just..DO this? the whole time? we could have been wiping it instead of keeping track carefully of how much it went up/down? Up is down and down is up. Existence is meaningless. Everything I knew was wrong. The sky has fallen, it hit the directv dish and my moogle is freaking out that he can't watch the apocalypse in HD.

Kreidos:
> we could..just..DO this? the whole time? ...

I think this is really only an issue with functions called directly from C, as Lua itself should cleanup its own stack upon returning. In fact I'm not entirely sure how Lua would react to having the stack values it passed-in hosed clean by C++ before returning (never tried)... I should perhaps change that here.

Edit: Indeed. I tried and it doesn't always like when you do. I'm going to put a few extra checks to ensure nothing gets thrown out that shouldn't.

zach2good (approved):

PR #1157, opened 2020-09-18
Resolve bazaar item duplication vector

ibm2431:
Thanks to @InoUno for pointing this out.

Explanation at a later date~!

**_I affirm:_**

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -5908,13 +5908,6 @@ void SmallPacket0x106(map_session_data_t* session, CCharEntity* PChar, CBasicPac
             return;
         }
 
-        PItem->setCharPrice(0);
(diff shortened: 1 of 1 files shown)


zach2good:
Already distributed this to the community, going straight in and will end up in canary tomorrow. Thanks!

ibm2431:
**Bazaar Exploit Dupe:**
Basically, a check if a player can afford all quantity of a bazaar item - and then returning if they can't - was accidentally inserted on lines _after_ creating the item and giving it to the buyer. Purchase packet injection from there (to get around the client-side "you don't have enough gil" check).

But here's the video anyway, which is really just an excuse to show off a [FFXI remix album that I like](http://landofvanadiel.com/lov/):
https://www.youtube.com/watch?v=Tpc3sdQH4s0

PR #1164, opened 2020-09-19
Migrate Guild Trade State cvars to LocalVars

neuromancerxi:
Migrates cvar sets and checks from char to local.
Changes ExpertQuest flag at completion from 2 to 0 so it doesn't stick around like Craft Glitter.
ExpertQuest flag is needed to persist the "Quest" status once it's accepted (otherwise you'll constantly get the menu prompt when the CS fires.) To the best of my knowledge, no actual quest exists that can be started/completed in the quest log.

**_I affirm:_**


zach2good (approved):

zach2good:
Looks good, I'll merge this in and then merge your feature into canary 👍

PR #1170, opened 2020-09-20
raise2 for RDM

jarmengaud:
**_I affirm:_**

Missing RDM from list of jobs that can learn Raise II (at level 95)


zach2good:
Yup, looks right to me!

PR #1171, opened 2020-09-20
update Wip/effect rework with latest release code

TeoTwawki:
**_I affirm:_**


TeoTwawki:
its just a pull of release into it, the only conflicts were 2 scripts that had spaces added to their function arguments _and every single add effect item I had already removed because those got touched in a mass replace in release_

PR #1175, opened 2020-09-21
Adoulin outdoor zone settings

jarmengaud:
**_I affirm:_**

Incorrect zone settings for Adoulin outdoors and dungeons.
I have on purpose let the 'allow escape' flag off for dungeons, as it would not work
(more work needed to define escape areas, etc)


zach2good:
Yup, I remember these values from when I did trust zones, looks good to me 👍

PR #1176, opened 2020-09-22
Fix crimson orb mini quest bug

MarianArlt:
The value for `miniQuestForORB_CS` needs to be set conditionally to prevent being issued again while already on the quest. Not doing so resets the variable on characters that are already on the quest and touch the wall of banishing, ultimately messing up the NPC trigger (will trigger the quest anew and hand out a new white orb).

**_I affirm:_**


zach2good:
Congrats on your first PR! Looks pretty good and sane 👍

Issue #1163, opened 2020-09-19
RDM AF2 time check outdated


MarianArlt:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

The code for the Red Mage Artifact quest **Enveloped in Darkness** (RDM AF2) seems to be outdated if one can trust either [ffxiclopedia.fandom](https://ffxiclopedia.fandom.com/wiki/Enveloped_in_Darkness) or [bg-wiki](https://www.bg-wiki.com/bg/Enveloped_in_Darkness). The last step of this quest is an interaction with a ??? in Crawler's Nest which takes two key items and then waits for the next day to turn over to be able to click again and receive the item **Warlock's Boots**.

The code for this quest is written in [/scripts/zones/Crawlers_Nest/npcs/qm8.lua](https://github.com/project-topaz/topaz/blob/release/scripts/zones/Crawlers_Nest/npcs/qm8.lua)

The file still uses a real day wait timer from back in the day for this, when on the wiki sites it says "max 40s" and "re-zone". **I did not create a pull request for this as I wouldn't know the *exact* timing used by current retail.** But from common sense it seems that the timer was simply removed and due to server connectivity it may take some time. Making a re-zone condition is just pretty contrary to the current retail concept it would seem.

The variable is set on line [19-20](https://github.com/project-topaz/topaz/blob/6a1473ca2640d5226c2571b097de305adb171a3e/scripts/zones/Crawlers_Nest/npcs/qm8.lua#L19):
*(btw note how the comment refers to an old unused variable)*

    local cdate = player:getCharVar("theCrimsonTrial_date")
    local realday = tonumber(os.date("%j")) -- %M for next minute, %j for next day

And the check for this is made from line [22-27](https://github.com/project-topaz/topaz/blob/6a1473ca2640d5226c2571b097de305adb171a3e/scripts/zones/Crawlers_Nest/npcs/qm8.lua#L22), specifically on line 24-27:

    if (player:hasKeyItem(tpz.ki.CRAWLER_BLOOD) == true and player:hasKeyItem(tpz.ki.OLD_BOOTS) == true) then
        player:startEvent(4)
    elseif (cprog == 1 and cdate == realday) then
        player:messageSpecial(ID.text.EQUIPMENT_COMPLETELY_PURIFIED)
    elseif (cprog == 1 and cdate ~= realday) then
[...]

PR #1179, opened 2020-09-22
RDM AF2 timeout adjustment

MarianArlt:
This commit adapts the outdated timeout of "until next server day" mentioned in #1163 and turns it into a timeout between 10 and 40 seconds. `math.random(10,40)` was chosen to not have this be a fixed 40s neither an instant return. This is in line with what is said in the current game guides.

The commit also optimizes the functions for this script, corrects erroneous quest naming and reduces the overall usage of variables from 3 hard triggered locals and 2 charVars to 1 conditioned local and 1 charVar.

It is unclear to me though how flushing charVar "needs_crawler_blood" (from previous steps) to 0 means "being unable to start next quest" (see comment original file). Also it is not explicitly commented what quest is being referred to. I tested my commit and if the "next quest" was meant to be RDM AF3 (Peace for the Spirit [same starter NPC]) I can say that it's perfectly working as expected.

**_I affirm:_**
- [x] that I've _tested my code **extensively**_ since the last commit in the PR, and will test after any later commits


scripts/zones/Crawlers_Nest/npcs/qm8.lua
```diff
@@ -14,46 +14,38 @@ function onTrade(player, npc, trade)
...
-        else
-            player:addItem(14093)
-            player:messageSpecial(ID.text.ITEM_OBTAINED, 14093) -- Warlock's Boots
-            player:setCharVar("theCrimsonTrial_date", 0)
-            player:setCharVar("theCrimsonTrial_prog", 0)
-            player:setCharVar("needs_crawler_blood", 2) -- Fixed being unable start next quest
-            player:addFame(SANDORIA, 40)
-            player:completeQuest(SANDORIA, tpz.quest.id.sandoria.ENVELOPED_IN_DARKNESS)
-        end
+        npcUtil.completeQuest(player, SANDORIA, tpz.quest.id.sandoria.ENVELOPED_IN_DARKNESS, {
```
kaincenteno, line 46: does this work without having `require("scripts/globals/npc_util")` header?

scripts/zones/Crawlers_Nest/npcs/qm8.lua
```diff
@@ -14,46 +14,38 @@ function onTrade(player, npc, trade)
...
+            item = 14093, -- Warlock's Boots
+            fame = 40,
+            var = {"envelopedInDarkness_timer", "needs_crawler_blood"}
+        })
     end
-
-end
+end
```
kaincenteno: I know this is not on you but please add an empty line at the end (i know it's just a guideline thing)

kaincenteno (commented):
nice to have to remove those () on the if statements, but not really needed since that was untouched.

scripts/zones/Crawlers_Nest/npcs/qm8.lua
MarianArlt: Funny you mention it, because I tested this extensively. Yes it seems so. Should I rather include it anyways?

kaincenteno: @TeoTwawki  what u think?

MarianArlt: Btw the parentheses were removed because I shortened the superfluity of `:hasKeyItem() == true`

kaincenteno: no need for all if ( ) parenthesis even if it has an `and` or `or` unless it is really needed for order of resolution

if (something) then <- not good accourding to styleguide should be
if something then
^dont get mad at me, get mad at the styleguide xD

MarianArlt: Sorry, what do you refer to? Should I just resolve them all? I'll gladly do so! (I read the style guide but mostly referred to the original file) I happily get rid of any parentheses where applicable :> (Done)

MarianArlt: Since I couldn't find any file where npc_util is not explicitly required I also added it accordingly. Marking all of this as resolved then.

TeoTwawki:
> math.random(10,40) was chosen to not have this be a fixed 40s neither an instant return

We know its not going to be random, so I'd rather we guessed at the time with a good ballpark figure we can adjust later rather than doing a random call. On retail it's _probably_ the length of an in game minute give or take some fraction from lag. The wiki's disagree with each other about the actual conditions now. A consistent 30 seconds seems pretty reasonable compared to JP midnight if you ask me even if we're not positive.

Just thinking here but: We could potentially refactor this to use localvar instead of CharVar as well, since the time period is short and logging back in or zoning would be enough time in the event that didn't clear the wait condition anyway (BG wiki is instruction people to zone instead of wait).

scripts/zones/Crawlers_Nest/npcs/qm8.lua
TeoTwawki: It wasn't a prob in testing was because something else happened to be loaded that already required it, putting it in memory for this script by chance. Always require teh things you use :+1:  (and only the ones you use - have seen ppl just add the kitchen sink before lol)

MarianArlt:
I heard the "we can use localVar" comment from so many people now to the extent that I got confused at first (for I am pretty new to the backend of this project). But by now I think I can safely argue that we can not. This is merely a minimal required wait time. The player can turn in the key items (thereby activating the timer), then walk away from the quest for weeks, log out, log in, log out again, wait for some server restarts and not come back until he feels like it. Have to have the flag when that day comes he decides to come back for his shiny RDM AF boots.

I can do that a fixed +30s no problem. I thought it would be actually most adequate (referring to the "up to 40s") to randomize it a bit. My first change used `os.date("%M")` which can result in anywhere from 1 to 59s which I find even more off, but is probably what SE actually uses. Personally I feel it's pretty useless anyways, would be way less resource expensive to just remove all that crap and return the friggin boots right away...(no big difference to retail anyways when the wiki says "up to 40s" [which sounds like "next minute" indeed]). I'd bet my RDM AF boots that the SE dev at the time simply changed `%j` to `%M` like we are discussing and was off with it to not even bother with rewriting anything (prioritization of [time]investment instead of resources).

Please tell me how to proceed with the exact timeout and I'll happily commit accordingly 😀

TeoTwawki:
> This is merely a minimal required wait time. The player can turn in the key items (thereby activating the timer), then walk away from the quest for weeks, log out, log in, log out again, wait for some server restarts and not come back until he feels like it. Have to have the flag when that day comes he decides to come back for his shiny RDM AF boots.

that doesn't necessarily mean it can't be the other type of var. if we make the condition to continue being our stored value less than something we're comparing it to - maybe not this instance of it but I just want you to know in many cases we can still avoid saving to the database, by checking other requirements and making changing the logic of what exactly our variable is intended to hold.

> Please tell me how to proceed with the exact timeout and I'll happily commit accordingly grinning

I'd just use 30 seconds where you have math.random and call it a day till someone runs through it on retail and tells us exactly what it did (maybe they already did - I have not looked in the [captures discord](https://discord.gg/XSNJnUE) to see)

MarianArlt:
Well I'd really like to see this very example with a localVar to be honest, cause everybody keeps saying. Sadly this is what I can do ^^; I wouldn't mind an edit/comment here. If anything, I'll change to fix 30s and call it a day for today :)

kaincenteno:
> Well I'd really like to see this very example with a localVar to be honest, cause everybody keeps saying. Sadly this is what I can do ^^; I wouldn't mind an edit/comment here. If anything, I'll change to fix 30s and call it a day for today :)

looks good to me as it is, if someone wants to change it to use a localVar, they can after it gets merged :smile_cat:

scripts/zones/Crawlers_Nest/npcs/qm8.lua
TeoTwawki: repeat because for the life of me I cannot type correctly the example in one try >.< so sorry


as for parentheses what kain means is the outmost ones are usually not needed:
```lua
if (condition1 and condition2) then
```
vs:
```lua
if condition1 and condition2 then
```
If things get complicated we'll wrap things in parentheses but otherwise they only are involved in function calls: `entity:doStuff(arguments)`

TeoTwawki (approved):
lgtm

zach2good (approved):
Looks good to me, thanks for the cleanup also!

PR #1188, opened 2020-09-25
Fix THF ability Collaborator cast distance

Aeshur:
**_I affirm:_**

Quick easy fix I noticed when auditing some abilities.

https://www.bg-wiki.com/bg/Collaborator


zach2good:
And for completions sake, for anyone reading:
The 20.6 range (to the target monster) is present in the ability script:
https://github.com/project-topaz/topaz/blob/27b97b7bf30b8336ca5a0a7b9117ce7eed205f18/scripts/globals/abilities/collaborator.lua#L22
This change is for the range to the target player 

Looks good, thanks 👍

zach2good (approved):

PR #1190, opened 2020-09-25
Respawn delay for normal mobs

jarmengaud:
**_I affirm:_**

Retail respawn delay for normal mobs, Garlaige and Crawlers Nest


TeoTwawki:
Please cite sources for the new values, thanks. It looks like you gave 5 min across the board for all, and last I knew that was only correct for some outdoor are mobs not all game wide - dungeon mobs in particular were longer. I also know retail had adjusted them more than once.

Just want to see some statement of “I got 300 from xyz”

jarmengaud:
I tested 5 families in each dungeon. They were all 5min. Even mobs that are not part of any GoV page.
I'm pretty sure this was done accross the board for all non-NM mobs, when GoV was introduced.
You want the list of the mobs that I actually tested?

kaincenteno:
> I tested 5 families in each dungeon. They were all 5min. Even mobs that are not part of any GoV page.
> I'm pretty sure this was done accross the board for all non-NM mobs, when GoV was introduced.
> You want the list of the mobs that I actually tested?

I think what Teo needs is the FFXI forum update that mentions this change.

If thats the case then all areas should have 5 min pop, that'd be awesome for leveling up =)

jarmengaud:
damn Kain you realize this change is probably like 6-7 years ago lol

jarmengaud:
Found this old discussion here:
https://forum.square-enix.com/ffxi/archive/index.php/t-10269.html?s=4e177444d020fa7d4d45328406c31160

where the switch from 16 min to 5 min spawn rate is mentioned.
Could not find the official one though

TeoTwawki:
thaannk yooooouuuu :star: 

"I tested personally on retail" or "this webpage -linky-" or "SE says -linky-" just anything so we know where the new data comes from.

TeoTwawki:
p.s. in the future, prolly a good idea to put source right in the commit message, as it stops future editors who aren't in the loop from mistakenly undoing your changes. Teo has been there and (re)done that after ppl thought Teo's change was wrong coz old arse wiki page says this or that.

zach2good:
From that forum post:
> I will say though, the 5 min respawn on mobs was a very welcome adjustment. I find getting coffer keys in these zones is easier now cause the mobs spawn faster.

Seems fair to me 👍

zach2good (approved):

PR #1196, opened 2020-09-26
Reduce lockdown level on NPC trades

ibm2431:
Previous resolution to the NPC trade replication exploit may be overzealous in locking down trade items, preventing players from exchanging items when they legitimately should be able to.

So here is an updated version which resolves some reported issues, while offering the same level of protection!

I'll be posting the specifics of the exploits in #1146 on Monday (2020/09/28), so make sure you're patched!

**_I affirm:_**


zach2good:
**DON'T PANIC** This is the fix we've already distributed to server owners! If you've followed the previous instructions you'll already have this.

TeoTwawki (approved):

PR #1198, opened 2020-09-27
Add new message type for possible exploit attempts 

TeoTwawki:
Note: seeing this in your log doesn't guarantee an exploit was being performed.

*Someday* all of these message functions should be changed into just one function with an additional parameter at the beginning telling us which type it is, rather than identical code repeated for each message type. I didn't have time for that right now though.

**_I affirm:_**

### DIFF
--- a/src/common/showmsg.cpp
+++ b/src/common/showmsg.cpp
@@ -646,6 +646,9 @@ int _vShowMessage(MSGTYPE flag, const std::string& string)
         case MSG_ACTION: // Bright White  (mostly useless "player did this" info)
             strcat(prefix, CL_WHITE"[Action Info]" CL_RESET);
             break;
+        case MSG_EXPLOIT: // Bright Red (Detected a likely exploit)
+            strcat(prefix, CL_RED"[Error]" CL_RESET);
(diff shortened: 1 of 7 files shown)


src/common/showmsg.cpp
```diff
@@ -646,6 +646,9 @@ int _vShowMessage(MSGTYPE flag, const std::string& string)
         case MSG_ACTION: // Bright White  (mostly useless "player did this" info)
             strcat(prefix, CL_WHITE"[Action Info]" CL_RESET);
             break;
+        case MSG_EXPLOIT: // Bright Red (Detected a likely exploit)
+            strcat(prefix, CL_RED"[Error]" CL_RESET);
```
zach2good: Should this be `Exploit` instead of `Error`?

zach2good (approved):
Looks good, I'm looking forward to plugging this into my current exploit patches. Just had that one point about the logging title, then good to merge 👍

src/common/showmsg.cpp
TeoTwawki: good catch

PR #1200, opened 2020-09-27
Retail output on EXP ring error

### LINKED ISSUE #1197: Conquest point EXP ring error messages
**_I have:_**

When trying to buy any EXP ring from a gate guard, and one is already found in possession of the player it can not be obtained by default. An error message is shown (see example for Port Windurst below): "**You cannot obtain the [x] band. Come back after sorting your inventory.**" This error message seems extremely inadequate. Is this retail accurate to tell the player to sort his inventory when checked for possession of another item?!
https://github.com/project-topaz/topaz/blob/249b7e54ede2c6cfae73c7c3ce7df83e408567b9/scripts/globals/conquest.lua#L836
https://github.com/project-topaz/topaz/blob/249b7e54ede2c6cfae73c7c3ce7df83e408567b9/scripts/zones/Port_Windurst/IDs.lua#L39

Just as weird appears the message printed when the current conquest tally is not yet reset and the player wants to buy another ring (which he can not):
https://github.com/project-topaz/topaz/blob/249b7e54ede2c6cfae73c7c3ce7df83e408567b9/scripts/globals/conquest.lua#L844
I don't know where to find these ID strings (would be 11952 in this case; are they client side?) but it simply prints:
"**You do not meet the requirements to purchase the [x] band**"
Which also is extremely uninformative and does not seem like something you would experience on retail.

- Try to buy an EXP ring from a gate guard while already owning another/same one; I have no idea what the expected message would be, but I doubt it's the one generated.
- Also try to buy an EXP ring directly after dropping one during the same conquest tally period; again same as before.

MarianArlt:
**_I affirm:_**

This would fix #1197

I created a retail trial account to check this. The guards are apparently extremely repetitive. When choosing an EXP ring from the guards inventory he/she will trigger three descriptive strings. (special nature > possible recharge > great war)
Then if you choose to buy it and one of the conditions discussed in #1197 triggers, the same strings will again be called and printed to the chatlog including a preceding "you don't meet the requirements"-string. This applies to both, trying to buy with another ring in inventory and trying to buy before tally reset. See the screenshot (the first time the three strings are called, they're not logged):

img_20200927_002300

The commit is a little bit less superfluous and only prints:

> **You do not meet the requirements to purchase [item].**
> **Due to its special nature, you can only purchase or recharge [item] once until the conquest results tally is performed. Also, you cannot purchase this item if a similar item is already in your possession.**

### DIFF
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -833,7 +833,8 @@ local function canBuyExpRing(player, item)
     if ALLOW_MULTIPLE_EXP_RINGS ~= 1 then
         for i = 15761, 15763 do
             if player:hasItem(i) then
-                player:messageSpecial(text.ITEM_CANNOT_BE_OBTAINED, item)
+                player:messageSpecial(text.CONQUEST + 60, 0, 0, item) -- You do not meet the requirements to purchase the <item>.
+                player:messageSpecial(text.CONQUEST + 50, 0, 0, item) -- Due to its special nature, you can only purchase or recharge <item> once until the conquest results tally is performed. Also, you cannot purchase this item if a similar item is already in your possession.
                 return false
             end
         end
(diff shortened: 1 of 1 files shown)


zach2good (approved):
Looks good to me!

PR #1201, opened 2020-09-27
Trust: HP/MP nerfs, SJ tweaks & Trust: Prishe

zach2good:
**_I affirm:_**

**Done**
- Dropped the HP/MP base for growth for Trusts, allowed trusts to get MP based off their subjobs.
- Take Trust SJs into account when seeing if they can cast a spell
- Add Trust: Prishe. Luckily, her moves are already programmed as they are used at the end of COP, so just has to assign them.

...she'd not very good.. Oof.

**Not in this PR**
- Getting the cipher
- The cutscenes and associated TrustMemory arguments for when you receive Prishe's cipher.

### DIFF
--- a/src/map/utils/trustutils.cpp
+++ b/src/map/utils/trustutils.cpp
@@ -343,11 +343,23 @@ void LoadTrustStatsAndSkills(CTrustEntity* PTrust)
     uint8 jobHPGrade = grade::GetJobGrade(mJob, 0);
     uint8 jobMPGrade = grade::GetJobGrade(mJob, 1);
 
+    uint8 sjobHPGrade = grade::GetJobGrade(sJob, 0);
+    uint8 sjobMPGrade = grade::GetJobGrade(sJob, 1);
+
     float hpGrowth = 1.0f + ((7.0f - (float)jobHPGrade) / 100.0f);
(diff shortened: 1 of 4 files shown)


src/map/utils/trustutils.cpp
```diff
@@ -343,13 +343,13 @@ void LoadTrustStatsAndSkills(CTrustEntity* PTrust)
     uint8 jobHPGrade = grade::GetJobGrade(mJob, 0);
     uint8 jobMPGrade = grade::GetJobGrade(mJob, 1);
 
-    uint8 sjobHPGrade = grade::GetJobGrade(sJob, 0);
+    //uint8 sjobHPGrade = grade::GetJobGrade(sJob, 0);
```
TeoTwawki, line 346: ugh. I thought this was refactored out. They should be built from mob family system. No wonder you've had issues they are still being built like players and that's never going to align with retail. This was a part of my originally holding back the merge on DSP.

zach2good: Yeah, this is the _last_ set of tweaks before I rewrite everything properly. I just wanted to bring trusts down to reasonable and non-facerolly levels of hp/mp

PR #1202, opened 2020-09-27
Add support for configuring from system environment

zach2good:
**_I affirm:_**

Salvaged from Dan's PR: `Beginning support for Ansible, Docker, and Jenkins #867`

This will aid in letting people automate the different processes inside containers, otherwise, it won't affect anyone at all (That is of course, unless you have these things defined on your regular system for _something else_, then things will get weird.)

**TODO**
Test it

### DIFF
--- a/src/login/login.cpp
+++ b/src/login/login.cpp
@@ -515,13 +515,13 @@ void login_config_default()
 
 void login_config_read_from_env()
 {
-    login_config.mysql_login     = std::getenv("DB_USER") ? std::getenv("DB_USER") : login_config.mysql_login;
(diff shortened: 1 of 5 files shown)


zach2good:
For later reference:
https://dev.mysql.com/doc/refman/5.6/en/environment-variables.html
https://mariadb.com/kb/en/mariadb-environment-variables/

m241dan:
Would it be inappropriate for me to approve this PR?

zach2good:
My main source of decision paralysis here has been "what if someone else has these things already defined, and then their stuff will fail at random and it'll be nearly impossible to debug with them".

This can be solved by prepending "TPZ_" onto the front of each var, so I'll do that later

EDIT: The goal being: Can someone define these by hand, or have them defined when spinning up a container, so they don't have to modify any conf files. Automatically modifying our conf files ... is hard, so we want to avoid that

PR #1203, opened 2020-09-27
Incorrect IDs for PH and NM Antican Praefectus

jarmengaud:
**_I affirm:_**

Incorrect IDs for PH (Antican Princeps) and lottery NM Antican Praefectus


zach2good (approved):
Yup, checked in the db, these are the right values. Thanks! 👍

PR #1206, opened 2020-09-28
Straighten Moogles away from their lives of crime

ibm2431:
I know Moogles need to make a living, but trafficking in duplicates of adventurers' swimsuits is not the way!

**_I affirm:_**


zach2good (approved):
They must repent!

image

ibm2431:
More descriptive description: Porter Moogles use their own special trade function which wasn't hit in #1196

zach2good:
The problem with patching holes like this is that, since our systems are so _jank_, cases like this are bound to pop up. I'm not looking forward to these sorts of shenanigans when I roll out packetguard...

drlongnecker:
Thanks for the quick help on this! =]

ibm2431:
Teamwork makes the dream work!

PR #1211, opened 2020-09-28
Add JST timekeeping functions to vanatime

Kreidos:
This adds functions to retrieve JST-based time. It is agnostic to the local timezone setting, only expecting the clock be set correctly for their own local timezone.

Should greatly help PR #914 

**_I affirm:_**

### DIFF
--- a/src/map/time_server.cpp
+++ b/src/map/time_server.cpp
@@ -87,6 +87,28 @@ int32 time_server(time_point tick,CTaskMgr::CTask* PTask)
         }
     }
 
+    //JST Midnight
+    static time_point lastTickedJstMidnight = tick - 1s;
+    if (CVanaTime::getInstance()->getJstHour() == 0 && CVanaTime::getInstance()->getJstMinute() == 0)
+    {
+        if (tick > (lastTickedJstMidnight + 1h))
+        {
+            // roeutils::CycleDailyRecords();
+            lastTickedJstMidnight = tick;
+        }
+    }
+
+    //4-hour RoE Timed blocks
+    static time_point lastTickedRoeBlock = tick - 1s;
+    if (CVanaTime::getInstance()->getJstHour() % 4 == 0 && CVanaTime::getInstance()->getJstMinute() == 0)
+    {
+        if (tick > (lastTickedRoeBlock + 1h))
+        {
+            // roeutils::CycleTimedRecords();
+            lastTickedRoeBlock = tick;
+        }
+    }
+
     if (CVanaTime::getInstance()->getHour() == 0 && CVanaTime::getInstance()->getMinute() == 0)
     {
         if (tick > (CVanaTime::getInstance()->lastVDailyUpdate + 4800ms))
--- a/src/map/vana_time.cpp
+++ b/src/map/vana_time.cpp
@@ -19,6 +19,10 @@
 ===========================================================================
 */
 
+#ifdef WIN32
+#define timegm _mkgmtime
+#endif
+
 #include "../common/showmsg.h"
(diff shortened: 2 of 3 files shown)


src/map/time_server.cpp
```diff
@@ -87,6 +87,28 @@ int32 time_server(time_point tick,CTaskMgr::CTask* PTask)
         }
     }
 
+    //JST Midnight
+    static time_point lastTickedJstMidnight;
+    if (CVanaTime::getInstance()->getJstHour() == 0 && CVanaTime::getInstance()->getJstMinute() == 0)
+    {
+        if (tick > (lastTickedJstMidnight + 1h))
+        {
+            // roeutils::CycleDailyRecords();
```
Kreidos, line 96: These commented lines are left to avoid future merge conflicts if other functions are moved into these areas before RoE Phase 3 is ready.

kaincenteno:
Awesome! finally real vanatime

cocosolos:
Looks good, I just have one question about the JST midnight function. Am I reading it right that it returns last midnight instead of next midnight? I only ask because the Lua midnight function returns next midnight so there might be some confusion.

Kreidos:
> Looks good, I just have one question about the JST midnight function. Am I reading it right that it returns last midnight instead of next midnight? I only ask because the Lua midnight function returns next midnight so there might be some confusion.

It returns the upcoming midnight. However, the easiest way to calculate that is to calculate the previous midnight and add 1 day.

cocosolos:
Oh I see, the comment confused me, carry on!

Kreidos:
Thanks for the catch, I didn't notice the comment!
I'm fixing it along with the Win32 builds. :unamused:

zach2good (approved):
I don't pretend to really understand all the nuances and things people would want in regard to JP Midnight, JST, guilds and records getting swapped out at intervals relating to JP midnight because of the client, not the server's locale etc etc etc.

Hopefully @TeoTwawki will have something interesting to say.

I was getting myself all prepared to question your use of `static` until I saw that `time_server` is just a standalone function getting called from the `taskmanager`. Ew. Good call 😆

Kreidos:
> I was getting myself all prepared to question your use of `static` until I saw that `time_server` is just a standalone function getting called from the `taskmanager`. Ew. Good call laughing

I love statics when the case is right.<3 
The fact that all the other categories still use class variables unnecessarily still irks me to no end, but :shrug:

TeoTwawki:
> Hopefully @TeoTwawki will have something interesting to say

something interesting

*ducks and covers*

zach2good:
_Laughs in windows_

TeoTwawki (approved):

cocosolos:
Guild patterns are hard coded in the client to use JST midnight, and since that's the only thing using midnight it might be worth it to just get rid of that block and call the guild update in the new JST midnight block, I think it only requires 2 changes in guildutils.cpp from `getSysYearDay` to `getJstYearDay`.

TeoTwawki:
_Sees build fail a hair too late_

😱

zach2good:
> Guild patterns are hard coded in the client to use JST midnight, and since that's the only thing using midnight it might be worth it to just get rid of that block and call the guild update in the new JST midnight block, I think it only requires 2 changes in guildutils.cpp from getSysYearDay to getJstYearDay.

You could close two PRs with this... if you get my drift

Kreidos:
> You could close two PRs with this... if you get my drift

if @jarmengaud is fine with me usurping his PR, I'll move the line up and trash the system local midnight block.

PR #1214, opened 2020-09-29
Dangruf Wadi Trimmer is not aggressive but links

MarianArlt:
**_I affirm:_**

Very straight forward. Trimmer crabs in Dangruf Wadi are supposed to be non aggressive according to [here](https://ffxiclopedia.fandom.com/wiki/Trimmer) and [here](https://www.bg-wiki.com/bg/Trimmer).


zach2good (approved):
Yup, looks good!

This moves the flag from `aggro` to `links`
```sql
  `aggro` tinyint(1) unsigned NOT NULL DEFAULT '0',
  `true_detection` tinyint(1) unsigned NOT NULL DEFAULT '0',
  `links` tinyint(1) unsigned NOT NULL DEFAULT '0',
```

PR #1215, opened 2020-09-29
Change Conquest update + Guild Shop resets to JST time

### LINKED ISSUE #22: Guild Points
<a href="https://github.com/dazusu"></img></a> **Issue by [dazusu](https://github.com/dazusu)**
_Thursday Jun 04, 2015 at 12:49 GMT_
_Originally opened as DarkstarProject/darkstar - Issue 1519_

----

1. The daily turn-in item doesn't always reset. (meaning the same item for the same rank is requested sometimes 3-4 times in a row)
2. The "reset" period of when you can hand in items again should be affected by the "time offset" applied to the server for Japanese midnight; it is not (the NPC in game will tell you 3 in-game days), when in fact, due to the time offset on the server, it is more like 10 in-game days.

### LINKED ISSUE #552: Two different settings for server JP midnight time.

### LINKED ISSUE #750: Cooking guild NPC does not accept bowl of puls for guild points

### LINKED ISSUE #875: Daily reset time does not honor "JP Midnight" default setting

Kreidos:
+ Move Guild Reset over to JST midnight instead of local.
+ fix + remove single-purpose vars from global Vanatime class.
+ Move Conquest over to JST instead of local time.
+ Migrate Lua over to new JST function for getMidnight.

Fixes #750
Fixes #875
Fixes #22 
Closes #552 

**_I affirm:_**

### DIFF
--- a/scripts/globals/common.lua
+++ b/scripts/globals/common.lua
@@ -46,29 +46,10 @@ end
 
 -----------------------------------
 --  getMidnight
---  Returns midnight for the current day in epoch format
+--  Returns the next upcoming JST midnight
 -----------------------------------
 
-function getMidnight(day)
-    -- Get time, because this is going to get ugly.  Using ! for UTC won't work where we're going.
-    local hometime = os.date("*t")
-    if day ~= nil then
-        hometime = os.date("*t", day)
-    end
-    -- Set to 24:00 to get end of the day.
-    local midnight = os.time{year=hometime.year, month=hometime.month, day=hometime.day, hour=24}
-    -- And determine the timezone in seconds, because we'll need that to get UTC and LUA doesn't make it easy.
-    local timezone = os.difftime(os.time(), os.time(os.date("!*t")))
-
-    -- Midnight adjusted for timezone, then timezone offset * 3600 to get us where we want to be.
-    local finaltime = midnight + timezone - (TIMEZONE_OFFSET * 3600)
-    -- And make sure that the offset midnight isn't already passed
-    if day ~= nil and finaltime <= day or day == nil and finaltime <= os.time() then
-        finaltime = finaltime + 86400
-    end
-
-    return finaltime
-end
+getMidnight = JstMidnight
 
 -----------------------------------
 --  getVanaMidnight(day)
--- a/scripts/globals/settings.lua
+++ b/scripts/globals/settings.lua
@@ -143,7 +143,6 @@ HOMEPOINT_HEAL = 0 --Set to 1 if you want Home Points to heal you like in single
 RIVERNE_PORTERS = 120 -- Time in seconds that Unstable Displacements in Cape Riverne stay open after trading a scale.
 LANTERNS_STAY_LIT = 1200 -- time in seconds that lanterns in the Den of Rancor stay lit.
 ENABLE_COP_ZONE_CAP = 0 -- enable or disable lvl cap
-TIMEZONE_OFFSET = 9.0 -- Offset from UTC used to determine when "JP Midnight" is for the server.  Default is JST (+9.0).
 ALLOW_MULTIPLE_EXP_RINGS = 0 -- Set to 1 to remove ownership restrictions on the Chariot/Empress/Emperor Band trio.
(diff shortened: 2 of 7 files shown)


Kreidos:
Moved conquest updates over to JST as well, because why not!

scripts/globals/common.lua
```diff
@@ -46,29 +46,10 @@ end
...
-    -- Midnight adjusted for timezone, then timezone offset * 3600 to get us where we want to be.
-    local finaltime = midnight + timezone - (TIMEZONE_OFFSET * 3600)
-    -- And make sure that the offset midnight isn't already passed
-    if day ~= nil and finaltime <= day or day == nil and finaltime <= os.time() then
-        finaltime = finaltime + 86400
-    end
-
-    return finaltime
-end
+getMidnight = JstMidnight
```
TeoTwawki, line 52: I'ma  lil too late, but I would have said just give it the name we had already been using and skip the assignment lol. Thanks for making this happen Kreidos.

ibm2431:
Explanation for those not following Discord:
Settings to change when "midnight" occurs can't make in-game things happen _at_ the midnight in an _arbitrary_ timezone - they can only be used to change _your_ midnight _from_ your local midnight _to_ JST, so you can run the server _without having to change your system's timezone_.

Settings to change when in-game midnight "is" (ie: making things happen at midnight in _your local_ timezone) **_can't_** function as desired. **The client is hardcoded to _always expect certain things_ to occur at midnight in JST. No server-side shenanigans can ever change these hardcoded client events and expectations.**

scripts/globals/common.lua
Kreidos: I considered that. We can still call it either way, just that I was concerned the name wasn't descriptive enough before, and that it didn't follow capitalization standards the way other Luautils functions do. Plus, maybe there'd be momentum behind the existing name. And I questioned changing 13 other files to match. :P I'm indifferent if someone wants to go ahead and change the 13 usages to the new name.

TeoTwawki:
Before anyone brings up any formerly JP midnight things that now aren't: _those_ ones aren't hardcoded into the event in the client like we're talking about, those just used it for a trigger condition. _Those_ can still change. What time is actually JP Midnight cannot, as ibm2431 has stated.

ibm2431:
Basically:
- We can make an NPC accept a trade whenever we want. (trigger condition)
- We can't make guilds change their pattern at any time other than true JST. (hardcoded client expectation)

Kreidos:
> Before anyone brings up any formerly JP midnight things that now aren't: _those_ ones aren't hardcoded into the event in the client like we're talking about, those just used it for a trigger condition. 

I believe all things should correctly be JST-midnight now, but if things aren't please open issue or discuss on discord; should be very easy to *make* it correct. :)

PR #1225, opened 2020-10-01
Effects Rework: Simple effect items

zach2good:
**_I affirm:_**

**Tested with**
Acid Claws (`!additem 16430` on NIN99)
Seems to proc about 10% of the time, message comes up (but doesn't show in BattleMod logs, should it?)


TeoTwawki (approved):
- [x] add-effect type "normal" for status effect items
 - [x] subeffect animation set
 - [x] chance i.e. proce rate is set
 - [x] status effect id set
 - [x] duration of the effect is set

lgtm, it ticked all the needed boxes. I will look into the message issue if that's confirmed to only happen on this branch. I used the same logic we had so I'm not sure how that would have changed.

PR #1226, opened 2020-10-01
Blu Spell: Pinecone Bomb

zach2good:
**_I affirm:_**

Seeing as `feature/blue_mage` is being put to the side, for now, it's time to start cannibalizing all the good bits! The plumbing went in for this ages ago, so this is an easy first piece. 

I know there are servers out there (new and old) that want this...

Works, and the mob goes to sleep, but there isn't a message. Should there be?
Theres a message when they wake up...
image


TeoTwawki:
I think I recall us determining that seeing only the damage was normal.

TeoTwawki (approved):

PR #1234, opened 2020-10-02
You can only Familiar any given pet once

zach2good:
**_I affirm:_**

Was recently reminded about this.

image


Era-Lusiphur:
Works well.

PR #1241, opened 2020-10-03
Advanced Trust Positioning & Trust: Koru-Moru

zach2good:
**_I affirm:_**

Uses _maths_ to determine where trusts should stand.

**DONE**
Different "types" of movement will run to valid positions and stay there until conditions change
1-2 Trusts that use this type of movement
A sane way of attaching these movement styles to trusts (Lua, DB etc.)
Update `feature/trust` with the fix commit that's lurking in here, because we're changing branch permissions and I got impatient.

image


zach2good:
Like I mentioned in Discord, this is literally the most checking I can be doing to see if a new position is valid:

```cpp
        // Validate position
        if (POwner->PAI->PathFind->ValidPosition(potential_position) &&
            POwner->loc.zone->m_navMesh->raycast(PTarget->loc.p, potential_position) &&
            !POwner->loc.zone->m_navMesh->findPath(POwner->loc.p, potential_position).empty())
```

Without updated navmeshes, there will always be problems with pathing 🤷

PR #1243, opened 2020-10-03
Notoriety Container

zach2good:
**_I affirm:_**

```
// If enmity is a one-to-many list held by monsters to track who "they hate",
// notoriety is the many-to-one opposite list held by players (or other targets of enmity) to track who "hates them".
```

A PC should be able to do a simple lookup to see if any mobs have hate towards them. Currently, they have to look at their `mobSpawnList`, iterate through all of their `EnmityContainers` and see if they're on any of those lists. It's been this way forever, so it doesn't appear to be too much of a performance drain, but we also need this sort of lookup for something like: `char:hasEnmity()` or generating lists of things to target with auras etc.

It's actually very simple, it piggybacks _entirely_ off of `EnmityContainer`.

**Design**
To support stability, every call into this container should fail gracefully. I'm starting to think that `clear` doesn't need to exist, since the regular enmity container checks for if a target is in zone, or is alive: the two conditions that this container might clear itself in. 
It would make more sense to just use the underlying data structure, but not all exposed calls are safe. So wrapping it and only exposing safe calls was the way I chose.

**Testing**
- Being aggro'd adds the mob to your list
- Zone with hate
- Die with hate
- AOE a bunch of mobs who hate you
- Kiting and counting mobs with: `ShowStatus("Aggro: %n", PNotorietyContainer->size());`
- Tested with PLD, WAR & DRG enmity changing abilities
- I've also been testing this in my `trust` branch, the checks to see if you have enmity before you can summon trusts, and the checks before trusts start "resting" now work as intended.
- Wait long enough for zone cleanup after logout

**Done**
- Container, piggybacking on `EnmityContainer`
- Basic testing
- Binding: `hasEnmity()`
```lua
local hasEnmity = player:hasEnmity()
```
- Binding `getNotorietyList()`
```lua
    local list = player:getNotorietyList()
    for _, entry in pairs(list) do
        print(entry:getName())
    end
```
- Plug into Auras
- Test with enmity reducing/resetting abilities


zach2good:
I've been hyping this up as the blocker for many different features for about 6 months, so now that it's finally here and it piggybacks off of something that is incredibly battle-tested, I'm happy to merge it straight into `release`. I will of course act as front-line support if it breaks anything and have my finger hovering over the revert button.

TeoTwawki:
`PNotorietyContainer`

not sure about that naming but I got nothing better yet lol

zach2good:
Iterators... that was painful to fix

PR #1249, opened 2020-10-04
Guildshops truncated list of items

jarmengaud:
**_I affirm:_**

Cooking guild had truncated list of items because of incorrect packet flag


zach2good (approved):
Looks simple enough!

PR #1254, opened 2020-10-05
styleguide fixes for windurst walls + switch to npcUtil + fix global vars

kaincenteno:
**_I affirm:_**

Would love a second set of eyes as it was a mass change and nothing was tested.

### DIFF
--- a/scripts/zones/Windurst_Walls/npcs/Ambrosius.lua
+++ b/scripts/zones/Windurst_Walls/npcs/Ambrosius.lua
@@ -6,54 +6,42 @@
 -----------------------------------
 require("scripts/globals/quests")
 require("scripts/globals/settings")
+require("scripts/globals/npc_util")
 -----------------------------------
 
-function onTrigger(player, npc)
-    QuestStatus = player:getQuestStatus(WINDURST, tpz.quest.id.windurst.THE_POSTMAN_ALWAYS_KO_S_TWICE)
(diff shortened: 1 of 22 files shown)


zach2good (changes requested):
Styling looks good, apart from what we spoke about on discord:
```lua
    elseif csid == 102 and npcUtil.completeQuest(player, WINDURST, tpz.quest.id.windurst.ROCK_RACKETEER, {
            gil = 2100,
            var = "rockracketeer_sold"
        })
    then
```
Note that this leads to an indent of 8 adter the completeQuest call. `npcUtil.completeQuest` is so big and complex it can't reasonably be grouped in with "simple logic can stay on one line", so break it out:
```lua
    elseif 
        csid == 102 and 
        npcUtil.completeQuest(player, WINDURST, tpz.quest.id.windurst.ROCK_RACKETEER, {
            gil = 2100,
            var = "rockracketeer_sold"
        })
    then
```

```
[   ] that I've tested my code sin
  ^---
```
```
... nothing was tested ...
```

I know that a lot of the old NPCs in main areas (Shantotto, Halver etc.) are nightmares and their code is utter spaghetti. They are also _very important_ NPCs with a lot of quests relying on them. They **NEED** to be tested and confirmed working after changes.

kaincenteno:
conflicts with other PR.

PR #1256, opened 2020-10-05
Add RoE Disable switch in settings.lua

Kreidos:
This still maintains support for showing the player's Spark amount and giving/removing spark currency (In case any server wants alternate ways to obtain them or custom usage.)

Summary of fail-out conditions when RoE Disabled:
- Client RoE packets won't be replied to (Take record, Remove record, Quest Log.) RoE client menu will be blank.
- RoE Events bail-out immediately.
- No RoE writes to DB

**_I affirm:_**

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -6062,8 +6062,11 @@ void SmallPacket0x10B(map_session_data_t* session, CCharEntity* PChar, CBasicPac
 
 void SmallPacket0x10C(map_session_data_t* session, CCharEntity* PChar, CBasicPacket data)
 {
-    roeutils::AddEminenceRecord(PChar, data.ref<uint32>(0x04));
-    PChar->pushPacket(new CRoeSparkUpdatePacket(PChar));
+    if (roeutils::RoeSystem.RoeEnabled)
+    {
+        roeutils::AddEminenceRecord(PChar, data.ref<uint32>(0x04));
+        PChar->pushPacket(new CRoeSparkUpdatePacket(PChar));
+    }
     return;
 }
 
@@ -6075,8 +6078,11 @@ void SmallPacket0x10C(map_session_data_t* session, CCharEntity* PChar, CBasicPac
 
 void SmallPacket0x10D(map_session_data_t* session, CCharEntity* PChar, CBasicPacket data)
 {
-    roeutils::DelEminenceRecord(PChar, data.ref<uint32>(0x04));
-    PChar->pushPacket(new CRoeSparkUpdatePacket(PChar));
+    if (roeutils::RoeSystem.RoeEnabled)
+    {
+        roeutils::DelEminenceRecord(PChar, data.ref<uint32>(0x04));
+        PChar->pushPacket(new CRoeSparkUpdatePacket(PChar));
+    }
     return;
 }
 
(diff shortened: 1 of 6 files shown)


src/map/roe.cpp
```diff
@@ -33,13 +33,16 @@
...
 std::array<RoeCheckHandler, ROE_NONE> RoeHandlers;
-RoeSystemData roeutils::RoeCache;
+RoeSystemData roeutils::RoeSystem;
 
 namespace roeutils
 {
 void init()
 {
     lua_State* L = luautils::LuaHandle;
+    lua_getglobal(L, "ENABLE_ROE");
```
zach2good, line 43: Smart!

src/map/utils/charutils.cpp
```diff
@@ -4006,6 +4007,9 @@ namespace charutils
 
     void SaveEminenceData(CCharEntity* PChar)
     {
+        if (! roeutils::RoeSystem.RoeEnabled)
```
zach2good: No space after ! please

src/map/packet_system.cpp
```diff
@@ -6137,12 +6143,17 @@ void SmallPacket0x111(map_session_data_t* session, CCharEntity* PChar, CBasicPac
...
-    for(int i = 0; i < 4; i++)
-        PChar->pushPacket(new CRoeQuestLogPacket(PChar, i));
+
+    if (roeutils::RoeSystem.RoeEnabled)
+    {
+        // Current RoE quests
+        PChar->pushPacket(new CRoeUpdatePacket(PChar));
+        // 4-part Eminence Completion bitmap
+        for(int i = 0; i < 4; i++)
+            PChar->pushPacket(new CRoeQuestLogPacket(PChar, i));
```
zach2good: Space after for
Brackets around the statement

zach2good (approved):
Looks like exactly what the doctor ordered! Just some style things before ready to merge

src/map/packet_system.cpp
Kreidos: This line actually pre-dates this commit. I didn't modify it. and both lines really represent a singular command, as the questLog must **always** be as a set of 4.

Kreidos: Are we averse to having it all on one line? It will never change as it's concrete in its required function.

zach2good: The end result I meant was:
```cpp
for (int i = 0; i < 4; i++)
{
    PChar->pushPacket(new CRoeQuestLogPacket(PChar, i));
}
```

I think the only thing I like for single line ifs are returns.

Maybe it wouldn't be such a bad thing to just have these 4 lines without a loop?
```cpp
PChar->pushPacket(new CRoeQuestLogPacket(PChar, 0));
PChar->pushPacket(new CRoeQuestLogPacket(PChar, 1));
PChar->pushPacket(new CRoeQuestLogPacket(PChar, 2));
PChar->pushPacket(new CRoeQuestLogPacket(PChar, 3));
```

Kreidos: ...I will put the brackets... I just hate how bloated single-line braced sections look. :unamused: 

My preference would to avoid bracing would be:
```c++
for (int i = 0; i < 4; i++) PChar->pushPacket(new CRoeQuestLogPacket(PChar, i));
```
(If this looks offensive, then don't throw stones unless you've never use lambdas! :joy: )

But I didn't want to unroll it, since the compiler will be doing that anyway, and I suspected it's possible that SE may expand the record system in the future; there's clear evidence that's why this multi-packet situation already exists.

zach2good:
`clang-format` has been disabled as a check for a while, mysteries continue

PR #1257, opened 2020-10-05
Correcting ZilartStatus checks

nsabott:
1: ZilartStatus needs to be '257' not '255', because it started at 2 and 255 was added.
2: You need to re-get the char var when checking to start cutscene upon selecting the last Pedestal because it was added to, so the check looks at the old stored variable instead of the updated one.

**_I affirm:_**


zach2good:
Hi! Welcome to the project and congrats on your first PR (so says GitHub)!

If you haven't already, please join us in discord:
https://discord.gg/nbYJfA5

I'll have time to review this properly in a few days, but at a quick glance it looks good 👍

ibm2431:
It looks like the pedestals are already using a standard bitmask that starts at 1 with the Pedestal of Fire. Bitmask of 8 bits = 255.

[I don't see ZilartStatus being set to an initial value after the completion of the BCNM](https://github.com/project-topaz/topaz/blob/release/scripts/zones/Chamber_of_Oracles/bcnms/through_the_quicksand_caves.lua#L34-L36).

PR #1260, opened 2020-10-05
Changed pos so it would display cutscene without freezing.

nsabott:
Changed the pos so that it leaves the character outside the instance afterwards, having it at <0, 0, 0, 0, zoneId) was causing issues with the cutscene displaying properly, it would freeze.

**_I affirm:_**

### DIFF
--- a/scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
+++ b/scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
@@ -35,15 +35,15 @@ function onInstanceProgressUpdate(instance, progress)
     if (progress == 4) then
         local chars = instance:getChars()
 
-        for i, v in pairs(chars) do
-            v:startEvent(203)
-        end
-
         DespawnMob(ID.mob[59].RAZFAHD, instance)
         SpawnMob(ID.mob[59].ALEXANDER, instance)
 
     elseif(progress == 5) then
         instance:complete()
+		v:setPos(126.6543, 0, 20.1798, 0, 72)
+		for i, v in pairs(chars) do
+			v:startEvent(203)
+        end
     end
 end
 
@@ -56,7 +56,7 @@ function onInstanceComplete(instance)
             v:setCharVar("AhtUrganStatus", 2)
         end
 
-        v:setPos(0, 0, 0, 0, 72)
+        v:setPos(126.6543, 0, 20.1798, 0, 72)
     end
 end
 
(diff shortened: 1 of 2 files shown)


scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
```diff
@@ -35,15 +35,15 @@ function onInstanceProgressUpdate(instance, progress)
...
-
         DespawnMob(ID.mob[59].RAZFAHD, instance)
         SpawnMob(ID.mob[59].ALEXANDER, instance)
 
     elseif(progress == 5) then
         instance:complete()
+		v:setPos(126.6543, 0, 20.1798, 0, 72)
+		for i, v in pairs(chars) do
+			v:startEvent(203)
+        end
```
cocosolos, line 46: Is this cs supposed to play here? [This video](https://youtu.be/SLr-slWFjxQ?t=954) looks like its supposed to play when Alexander spawns.

zach2good:
If this was the fix to the black screen in this fight I'm going to lose my mind, it's the one thing I didn't try! 😆 

Please check your styling: 4 space indents, no tabs, etc.
https://github.com/project-topaz/topaz/blob/release/CONTRIBUTING.md#general-code-guidlines-all-languages

scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
zach2good: Yeah, this moves the order of the CS to the end of the battle, instead of between phases.

scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
```diff
@@ -35,15 +35,15 @@ function onInstanceProgressUpdate(instance, progress)
...
-        for i, v in pairs(chars) do
-            v:startEvent(203)
-        end
-
         DespawnMob(ID.mob[59].RAZFAHD, instance)
         SpawnMob(ID.mob[59].ALEXANDER, instance)
 
     elseif(progress == 5) then
         instance:complete()
+		v:setPos(126.6543, 0, 20.1798, 0, 72)
```
zach2good, line 43: v isn't accessible until you're inside the loop

scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
```diff
@@ -35,15 +35,15 @@ function onInstanceProgressUpdate(instance, progress)
...
+		for i, v in pairs(chars) do
+			v:startEvent(203)
```
zach2good, line 45: Be careful not to introduce full or half-tabs, only 4x spaces for indentation please

zach2good (changes requested):
You might be barking up the right tree for how to fix this, but as it stands this PR doesn't work :(

PR #1261, opened 2020-10-05
Move bitmask functions from core to LUA

wrenffxi:
**_I affirm:_**

No need to call core for these functions.  The only one that touched anything in core was setBitMask, and 1. it wasn't correctly documented and was misused in several places; and 2. we can just as easily setCharVar after setting the bit via LUA.

I kept each step of this process in its own commit for easier reviewing:

* Create bitmask functions on utils.mask
* Convert existing use cases
* Remove bitmask functions from lua_basentity
* Some minor lua cleanup
* Rewrite Mamaulabion's trade code because it bothered me

Before PRing, I tested without errors: Dynamis flagging; quest "Mama Mia"; and quest "Lure of the Wildcat (Windurst)"


scripts/globals/utils.lua
```diff
@@ -394,3 +394,62 @@ function utils.hasRelic(player, relic, tier)
...
+            len = 32
+        end
+
+        local count = 0
+
+        for i = 0, len - 1 do
+            if bit.band(mask, bit.lshift(1, i)) ~= 0 then
+                count = count + 1
+            end
+        end
```
Kreidos, line 437: Cool beans!

I'm sure this doesn't *have* to be micro-optimized, but simply adding all the bits without the `if` would be a bit neater.
```lua
count = count + bit.band(bit.rshift(mask, i), 1)
```

Kreidos (approved):
I am pro anything that removes lines from lua_baseentity. :joy: Good stuff.

Assuming all the quest script refactors are tested or at least mostly tested, the new functions look good to me.

TeoTwawki (approved):
I approves of deez utils that don't make me go typing out bit.whatevertheF instead of 'isFull' etc.

scripts/globals/utils.lua
wrenffxi: I like it!  Implemented, retested, and looks good.

zach2good (approved):
A much better solution than calling out into Core! Nice work!

PR #1266, opened 2020-10-06
use GetNPC/MobByID for getting instanced entities

wrenffxi:
**_I affirm:_**

We fixed GetMobByID [#642] and GetNPCByID [#668] a while back, so let's use those for getting instanced entities, rather than instance:getEntity + bit.band + objType (which may confuse and deter new contributors)


TeoTwawki (approved):
_-slams approve button repeatedly-_

TeoTwawki (approved):

zach2good (approved):
It _just makes sense_! Thanks!

PR #1268, opened 2020-10-07
Fix issue with claim dropping if owner zones

cocosolos:
**_I affirm:_**


zach2good (approved):
Aha! Good catch

PR #1278, opened 2020-10-08
Trust Chatter

zach2good:
**_I affirm:_**

- `!addalltrusts` was missing a bunch of entries
- Swapped out trusts using their IDs to try and work out their chatter page offsets, because after Klara, they're all jumbled and there's not really a pattern
- Added a little hack to inject chatter on despawn without it being doubled - will need to do this nice and cleanly in the future
- Added a limit check so you can't try and access a chatter page above 120 - which leads to your client not being able to accept messages anymore
- Added chatter for Monbereaux and Sakura, just because


kaincenteno:
Peep! Peep!♪ :chicken:

PR #1287, opened 2020-10-08
Added RoV 2-1 [SPIRITS_AWOKEN]

kaincenteno:
onZoneIn passing params is not implemented, so I commented it out

**_I affirm:_**

### DIFF
--- a/scripts/zones/Lower_Delkfutts_Tower/Zone.lua
+++ b/scripts/zones/Lower_Delkfutts_Tower/Zone.lua
@@ -21,19 +21,48 @@ end
 
 function onZoneIn(player, prevZone)
     local cs = -1
+    local params = {0, 0, 0, 0, 0, 0, 0, 0}
 
-    if player:getXPos() == 0 and player:getYPos() == 0 and player:getZPos() == 0 then
+    if
+        player:getXPos() == 0 and
+        player:getYPos() == 0 and
+        player:getZPos() == 0
+    then
         player:setPos(460.022, -1.77, -103.442, 188)
     end
-    if player:getCurrentMission(ZILART) == tpz.mission.id.zilart.RETURN_TO_DELKFUTTS_TOWER and player:getCharVar("ZilartStatus") <= 1 then
+
+    -- RETURN TO DELKFUTTS TOWER
+    if
+        player:getCurrentMission(ZILART) == tpz.mission.id.zilart.RETURN_TO_DELKFUTTS_TOWER and
+        player:getCharVar("ZilartStatus") <= 1
+    then
         cs = 15
-    elseif ENABLE_COP == 1 and prevZone == tpz.zone.QUFIM_ISLAND and player:getCurrentMission(COP) < tpz.mission.id.cop.THE_RITES_OF_LIFE then
+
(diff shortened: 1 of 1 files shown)


scripts/zones/Lower_Delkfutts_Tower/Zone.lua
```diff
@@ -21,19 +21,48 @@ end
...
+    -- BORN OF HER NIGHTMARES
+    elseif
+        player:getCurrentMission(ACP) == tpz.mission.id.acp.BORN_OF_HER_NIGHTMARES and
+        prevZone == tpz.zone.QUFIM_ISLAND
+    then
         cs = 34
     end
 
-    return cs
+    return cs --,{params} need to pass a param for CS 51 to show extra dialog
```
Kreidos: These params aren't passed as part of the cs start.
They're passed as a response to an EventUpdate the client sends with option 1. That response and option needs to be coded in onEventUpdate below.

kaincenteno: :penguin:  ty

scripts/zones/Lower_Delkfutts_Tower/Zone.lua
```diff
@@ -22,14 +22,38 @@ end
 function onZoneIn(player, prevZone)
     local cs = -1
 
-    if player:getXPos() == 0 and player:getYPos() == 0 and player:getZPos() == 0 then
```
TeoTwawki, line 25: we actually didn't exceed the width there, it was ok.

scripts/zones/Lower_Delkfutts_Tower/Zone.lua
```diff
@@ -54,19 +78,36 @@ function onRegionLeave(player, region)
 end
 
 function onEventUpdate(player, csid, option)
+    if
+        csid == 51 and
```
TeoTwawki, line 82: short enough to be one line here

scripts/zones/Lower_Delkfutts_Tower/Zone.lua
```diff
@@ -54,19 +78,36 @@ function onRegionLeave(player, region)
...
         if player:getCharVar("option") == 1 then
             player:setPos(-28, -48, 80, 111, 157)
         else
             player:setPos(-51, -48, -40, 246, 157)
         end
         player:setCharVar("option", 0)
-    elseif csid == 4 and (option == 0 or option >= 3) then
+    elseif
+        csid == 4 and
+        (option == 0 or
```
TeoTwawki, line 108: so uh... I guess ya just did that to all the if/thens? [we normally only break the wide ones](https://github.com/project-topaz/topaz/blob/release/CONTRIBUTING.md#general-code-guidlines-all-languages)

kaincenteno:
@TeoTwawki  dunno if im doing something wrong but after the styleguide talk on discord i understood that if there's several and or then it can be done one per line. Under "Lua"

@staff would love some clarification :bow: i would hate to be doing something wrong

kaincenteno:
Closing as it needs more research for retail accuracy, but it should help with CSID and some of the params that are being passed.

PR #1289, opened 2020-10-09
Added RoV 2-2 [CRASHING WAVES]

kaincenteno:
added 4 extra cutscenes.
only thing missing is the mystic retriever for tenzen if inventory is full. Please open an issue to be added after merge. (Currently none of the mystic retrieveres are visible in map)

**_I affirm:_**


kaincenteno:
Closing as it needs more research for retail accuracy, but it should help with CSID and some of the params that are being passed.

PR #1290, opened 2020-10-09
Added RoV 2-3 [CALL TO SERVE]

kaincenteno:
extra params have been commented out as unsure when they are added. Needs to be added as an issue when this is merged, along with the Mystic Retriever if their inventory is full.

**_I affirm:_**

### DIFF
--- a/scripts/zones/Port_Jeuno/Zone.lua
+++ b/scripts/zones/Port_Jeuno/Zone.lua
@@ -9,6 +9,8 @@ require("scripts/globals/settings")
 require("scripts/globals/chocobo")
 require("scripts/globals/quests")
 require("scripts/globals/zone")
+require("scripts/globals/rhapsodies")
+require("scripts/globals/npc_util")
 -----------------------------------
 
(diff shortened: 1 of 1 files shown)


kaincenteno:
Here are some notes for whomever takes ownership:

These are some notes for RoV 2-4 and 2-5.

```
--------------------------------
-- marble bridge in port jeuno

cs = 10221 --

param1 = "prishe acts like a lady"

param2 = "YOU already know that phoenix is on tenzen's sword" ( param2 = 1 will give u some info, param2 = 2 will give more info)

para3 = "ulmia shows up"
--inscapable binds
-- CAN BE SKIPPED IF ALREADY OWN A BOAT PERMIT TO GO TO THE EAST
-- csid == 10064
-- param6 = 1 "it will show the message about "TENZEN"
-- it will return option 1 (but have to double check it
-- also completes THE ROAD TO AHT UGHAN
-- it completes the quest and goes to the next one DESSERT WINDS
```

PR #1293, opened 2020-10-09
Resolve BCNM entry exploit

ibm2431:
Exact explanation at a later date.

Most servers are already aware of this, and have had a chance to apply this fix. If you're a public server owner who isn't, message zach2good on Discord.

**_I affirm:_**

### DIFF
--- a/scripts/globals/bcnm.lua
+++ b/scripts/globals/bcnm.lua
@@ -965,11 +965,26 @@ function EventUpdateBCNM(player, csid, option, extras)
                 name, clearTime, partySize = battlefield:getRecord()
                 initiatorId, initiatorName = battlefield:getInitiator()
             end
+
             -- register party members
             if initiatorId == player:getID() then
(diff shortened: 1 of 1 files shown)


ibm2431:
When a player interacts with a Burning Circle, they trigger (or trade to) it. They then select various options to identify which battlefield they want to undertake. This initiates a series of responses between the client and the server in the form of eventUpdates.

When the server "finds" a valid, empty battlefield, it then gives the player - and their party members - the battlefield status with the ID of the battlefield they're about to enter. This status is what allows party members to enter the same battlefield as you.

Unfortunately, the orb used to initiate this process wasn't marked as "used" at this point. It was only marked as used when the client sent the final eventFinish back to the server, saying that the player was done with any animations of entering the battlefield. By preventing this final eventFinish from reaching the server, a player could give their party members the battlefield status without "consuming" the entry item. This would allow their party to enter the BC an infinite number of times on the same orb.

PR #1295, opened 2020-10-09
Lindwurm drops were not accurate for skin

jarmengaud:
Source: ffxidb

**_I affirm:_**

Fixing Lindwurm drops to roughly match ffxidb


zach2good (approved):
http://www.ffxidb.com/zones/205/lindwurm

image

PR #1303, opened 2020-10-10
Thunder Hawk and npcUtil key item option

### LINKED ISSUE #1292: Selbina Quest "The Rescue" If you complete a quest, you won't get a reward (Map of the Ranguemont Pass).
**_I have:_**

The parameter ("ki") in "npcUtil.completeQuest" seems to be wrong. (L40)

https://github.com/project-topaz/topaz/blob/5213062798b92405fb84ea90c3399bce75c4bf5e/scripts/zones/Selbina/npcs/Thunder_Hawk.lua#L37-L43

(L40) Use "keyItem" instead of "ki".
```
    elseif csid == 81 and npcUtil.completeQuest(player, OTHER_AREAS_LOG, tpz.quest.id.otherAreas.THE_RESCUE, {fame_area = SELBINA, keyItem = tpz.ki.MAP_OF_THE_RANGUEMONT_PASS, title = tpz.title.HONORARY_CITIZEN_OF_SELBINA, gil = 3000}) then

```

wrenffxi:
**_I affirm:_**

Fixes #1292 

Rather than fixing Thunder Hawk's npcUtil option from **ki** to **keyItem** (See #1292), I decided to go the other way with the fix, and rename the option to **ki** for consistency.  Updated all references in codebase.

Also switched some references to the short-hand **tpz.ki**

### DIFF
--- a/scripts/globals/npc_util.lua
+++ b/scripts/globals/npc_util.lua
@@ -325,8 +325,8 @@ end
 
     Example of usage with params (all params are optional):
         npcUtil.completeQuest(player, SANDORIA, ROSEL_THE_ARMORER, {
-            item = { {640, 2}, 641 },    -- see npcUtil.giveItem for formats
-            keyItem = tpz.ki.ZERUHN_REPORT,    -- see npcUtil.giveKeyItem for formats
+            item = { {640, 2}, 641 },   -- see npcUtil.giveItem for formats
+            ki = tpz.ki.ZERUHN_REPORT,  -- see npcUtil.giveKeyItem for formats
             fame = 120,                 -- fame defaults to 30 if not set
             bayld = 500,
             gil = 200,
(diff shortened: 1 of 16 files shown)


scripts/globals/npc_util.lua
```diff
@@ -349,8 +349,8 @@ function npcUtil.completeQuest(player, area, quest, params)
     end
 
     -- key item(s), fame, gil, bayld, xp, and title
-    if params["keyItem"] ~= nil then
-        npcUtil.giveKeyItem(player, params["keyItem"])
+    if params["ki"] ~= nil then
+        npcUtil.giveKeyItem(player, params["ki"])
```
zach2good, line 353: For anyone else passing-by, this is the important change here!

Looks good, thanks! 👍

zach2good (approved):

scripts/globals/npc_util.lua
TeoTwawki: I'd really like to see the terms ki and keyItem be interchangeable everywhere. we have a high likelihood of future users using the "wrong" one no matter which we keep, so I'd rather we plan for it than simply swap out. I had thought we already did this for the actual global table, just not for the params we're using in the utils functions 

If the "right" one is nil and the other not, lets just populate/duplicate it over before we do the rest of execution.

zach2good: Yeah, agree with this

zach2good: @TeoTwawki 
Just for this case:
https://github.com/project-topaz/topaz/commit/f828092c6ff49a030f331626bf3ac82e915603e8

Issue #1182, opened 2020-09-23
[discussion] Craft Recipe Information Sources


ibm2431:
Long story short (for now) - because some people are trashy I'm doing some work on recipes.

Major part of this is building a database of trustworthy crafting information. Absent a DAT scraper, I've been using information scraped from FFXIAH and BGWiki, procedurally melding it together via tools, and cross-checking the sources against each other.

I've generated the following lists of instances in which they don't agree:
[disagreements.zip](https://github.com/project-topaz/topaz/files/5275639/disagreements.zip)
(Apologies for the zip; the lists are HTML-ified for handy hyperlinks to the sources, and GitHub doesn't like straight attachment of HTML files.)

BGWiki and FFXIAH can disagree as to craft skill levels, result quantities, and even ingredient quantities.

**Crafting level or skill disagreements:**
- Main craft level: [FFXIAH Arhat's Tekko](https://www.ffxiah.com/recipes/1038) - [BGWiki Arhat's Tekko](https://www.bg-wiki.com/bg/Arhat's%20Tekko)
- Sub craft level: [FFXIAH Minnow](https://www.ffxiah.com/recipes/88) - [BGWiki Minnow](https://www.bg-wiki.com/bg/Minnow)
- Main craft skill: [FFXIAH Oak Bed](https://www.bg-wiki.com/bg/Oak%20Bed) - [BGWiki Oak Bed](https://www.bg-wiki.com/bg/Oak%20Bed)
- Both main skill and sub level: [FFXIAH Water Tank](https://www.ffxiah.com/recipes/16) - [BGWiki Water Tank](https://www.bg-wiki.com/bg/Water%20Tank)

"FFXIAH used to scrape the client, we can trust it"

**Result quantity / HQ  disagreements:**
- "Mass Production" Synths: [FFXI Lauan Lumber](https://www.ffxiah.com/recipes/3310) - [BGWiki Lauan Lumber](https://www.bg-wiki.com/bg/Lauan%20Lumber)
- "HQ as NQ" Quantity: [FFXI Rice Dumpling](https://www.ffxiah.com/recipes/1458) - [BGWiki Rice Dumpling](https://www.bg-wiki.com/bg/Rice%20Dumpling)
- HQ Quantity (Desynth): [FFXI Tekko Desynth](https://www.ffxiah.com/recipes/1182) - [BGWiki Tekko Desynth](https://www.bg-wiki.com/bg/Tekko)

"Maybe just the results were changed by retail?"

[...]

PR #1306, opened 2020-10-10
Completely replace all recipe data

ibm2431:
### Premise

Our current synthesis recipe database has some recipes with skill levels which aren't based on _any_ source I could find, regardless of how lax the source standards are. Some of these recipes are very high skill, but with easily sourced materials, which might allow trashy people to exploit said recipes to level their crafts when they probably couldn't on retail.

I'm not going to say what these recipes are.

Because frankly it doesn't matter!

Manual SQL verification looking for "suspicious" recipe is for scrubs! Why vet an existing table when you can **drop the whole thing** and **make a new one based on verifiable sources**? This resolves the "exploitable recipes" angle by simply making the entirety of all recipes match a source! Automation to the rescue!

### Goal

[Using a combination of FFXIAH and BGWiki as sources](https://github.com/project-topaz/topaz/issues/1182), generate a new recipe table which leverages the sources against each other whenever possible to find potential conflicts. Resolve these conflicts with a sensible ruleset and _as few human judgment calls as possible_. No "this feels right" skill guesses. No "I remember this recipe's ingredients from crafting it years ago". If a recipe has incomplete information that can't be resolved using one of the sources, exclude the recipe from being active until it can be verified further.

If a recipe is set to be active _without_ additional research (from retail), it'll be done during "these GitHub times" so there's a paper trail for the decision - no mystery recipes lingering around from the SVN days.

Below I've written a short story on the whole process - primarily for the benefit of The Ones Who Come After. I want there to be a record for this process.

But those who just want the short version:

### Key Notes
- **Future readers: All data taken from FFXIAH and BGWiki were as of September 20th, 2020.**
- Entire recipe table was dropped in favor of a new one
- Several tools and lua scripts were involved
- Craft skill columns were reordered to match the true skill order
- Recipes now have a new ID structure which follows a semantic order of Main Craft Skill - Rank - Increment, so that when recipes are sorted by ID there's some element of order to them, and IDs can help convey information. Ex: All recipes with IDs of 1XXXX are Woodworking, all with IDs of 7XXXX are Cooking. Formula is [Main Craft ID ([true order](https://github.com/Windower/Resources/blob/master/resources_data/skills.lua#L42-L49), 0-indexed) * 10000] + ["Rank ID" (Amateur - Authority, 0-12) * 500] + [Increment per recipe, sorted by craft level, reset to 0 each rank].
- 👀 **I've written a [HTML/JS tool to review our recipe SQL files](https://github.com/ibm2431/recipe-sql-viewer) - for the love of Altana, don't try to review the raw SQL**
- 👀 In the following zip there are many lists of decisions - **including recipe exclusions** - I had to make as a human during the cross-verification process(es). **These include recipes which require additional research, either from other sources, or _ideally_ from retail itself.**
- 👀 **In the zip is a master-recipes.xml which this table was generated from, which not only includes item names, but also FFXIAH recipe IDs _and_ "old database" IDs when applicable**
- GitHub doesn't allow several of these file types to be attached directly, so I've had to zip a folder and attach it instead. **Check the zip for all source files and tools mentioned.**

[new-recipe-database.zip](https://github.com/project-topaz/topaz/files/5358798/new.recipe.database.zip)

The history behind how this one SQL table was generated is as follows:

### Phase 1 - Generate FFXIAH XML

FFXIAH's master XML file was generated by batch downloading all their http://ffxiah.com/recipes/#### pages (ex: [Recipe 1074](https://www.ffxiah.com/recipes/1074)) to a local folder. Then there was a series of regex over the entire folder which replaced FFXIAH's uniform data wrapped in HTML tags to instead be wrapped in XML tags. After that it just took a small Lua "file stitcher" script to open all the files and append their contents into a single file.

**FFXIAH Recipes XML**: zip - sources/ffxiah-recipes.xml

The initial result didn't have the same format as the XML file above. It needed to be converted to match the format I ended up with for BGWiki later. This is touched on more in Phase 3. The original source at this point was:

**Initial FFXIAH Recipes XML**: zip - sources/ffxiah-recipes-phase-one-only.xml

### Phase 2 - Generate BGWiki XML

BGWiki doesn't have a uniform recipe page format like FFXIAH does. Instead it lists recipes in two locations: the individual item page for the synth result (ex: [Tarutaru Sash](https://www.bg-wiki.com/bg/Tarutaru_Sash)) or on the main craft page for an individual craft (ex: [Clothcraft - Veteran](https://www.bg-wiki.com/bg/Clothcraft#Veteran_.2891-100.29)). Since BGWiki is kind enough to list all their known recipes on just the craft pages, I obviously took those rather than trying to scrape all item pages in hopes a synth recipe will be listed on it.

From there I wrote a small javascript tool to traverse the HTML DOM and spit out the recipes in XML format. (Technically the BGWiki pages _are_ generated by template, with source available, but I presumed it'd be easier to traverse a DOM.) So for each craft's master page, download the HTML, use just a couple regex to wrap some important information in dummy spans, include the script in the HTML page's header, and then copy the output of the console log.

**Tool:** zip - tools/bgwiki-javascript-xmler

Since these HTML files were just table rows and didn't have "recipe IDs" like FFXIAH does, I generated IDs myself. These IDs were (craft-id-by-true-order * 1000) + auto-increment-per-craft. The IDs in the XML won't link back anywhere on BGWiki, but they can be used to identify recipes _in_ a XML when doing comparisons. The resulting XML files for each craft were then put together into a single file.

**BGWiki Recipes XML**: zip - sources/bgwiki-recipes.xml

I did not have item IDs at the onset, because BG editors obviously just put the item name in the recipe, and BGWiki links items by their short name. These needed to be added by Lua script later - with many corrections later when item ID lookup failed because an editor referred to an item by name other than its strict item name. (Ex: Editor wrote "Angel Flute +1" instead of proper name of "Angel's Flute +1"). This add-item-IDs process happened during the "meld stage" below.

### Phase 3 - Meld FFXIAH and BGWiki into Melded XML

Now that we have XML files for both FFXIAH and BGWiki, we need to get their recipes into the same file. This is where a _lot_ of Lua gets involved. Leveraging the xml2lua library to transform the XML files into Lua tables, I then began writing several Lua scripts to perform various tasks, step-by-step, to meld FFXIAH's information into BGWiki's information (BGWiki was taken as the "base"). This was a _bit_ of a process, to understate.

Basically, a recipe is the same as another recipe if all of its ingredients are the same. So for all recipes in our base file (BGWiki), compare against all recipes in FFXIAH's source, see if all ingredients are the same, and link the two IDs if so. Once we've cross-linked recipes, any remaining recipes in FFXIAH's file must be new ones that BGWiki doesn't know about, so include them as new entries at the end of the Melded XML.

The idea is simple, but there were a few hitches along the way, and during this phase I found many conflicts between BGWiki and FFXIAH. You can see #1182 , which I opened when this phase brought up these issues, for a discussion of some of these issues.

I'll spare the reader every facet of the scripts, but the summary is:
1. For FFXIAH, write skills to be the same structure as BGWiki's requirements table, decide which skill is main and which are subs, and mark craft rank while we're at it. Rewrite new FFXIAH XML file.
2. For BGWiki, assign item IDs to ingredients and results currently referenced only by name. I leveraged Windower's item resources library for this. While stepping through BGWiki's recipes, I linked any required key item IDs as well. Rewrite new BGWiki XML file.
3. Match recipes IDs. Concept described above. 👀  **While doing this we can begin checking for conflicts and printing out a list of things to investigate - see Phase 4 for more information**
4. Add "FFXIAH only" recipes which BGWiki doesn't have - jot them to a list for further investigation (see Phase 4)
5. Mark desynths. How? A desynth is when you apply a crystal to just one piece of equipment or beast-made item. It's that simple. Don't believe me? See Phase 4.
6. With desynths marked, spit out a list of "FFXIAH only" recipes which aren't desynths for manual validation later.
7. Adjust BGWiki craft levels to FFXIAH's when FFXIAH is "sensible" (see Phase 4)
8. For my sanity, remove "FFXIAH only" recipes which _aren't_ desynths, to be added back manually later - at this point I had fully recognized the conflicts between the two resulting in "clone" recipes which should instead be cross-linked (see - you guessed it - Phase 4)
9. Correct item IDs when there are [multiple items by the same name](https://github.com/project-topaz/topaz/issues/1182#issuecomment-699697555)
10. Try cross-linking FFXIAH recipes to BGWiki recipes again
11. Repeat 4~8 again now that fish item IDs are correct

The _truly invested_ in this gripping tale of triumph over data can find the series of scripts here:

**Lua Scripts** : zip - tools/meld-phase/

The main Lua script file contains "all the scripts" - after executing one task in ZeroBrane (IDE listed in the [Lua manual](https://www.lua.org/start.html)), I'd comment out the task and start writing the next one in the same file.

The result after all of the above was this:

**Silver Master Recipes XML** zip : - sources/master-recipes-silver.xml

(I forget at which point I abandoned bronze, probably around Phase 3 - Step 6, but I forget exactly so I won't include.)

### Phase 4. Conflict Resolution

At this point we've generated the following lists:
- Lists of "FFXIAH only" recipes, which may not really be "FFXIAH only" but ingredient conflicts
- Lists of craft skill mismatches between BGWiki and FFXIAH
- Lists of result quantity mismatches between BGWiki and FFXIAH
- Lists of ingredient mismatches between BGWiki and FFXIAH

A program can't determine the best course of action for such conflicts. A human needs to investigate. That was me.

First, as promised above, see:
**Desynths:** zip - conflicts/desynths.html
For the list of desynths marked above during Phase 3.

👀 **PEOPLE WHO WANT TO REVIEW MY HUMAN DECISIONS, EYES BELOW:** 👀 

The original HTML lists below were output by the Lua scripts through the numerous times they stepped through the XML files and cross-compared things. Once I got to this manual conflict resolution step, I would go through the lists, examine each link, make a call, and jot down what I did into the HTML.

For craft skill mismatches, [FFXIAH used _some_ sort of automated process to come to its stated levels](https://github.com/project-topaz/topaz/issues/1182#issuecomment-700867609). I put more confidence in it most of the time... except for when it clearly makes no sense. During Phase 3 I adjusted the craft levels from BGWiki to match FFXIAH when FFXIAH's was "sensible" (within a few levels) and raised no other red flags. I then made judgment calls for other recipes - **including in some cases to exclude the entire recipe**.

These are over two "rounds", due to Phase 3, Steps 9-11, and me later deciding to _not_ let BGWiki override "sensible automated levels" from FFXIAH [in the context of crafting kits](https://github.com/project-topaz/topaz/issues/1182#issuecomment-700890698).

**Craft Skill Conflicts (Round 1):** zip - conflicts/craft-skill-mismatches-round-1.html
**Craft Skill Conflicts (Round 2):** zip - conflicts/craft-skill-mismatches-round-2.html

For result quantity mismatches, these were the result of FFXIAH using data pulled from what its seen players end up with. As this data can be incomplete by nature (it only lists results it's seen, if no one running Guildwork ever crafts a certain HQ tier, it won't have it). BGWiki was picked for all such mismatches.

**Yield Conflicts:** zip - conflicts/yield-mismatches.html

I started resolving ingredient quantity mismatches at this point. A later list was generated during Phase 6 which was more exhaustive when comparing against the current/old database.

**Ingredient Conflicts (Round 1):** zip - conflicts/potential-mismatches-round-1.html

Then it was time to look into these supposed "FFXIAH only" recipes which _weren't_ desynths. These were pretty much just conflicts between FFXIAH and BGWiki - for example, one source saying the recipe is one crystal with the other source claiming another.

**"FFXIAH Only" (FFXIAH/BGWiki Conflicts):** zip - conflicts/ffxiah-only-no-desynths.html

### Phase 5. From SQL to XML

I wanted to be able to link this new "master" database to recipes from our current/old table. Not only because I wanted there to be a way to translate from "old" recipe IDs to "new" recipe IDs, but because comparing my in-progress "master" database against Topaz's current database might reveal potential conflicts.

I also know for a fact that absolutely no one will review a SQL PR with 4000+ changed recipes by looking at the file diff.

**So I decided to resolve both problems with an HTML/JS tool that can read our recipe SQLs:**

https://github.com/ibm2431/recipe-sql-viewer

For the context of Phase 5, it was only capable of reading our SQL tables, and had a way of printing out (to console) the data as XML structured in the same way as the previous files. Result:

**Old Recipes XML:** zip - sources/old-recipes.xml

### Phase 6. Lua Again - From Old to New

Similar to how I used Lua scripts to meld BGWiki and FFXIAH while cross-comparing their recipes, I did the same for crosslinking our old IDs to the new master file.

**Lua Scripts** : zip - tools/gold-phase/

1. Match recipes in the same manner we did before - comparing ingredients. Jot down potential mismatches for later verification.
2. After crosslinking "old database" recipe IDs into our master file, see what from the old table _wasn't_ matched and jot down what might need to be added

### Phase 7. Going Gold

With the ability to compare our old database to the new master XML, it was time for more manual verification in cases of mismatches.

I generated a second HTML list of potential ingredient mismatches at this point. **Like before, this was later subject to human review on my part.**

**Ingredient Conflicts (Round 2):** zip - conflicts/potential-mismatches-round-2.html

I then checked the BGWiki pages for source items for all the "desynth mismatches" which were flagged before.

**BGWiki Checked Desynths:** zip - conflicts/bgwiki-updated-desynths.html

And like the "FFXIAH only" recipes, investigated "old database only" recipes as well (including "old only" desynths):

**Old Database Only Recipes:** zip - conflicts/old-database-only.html

### Phase 8. To Print!

And then it was time to start spitting out SQL. I did this with the Lua scripts.

👀 **BUT FIRST I GAVE RECIPES NEW IDS** 👀 

Because why waste a huge SQL PR without reordering recipe IDs in some planned manner? It's all about recipes being laid out nicely when sorted by ID!

New IDs were generated as thus:
[Main Craft ID ([true order](https://github.com/Windower/Resources/blob/master/resources_data/skills.lua#L42-L49), 0-indexed) * 10000] + ["Rank ID" (Amateur - Authority, 0-12) * 500] + [Increment per recipe, sorted by craft level, reset to 0 each rank]

This leaves 500 ID slots per-craft-per-rank. Which is way overkill. But originally I had the rank multiplier be 100, leaving 100 slots per-craft-per-rank. While this _was_ sufficient for most craft/ranks, there was _one_ which needed more space (Smithing Veteran has 116 recipes). So alas, they must be spread out more, even if it does ruin the nice 0-12 "midfix" which IDs could have had (XYYZZ). But this means that each rank inside each craft has _plenty_ of space in case people need/want to move stuff around should more recipes be added to that rank.

Ranks which were unknown due to unknown main craft levels were put in the X99YY slot, so they'd be listed last when all recipes are sorted by ID.

👀  **I ALSO CHANGED THE ORDER THAT CRAFT SKILLS ARE LISTED IN** 👀 

Because why waste a 4000+ line PR without bringing our order of SQL columns to align [the order of craft skills](https://github.com/project-topaz/topaz/blob/release/src/map/entities/battleentity.h#L132-L139)?

👀 **AND I HAD SCRIPTS COMMENT OUT ANY RECIPE WITHOUT CONCRETE CRAFT LEVELS** 👀 

As that was the whole point of this project.

**Final Results**

**Final XML:** zip - sources/master-recipes-gold.xml
**Final SQL:** zip - sources/master-recipes-gold.sql

### Phase 9. To Review

"How do you expect people to review such a radical change to the database when you're changing IDs and column orders?"

With [the new recipe viewer tool](https://github.com/ibm2431/recipe-sql-viewer)! That's why I made it! No more having to look up item IDs, check column orders, or count commas to ensure stuff lines up! Just download the synth_recipes.sql locally and run it through the viewer, which will output it into nice HTML complete with links to BGWiki item pages!

There's a checkbox for "New Skill Order" - check that box _before_ loading this (or later) SQL files with the new column order. If you want to view the current/old SQL table, you can use the tool for that too - just leave the "new skill order" box unchecked.

It even includes handy features for showing excluded (commented out) recipes!

### Conclusion

- There are a lot of recipes which will require further discussion and/or retail research before they can be made active in the database while claiming we do our best to ensure accuracy. **I'm not interested in getting stuck in the mud on those.** Forest for the trees, my friends. We can always branch if excluded recipes are a concern, but I'm _done_ with recipes for a while.
- I recognize there may be servers who have custom craft recipes. **If your server has completely made up, not-at-all-on-retail recipes which might get trampled by this new table, throw your current synth SQL at me** and I can move them to new, safe IDs. I _currently_ have a lot of experience with these tools and the process, and can handle it quickly. Offer expires when I begin forgetting said experience!

**_I affirm:_**


ibm2431:
Emphasis: Using the SQL viewer, you can help review recipes with _zero_ knowledge of code! 😉

Xaver-DaRed:
My goodness, good job
Just one thing, i dont think its wise to merge it yet until someone also re-aligns crafts used in the craft container, since you shifted them

ibm2431:
Order of columns in DB doesn't matter to synthutils - [it fetches values by column name (which is already queried in the correct order, the _previous_ table didn't match)](https://github.com/project-topaz/topaz/blob/a07976545ec428808940251c989d92802b0bdb5e/src/map/utils/synthutils.cpp#L69).

zach2good:
I've been listening to this process and being rubber ducky, as it evolved and it all seems solid. Since this is such a brutal change I'll be doing many many spot checks (thanks for the tools!) before this can be merged.

That also gives time to servers watching open PRs to start collecting their custom recipes for migration. 

Again, nice work!

TeoTwawki (approved):
My only complaint is one you don't have to listen to (non retail shenanigans on my part) so stamping approve.

Wish you had not gone for the ID gaps, but I'll live.

zach2good:
Here we go!

Dirk-Dieters:
For continuing this work, what is the standard of evidence for restoring disabled recipes? It seems that for many of the disabled desynths, the only disagreement between BGWiki and FFXIAH is that FFXIAH has the level dummied to 255. 
image
For example.

In these scenarios, would FFXIclopedia agreeing on the level with BGWiki be sufficient to justify the pull?

ibm2431:
For levels I'd be okay with it - maybe include link to BGWiki (and any other sources) in the commit message so the recipe has a citation. (Note: Some exclusions were due to ingredient or yield mismatches.)

The main problem with the old database was that there were way too many mystery recipes with levels that couldn't be found on any source. One such example is the Royal Archer Sword, purchased for 1000 CP, which for unknown reasons had a desynth craft level of 90+.

BGWiki didn't list desynths on their master crafting pages - and I wasn't going to scrape every BGWiki item page for a chance at a desynth - so if there are hidden desynths only on an item page, they'd need to be added manually (with citation this time!).

Finally, for invested readers: general reminder that the [Recipe Viewer Tool](https://github.com/ibm2431/recipe-sql-viewer) has an option to show excluded recipes, and all hyperlinks are to the item's BGWiki page~!

Dirk-Dieters:
In the interest of reviewer sanity, I'll be breaking this into several pulls. The first of which will be strictly items described above, Desynths where the only mismatch is the level being listed as 255 on FFXIAH. After that, I'll start expanding to other types of mismatches.

PR #1320, opened 2020-10-11
Curilla variable renaming

MarianArlt:
**_I affirm:_**

Follow-Up to #1267
Quick variable rename for verbosity.
Also two conditional reductions.


zach2good:
Thanks for this! this was the sort of thing I was trying to ask for but struggling to find the words 👍

PR #1323, opened 2020-10-11
Geomancer Part 1

zach2good:
**_I affirm:_**

After some talk on Discord recently, and me sneaking in my NotorietyContainer, these are the last core changes required for someone to pick up GEO.

This is pretty much the early work me and MrSent did to get GEO up and running, so he is `--author`'d

The plan is to get this into `release`, then someone can start a feature branch from release and start submitting **CHUNKS** of GEO at a time to that feature branch.

**MEGA NOTE:**
The `ModelID` of Luopans should **NOT** be set in Lua, these are only used right now because elemental order isn't consistent yet. This work is on the way. The final signature of `spawnLuopan` should be:
```lua
tpz.geo.spawnLuopan = function(player, target, spell, tick_effect, tick_power, target_type)
```
 
**DONE:**
(Already in release) Geo-Regen effect
Geo-Poison Effect
Geo Global + Spawn Luopan Helper
PetID and Type additions
Luopan Pet Script
Geo-Poison, Geo-Regen, Indi-Poison, Indi-Regen spells
System Messages
Make sure all spell names and file names that have underscores and hyphens are correct

**NOT THIS PR**
Offensive Luopan placement
Bolster 1-Hour


src/map/utils/petutils.cpp
```diff
@@ -1718,6 +1728,19 @@ namespace petutils
                 PPet->addModifier(Mod::MDEF, PChar->PMeritPoints->GetMeritValue(MERIT_FINE_TUNING, PChar));
             }
         }
+        else if (PPet->getPetType() == PETTYPE_LUOPAN && PMaster->objtype == TYPE_PC)
+        {
+            PPet->SetMLevel(PMaster->GetMLevel());
+            PPet->addModifier(Mod::DMG, -50);
+            PPet->setEntityFlags(131);
+            PPet->health.maxhp = (uint32)floor((250 * PPet->GetMLevel()) / 15);
+            PPet->addModifier(Mod::REGEN_DOWN, PPet->GetMLevel() / 4);
```
ibm2431: Regen down is being set in the Lua spawn global (and should probably stay there for ease). I don't remember offhand if [Ecliptic Attrition](https://www.bg-wiki.com/bg/Ecliptic_Attrition) procs on an existing luopan, or if it needs to be used before casting.

src/map/utils/petutils.cpp
```diff
@@ -1718,6 +1728,19 @@ namespace petutils
                 PPet->addModifier(Mod::MDEF, PChar->PMeritPoints->GetMeritValue(MERIT_FINE_TUNING, PChar));
             }
         }
+        else if (PPet->getPetType() == PETTYPE_LUOPAN && PMaster->objtype == TYPE_PC)
+        {
+            PPet->SetMLevel(PMaster->GetMLevel());
+            PPet->addModifier(Mod::DMG, -50);
```
ibm2431: Damage mod should probably be added in the Lua global too, if only so people who want to tinker with bubbles can stay in Lua Land.

scripts/globals/geo.lua
```diff
@@ -0,0 +1,26 @@
...
+-----------------------------------
+
+tpz = tpz or {}
+tpz.geo = tpz.geo or {}
+
+tpz.geo.spawnLuopan = function(player, target, aura_effect, tick_effect, tick_power, target_type, spell_cost)
+
+    tpz.pet.spawnPet(player, tpz.pet.id.LUOPAN)
+    local luopan = player:getPet()
+    local modelID = aura_effect
```
ibm2431: I'm not going to fight the passed in model argument _for now_, but models _shouldn't_ be an argument for `spawnLuopan` in the Geo- spell scripts. Luopan models are just base model ID (depending if they're supportive or offensive) + the spell's element. When our spell elements are fixed at the fundamental level, all models can be generated by `spawnLuopan` itself. (Although _how_ we inform spawnLuopan of the luopan's intended element will be a _different_ matter.

scripts/globals/geo.lua
```diff
@@ -0,0 +1,26 @@
...
+    local petID  = luopan:getID()
+
+    -- Attach effect
+    luopan:addStatusEffectEx(tpz.effect.COLURE_ACTIVE, tpz.effect.COLURE_ACTIVE, 0, 3, 0, tick_effect, tick_power, target_type, tpz.effectFlag.AURA)
+    -- Save the mp cost for use with Full Circle.
+    luopan:setLocalVar("MP_COST", spell_cost)
```
ibm2431: It'd be real neat if we could pass the spell itself into `spawnLuopan` so we could just ping it to assign its cost instead of needing to pass it into `spawnLuopan`. We could get the luopan's element that way too. But I don't thing spells are currently structured to allow that yet.

**edit** Adding bindings while I'm typing comments is cheating, Zach! 😉

scripts/globals/spells/geo-poison.lua
```diff
@@ -0,0 +1,27 @@
...
+    elseif not caster:canUseMisc(tpz.zoneMisc.PET) then
+        return tpz.msg.basic.CANT_BE_USED_IN_AREA
+    else
+        return 0
+    end
+end
+
+function onSpellCast(caster, target, spell)
+    local geo_skill = caster:getCharSkillLevel(tpz.skill.GEOMANCY)
+    local spellCost = 37
```
ibm2431: Because I'm guessing we both dislike the need for this~

ibm2431 (commented):
My request to overeager people: Don't go crazy on bubbles. There's some core things that can be improved before these bubbles are subject to mass copy-pastes~

src/map/utils/petutils.cpp
zach2good: Agreed!

src/map/utils/petutils.cpp
zach2good: I moved all of that configuration stuff into Lua (and the entity flags to SQL), much cleaner!

scripts/globals/geo.lua
zach2good: I've added some big comments explaining how this should work in the future 👍

ibm2431 (approved):

PR #1328, opened 2020-10-11
TOAU44: Fix black screen mid-fight

### LINKED ISSUE #1180: ToAU 44 Nashmeira's Plea
**_I have:_**

Issues with ToAU 44:

- when you first click on the blank target in Nyzul, if you cancel and come back, you will no longer get the CS
  (you will skip it and need to directly click on the entry door)
- you can't call trusts
- when you have reached the 50% on Alexander, the intermediate cutscene triggers, but ends in a black screen.
  You will need to force close the client to exit this CS (you do get hit meanwhile, so will eventually die)

zach2good:
**_I affirm:_**

Fixes: https://github.com/project-topaz/topaz/issues/1180
Superceeds: https://github.com/project-topaz/topaz/pull/1260

The order of events was wrong, so you ended up getting attacked while you were in CS. This ends up breaking your cs and results in you being trapped in a black screen.

Also discovered that `entityvsetPos()` doesn't correctly accept a table - like it is programmed to do. Gonna make an issue for that. Skirting around that here for now by using the table's members.

### DIFF
--- a/scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
+++ b/scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
@@ -39,7 +39,7 @@ function onInstanceProgressUpdate(instance, progress)
         DespawnMob(ID.mob[59].RAZFAHD, instance)
         for i, v in pairs(chars) do
             v:startEvent(203)
-            v:setPos(entryPos.x, entryPos.y, entryPos.z, entryPos.r)
+            v:setPos(entryPos.x, entryPos.y, entryPos.z, entryPos.rot)
         end
         SpawnMob(ID.mob[59].ALEXANDER, instance)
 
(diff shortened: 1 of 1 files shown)


scripts/zones/Nyzul_Isle/instances/nashmeiras_plea.lua
```diff
@@ -34,12 +34,13 @@ end
 function onInstanceProgressUpdate(instance, progress)
     if (progress == 4) then
         local chars = instance:getChars()
+        local entryPos = instance:getEntryPos()
 
+        DespawnMob(ID.mob[59].RAZFAHD, instance)
         for i, v in pairs(chars) do
             v:startEvent(203)
+            v:setPos(entryPos.x, entryPos.y, entryPos.z, entryPos.rot)
```
ibm2431, line 42: Still have the `player:setPos(-444, -4, 420, 127)` down in the eventFinish~

zach2good: _Oof!_ Nice catch!

TeoTwawki (approved):
So the solution did turn out to be relocating the player out of aggro distance at the time we trip the cs (while not accidentally zoning the player out of the instance :+1: by including a zoneID in our setPos)

ibm2431 (approved):

zach2good:
Yeah, things just got confused by the setPos binding being screwwy with the return table from getEntryPos :(

PR #1334, opened 2020-10-12
Initial Dockerfile and docker-compose.yml

NassirAlawar:
I have added a way to launch a server using docker. I have included instructions in the DOCKER file. I also included a third party script called wait-for-it and the license to go with it. The docker server uses the configuration found in conf/docker with the key difference being the mysql_host is set to "db" so the two services can communicate. Lastly I included the adminer service which allows for easy access to the database through the web on localhost:8080

**_I affirm:_**


zach2good:
Congrats on your first PR! Thanks for these! 

I had a quick glance through all the docker stuff (but not the readme yet) and they're pretty much the level of simplicity I was wanting out of something like this, good job!

~~I had a look through the copied conf files, and I didn't _see_ anything special apart from filled out login details?~~ db, got it!

I also wonder, do we need this `wait-for-it.sh` thing in the repo? Is there no way we can call out for it when we need it and cache it?

Would the content in this PR help you? 
https://github.com/project-topaz/topaz/pull/1202

I've been holding off because I don't know what names I should be using (`MYSQL_PWD` vs `DB_PASS` etc.), would this make your life easier and allow you to not have to edit conf files?

NassirAlawar:
Thanks, hopefully more to come :)
That other PR is interesting for sure. That would simplify it since we could use the default config then just override the mysql_host and set it do "db". I could set it the same way I set this one `ENV DEBIAN_FRONTEND=noninteractive`. Let me know if you'd like me to change it, I can wait for that to get merged in.

As for `wait-for-it.sh` I could write something simplier. I mostly grabbed it as the easiest solution from docker's website. If you'd prefer I can work on writing something simple specifically for this use case. Could also pull his repo but he has other files in there other than just the license and that script but that seemed like overkill

zach2good:
You might be able to get around using `wait-for-it` with this stuff we have in the CI scripts:
https://github.com/project-topaz/topaz/blob/release/.github/workflows/build.yml#L154

Admittedly, a lot of the CI script stuff is cobbled together copy-paste, because I don't yaml 👀

NassirAlawar:
I can try it out tomorrow and see if I can come up with something in-house. The `"mysqladmin ping"` might be enough but I think it will still need to be a script. I don't think docker supports doing scripts like this in the docker-compose file ```run: |
        mysql -h 127.0.0.1 -uroot -proot -e "SHOW DATABASES"``` but I'll give that a shot too

zach2good:
I'm gonna sort out that other PR I listed later today (I hope), while that's happening - can you investigate the mounting of folders in docker images?
The thing I'm concerned about is this:
The current linux build, or running `cmake .` polutes the source tree - which is bad practice. We do this on CI because that filesystem will get nuked after run. We don't want to pollute the filesystem of the docker image's host. 

Solvable by doing an "out of tree" build:
```
mkdir build
cd build
cmake ..
cd ..
```

The CMake scripts are set up to put the finished binaries in the root folder anyway, so you don't have to worry about relative paths etc.

zach2good:
https://github.com/project-topaz/topaz/pull/1202 has now been merged, should make your life a little easier 👍

NassirAlawar:
Sure I should be able to do the build change 👍 I didn't get a chance to look at the wait-for-it yesterday but I should have time for all three today

NassirAlawar:
I've implemented the env variables settings and moved the build to a /build folder which I delete after it completes.
I think I can look into the wait-for-it tonight

NassirAlawar:
Let me know if there's anything else, I've solved all the issues listed and tested the build. I had to make a change to the wait script in order to bring up the server in the correct order with proper delay. I believe in the future we can make the server logic smarter and remove this wait script by allowing the server to retry connections.

I have tested and I am able to launch the server, create an account, create a character and log into the world 👍

zach2good (approved):
Looks good! I think this is a good starting point, and if people want to refine this later that's up to them. 

Before merging, would it be possible to move all docker related files into a ./docker/ folder in the repo root? Would that ruin everything?

My computer just doesn't want to enable hardware assisted virtualization in any way (BIOS, features etc.). So I did a quick test in CI in my fork using this:
```
  Docker_Build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v1
      with:
        fetch-depth: 0
    - name: Build the docker-compose stack
      run: | 
        docker-compose up -d
    - name: Check running containers
      run: docker ps
```
https://github.com/zach2good/topaz/runs/1257201358?check_suite_focus=true

All seems good 👍

NassirAlawar:
Sure, I can try that tomorrow, I think it's possible if I update the paths

NassirAlawar:
Took a look at moving everything to a docker folder, I don't think it's as trivial as it seems. For security reasons docker-compose cannot access any directory above the current directory.
I also tried leaving only the docker-compose in the topaz folder then moved everything else to a docker folder and pointed it to the Dockerfile but I got errors when trying to do `docker-compose up`. There might be ways around the directory part with additional command line args but not sure what's causing the other errors.

zach2good:
In that case I think we're good to go, DOCKER AWAY 🐳

NassirAlawar:
mergedparrot

PR #1341, opened 2020-10-12
Add step-down delays to Trust healing ticks

### LINKED ISSUE #1337: Trust HP/MP Tick Incorrect
**_I have:_**

From @MarianArlt 

I'll post this here to avoid even more issues. People are frustrated about the further Trust nerfs. I don't want to comment on that. I want to share a recent experience from trial retail.
Ironically I thought that Trusts were slightly OP with their MP refresh, so I started to watch and note down on retail (I created the trial account for the rank bar issue). I wanted to take a video today but it seems my trial expired.

I noted things down on good old paper and only had Mihli Aliapoh, Adelheid and Excenmille as lab rats.
I found that the following was true:

The first MP tick after disengaging was the longest. I got inconsistent notes, probably because of network lags I didn't notice at the time. Sometimes it didn't tick until after 25 seconds (pretty positive that was lag) and the fastest I got on several notes was 15 seconds.

The second and the third tick were 10 seconds.

All ticks after that were 3 seconds.

The first three ticks were a lower amount (you could call it the base amount) then they got doubled by the fourth tick.

The MP regenerated in that fashion was not a flat 3% but rather it increased by the Trust's level.

At Lv17 Mihli Aliapoh would regenerate 3mp for the first three ticks and then 6mp for every tick after that.
At Lv18 she would regenerate 4mp (2.45% of her total 163mp) and then 8mp for every tick after that.
At Lv19 she would regenerate 5mp (2.94% of her total 170mp) and then 10mp for every tick after that.
Adelheid was regenerating the same amount despite having less in total (first three ticks 2.68% of her total 149mp at Lv18).

I'm not 100% sure (cause I missed to write down his Lvl) but I think the note of Excenmille shows that at Lv18 he would regenerate 1 base mp less than the mages but with the same pattern.

using /heal did not affect neither the tick pattern nor the amount of mp restored.

Generally speaking healing was rather conservative and MP aware. I would not receive cure instantly when dropping to yellow. Rather a few HP below that.

I would not get healed at all for quite a while if the Trust had very little MP

Maybe more retail people could chime in with additional and higher lvl Trust observations. @jarmengaud? This is from a trial played this October.

zach2good:
**_I affirm:_**

Fixes #1337 

After combat, trust healing ticks will come in after 15s, 10s, 10s and then every 3s.
This counter is reset if the trust is hit, so they can't tank while being hit by stuff.

Checked against retail by counting in my head when the ticks happen, and writing down the amounts so I can work out the differences.
They're all 5% (still)

### DIFF
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -34,7 +34,8 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "../../recast_container.h"
 #include "../../mob_spell_container.h"
 
-CTrustController::CTrustController(CCharEntity* PChar, CTrustEntity* PTrust) : CMobController(PTrust)
+CTrustController::CTrustController(CCharEntity* PChar, CTrustEntity* PTrust)
+: CMobController(PTrust)
 {
     m_GambitsContainer = std::make_unique<gambits::CGambitsContainer>(PTrust);
 }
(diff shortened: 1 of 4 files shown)


TeoTwawki:
_in your head?_

stop watches are even on cell phones these days man, don't do that to yourself.

TeoTwawki (approved):
I've probably never seen that 4th tick. my trusts either heal full by the 3rd or have died. or I'd dismiss them. still I am glad I get sweet vindication on that "longer 1st tick".

PR #1344, opened 2020-10-12
Make BLM Trusts cast more conservatively

zach2good:
**_I affirm:_**

Went onto retail to do some testing, there is a much longer gap between when BLM Trusts will freely nuke. Left SCs alone.


TeoTwawki:
> there is a much longer gap between when BLM Trusts will freely nuke

There are likely differences between diff ones or depending what other actions they have queued up making it hard to pin down exactly what the base oughta be and how to modify it. This change brings it closer for sure.

PR #1347, opened 2020-10-13
Add safety check for PNotoriety container

zach2good:
**_I affirm:_**

Got a report of there being lingering entries in PNotorietyContainer, meaning you'd have to zone to clear it.
This makes player:hasEntity() wrongly return true, when the player isn't in trouble
It was in Castle O [S], so I'm guessing it has something to do with the broken giant beastmen parties.

Repro'd in canary by running around in Castle O [S] and killing Yagudos, and then trying to summon trusts -> rejected

Proven fixed by merging this branch into canary, doing the same, then being able to summon trusts when I've killed everything around me and I have no obvious aggro.

Also put a breakpoint on the `remove()` call to make sure invalid mobs were being taken out of the container

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -77,7 +77,6 @@
 #include "../packets/char_job_extra.h"
 #include "../packets/status_effects.h"
 #include "../mobskill.h"
-#include "../notoriety_container.h"
 
 CCharEntity::CCharEntity()
 {
@@ -529,8 +528,6 @@ void CCharEntity::Tick(time_point tick)
     {
         gardenutils::UpdateGardening(this, true);
     }
-
-    PNotorietyContainer->tryClear();
 }
 
 void CCharEntity::PostTick()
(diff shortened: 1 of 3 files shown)


ibm2431:
Each tick seems a little overkill for what's probably a bug elsewhere (the wrong mob got on the list somehow). Could we safely lighten the load by calling tryClear inside checks for notoriety (ie: hasEnmity)?

```
Do I have enmity?
Remove any invalid mobs from my list
Mobs remaining? Return Yes
```

ibm2431:
Now I'm getting crazy thoughts like the iterator for the container always checking for the player in the mob's enmity container, and removing the entry from the notoriety container if there's no corresponding match, so that the notoriety container is always "self cleaning" as its stepped through.

But feel free to ignore the mad science.

zach2good:
> Could we safely lighten the load by calling tryClear inside checks for notoriety (ie: hasEnmity)?

This is the exact use case that the bug appears in (so far), so this makes a lot of sense! This is also the _only_ place that the container is iterated by.

ibm2431 (approved):

Issue #480, opened 2020-04-09
Bridge in Ro'maeve dissapearing [and too high]


kaincenteno:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Two weird things are happening with the bridges:
* the highest bridge (the one right next to the fountain), dissapears entirely if you walk away from it by at least 50 meters.
* they raise a bit too much when night time on a full moon. Unsure if its retail accurate or if pos needs to decrease a bit.

TeoTwawki:
> the highest bridge (the one right next to the fountain), dissapears entirely if you walk away from it by at least 50 meters.

Entity packets not currently sent when farther than that. For most objects that should be retail accurate but then when you come in range they should "repop" on your screen. may need a way to make exceptions to that. Not sure what retail does in this case for these specific objects.

> hey raise a bit too much when night time on a full moon. Unsure if its retail accurate or if pos needs to decrease a bit.

Are you able to step onto them?

kaincenteno:
@TeoTwawki 
yeah once u get back to it the whole bridge is gone lol. it only reappears once the moon gate doors close

Yes you are able to step on them but it's raise too high it makes the taru feet dissapear lol

TeoTwawki:
Its eating taru feet? clearly a secret mimic, and a racial hate crime.  We can't let that stand. To war! [We'll show those packets!](https://github.com/project-topaz/topaz/blob/4d09173b57dd5867dd3997436fa3ee66eeea0fbb/src/map/zone_entities.cpp#L787)

TeoTwawki:
https://github.com/project-topaz/topaz/blob/4d09173b57dd5867dd3997436fa3ee66eeea0fbb/src/map/zone_entities.cpp#L787

kaincenteno:
@TeoTwawki  hopefully someone takes a look at this =) funny thing is that it only happens to the top bridge but not the bottom.

This should help also with the gates in the desert (those high towers) for opening that door that u need to hit on 4 switches in the map. cause if u are far away the look as if they are off, but if u get close they look as if the switch is on

PR #1354, opened 2020-10-13
Hourly check for Ro'Maeve full moon

MarianArlt:
**_I affirm:_**

This PR would successfully activate Ro'Maeve's bridges, gates and the fountain not only at exactly 6pm but at any hour that hits full moon between 6pm and (not including) 6am for the remaining amount of time until 6am. The current logic in release only once can activate the whole area at exactly 6pm not thereafter. This is not desirable.
~~It also adds the missing weather condition. The area must have no weather for the full moon to affect the mentioned objects.~~
Furthermore when cleaning up at 6am only the two Moongates should become targetable again.

This was a bummer to debug and test cause the !time command apparently throws off the sync between the in-game shown /clock and the server clock, but nonetheless:
- [x] Tested working locally at several hours, moon phases and weather conditions.

### DIFF
--- a/scripts/zones/RoMaeve/Zone.lua
+++ b/scripts/zones/RoMaeve/Zone.lua
@@ -35,20 +35,38 @@ function onRegionEnter(player, region)
 end
 
 function onGameHour(zone)
-    local hour = VanadielHour()
+    local vanadielHour = VanadielHour()
+    
+    -- Make Ro'Maeve come to life between 6pm and 6am during a full moon
+    if IsMoonFull() and (vanadielHour >= 18 or vanadielHour < 6) then
+        local moongate1 = GetNPCByID(ID.npc.MOONGATE_OFFSET)
+        local moongate2 = GetNPCByID(ID.npc.MOONGATE_OFFSET + 1)
 
-    if IsMoonFull() and hour == 18 then
-        for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET+7 do
-            GetNPCByID(i):openDoor(1728)
-            GetNPCByID(i):untargetable(true)
+        if moongate1:getLocalVar("romaeveActive") == 0 then
+            -- Loop over the affected NPCs: Moongates, bridges and fountain
+            for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET + 7 do
+                GetNPCByID(i):setAnimation(tpz.anim.OPEN_DOOR) -- Open them
+            end
+            moongate2:untargetable(true)
+            moongate1:untargetable(true)
+            moongate1:setLocalVar("romaeveActive", 1) -- Make this loop unavailable after firing
         end
-    elseif hour == 6 then
-        for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET+7 do
-            GetNPCByID(i):untargetable(false)
+
+    -- Clean up at 6am
+    elseif vanadielHour == 6 then
+        local moongate1 = GetNPCByID(ID.npc.MOONGATE_OFFSET)
+        local moongate2 = GetNPCByID(ID.npc.MOONGATE_OFFSET + 1)
+
+        if moongate1:getLocalVar("romaeveActive") == 1 then
+            for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET + 7 do
+                GetNPCByID(i):setAnimation(tpz.anim.CLOSE_DOOR)
+            end
+            moongate2:untargetable(false)
+            moongate1:untargetable(false)
+            moongate1:setLocalVar("romaeveActive", 0) -- Make loop available again
         end
     end
 end
-
 
 function onEventUpdate(player, csid, option)
 end


MarianArlt:
~~Please don't merge yet, I think I have a typo.~~

wrenffxi:
We used to have the weather requirement coded, but removed it; I thought this was because specific weather was no longer required in retail.
https://github.com/DarkstarProject/darkstar/issues/3957

MarianArlt:
Ty for the reference @wrenffxi I almost feared as much...
Makes absolutely no sense to see the reflection of a nice full moon in a clear pond during a heavy clouded rain...I think some QoL changes are sad.
I'll take it out.

MarianArlt:
~~Took out the weather check (makes me sad) and removed `math.abs()` from one of the calculations cause not necessary. The subtraction will reliably always be negative, so just multiply by negative.~~ Outdated, see below.

Maybe I can raise awareness for https://github.com/project-topaz/topaz/issues/480 which is related to the content of this PR but not affected by it.

scripts/zones/RoMaeve/Zone.lua
```diff
@@ -35,21 +35,43 @@ function onRegionEnter(player, region)
...
+    local romaeveActive = moongate:getLocalVar("romaeveActive")
+    
+    -- Make Ro'Maeve come to life between 6pm and 6am during a full moon
+    if
+        romaeveActive == 0 and IsMoonFull() and
+        (VanadielHour() >= 18 or VanadielHour() < 6)
+    then
+        -- Calculate remaining seconds till 6am from this hour
+        local secondsTillMorning
+        if VanadielHour() < 6 then
```
TeoTwawki: We really should keep the variable rather than repeatedly call the same function. OnGameHour is going to re execute every hour, so the results isn’t able to change before reaching the next place it is checked

MarianArlt: It won't be called repeatedly at all since we set `setLocalVar("romaeveActive", 1)` though? none of all of that will be called again after firing once. If anything we could put the two locals `moongate` and `romaeveActive` outside of the `onGameHour()` block indeed, so they're not called every hour for no reason.

TeoTwawki: you have a condition inside of a condition that both ask what the hour is. you can store the function call result instead of calling it twice.

TeoTwawki: I commented at the line where the 2nd one happens - I'm not talking a bout your loop. But while we're at it, why are we looping for hrs left instead of using >= and <= since we know the exact hr ranges on both sides of 0:00 and 24:00..does the change not land on exactly an hr mark?

MarianArlt: Oh yeah, that's a different thing. Makes sense.

TeoTwawki: if not active and full moon and hour <6  then.. If hour <6  :thinking:

MarianArlt: ~~On hold for now after talking to @TeoTwawki some more. More changes incoming.~~ Removed all the bloat that was written to calculate the seconds left until 6am needed for `openDoor()`. Now uses `:setAnimation()` instead. Done.
- [x] Tested working locally

scripts/zones/RoMaeve/Zone.lua
```diff
@@ -34,22 +34,32 @@ end
 function onRegionEnter(player, region)
 end
 
+-- Make Ro'Maeve come to life between 6pm and 6am during a full moon
+local moongate1 = GetNPCByID(ID.npc.MOONGATE_OFFSET)
+local moongate2 = GetNPCByID(ID.npc.MOONGATE_OFFSET + 1)
+local vanadielHour = VanadielHour()
```
TeoTwawki: you'll want to define your variables inside onGameHour rather than above/before it, (they aren't being used outside it, so by putting them with it we're restricting them to the [scope ](http://lua-users.org/wiki/ScopeTutorial)they need to be in)

I would also keep the original `local hour = VanadielHour()` that was here before, since there are no earth hours here to accidentally confuse it with.

MarianArlt: Are you sure about this? Wouldn't that just mean unnecessary requests being made every 144s completely despite the fact whether it is full moon or whatever hour? For all the lifetime of the server? Because if its not full moon, none of this will get called if not after a server boot. Also I feel that renaming the variable kinda goes against what Zach has been pushing were the code itself should be as explicit as possible at all times.

What I do see is that we should add `romaeveActive == 1` to the clean up condition, so it also only gets called in that specific event, not every day. I had this clear every day rather than unconditionally because I didn't have the flag bit before.

TeoTwawki: The only ones that wouldn't be being called anyway are the 2 getByID, which even though it is probably ok the vars themselves can be inside a conditional to prevent running them when not needed, but when you have them outside the function like this they exist in a larger scope which I did not catch when you showed me in discord (sorry)

proposed adjustment:
```lua
function onGameHour(zone)
    -- Make Ro'Maeve come to life between 6pm and 6am during a full moon
    local romaeveActive = moongate1:getLocalVar("romaeveActive")
    local vanadielHour = VanadielHour()

    if romaeveActive == 0 and IsMoonFull() and (vanadielHour >= 18 or vanadielHour < 6) then
        local moongate1 = GetNPCByID(ID.npc.MOONGATE_OFFSET)
        local moongate2 = GetNPCByID(ID.npc.MOONGATE_OFFSET + 1)

        -- Loop over the affected NPCs: Moongates, bridges and fountain
        for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET + 7 do
            GetNPCByID(i):setAnimation(tpz.anim.OPEN_DOOR) -- Open them
        end
        moongate2:untargetable(true)
        moongate1:untargetable(true)
        moongate1:setLocalVar("romaeveActive", 1) -- Make this loop unavailable after firing

    -- Clean up at 6am
    elseif romaeveActive == 1 and vanadielHour == 6 then
        local moongate1 = GetNPCByID(ID.npc.MOONGATE_OFFSET)
        local moongate2 = GetNPCByID(ID.npc.MOONGATE_OFFSET + 1)
        for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET + 7 do
            GetNPCByID(i):setAnimation(tpz.anim.CLOSE_DOOR)
        end
        moongate2:untargetable(false)
        moongate1:untargetable(false)
        moongate1:setLocalVar("romaeveActive", 0) -- Make loop available again
    end
end
```
it won't matter that we're doing a local define of those npc's in both conditions, because only one of the 2 conditions can be trues and each one is inside the scope of that condition (so no collision, and only one set is used). 

hour and romaeveActive are going to be check every hours anyway.

MarianArlt: I for one like that solution very much. Thank you for your patience @TeoTwawki 
Changes were made as proposed.

scripts/zones/RoMaeve/Zone.lua
```diff
@@ -35,21 +35,36 @@ function onRegionEnter(player, region)
 end
 
 function onGameHour(zone)
-    local hour = VanadielHour()
+    -- Make Ro'Maeve come to life between 6pm and 6am during a full moon
+    local romaeveActive = moongate1:getLocalVar("romaeveActive")
```
ibm2431: moongate1 hasn't been defined yet~!

ibm2431 (changes requested):

scripts/zones/RoMaeve/Zone.lua
TeoTwawki: woopsie

TeoTwawki: my fault

TeoTwawki:
```lua
function onGameHour(zone)
    local hour = VanadielHour()

    -- Make Ro'Maeve come to life between 6pm and 6am during a full moon
    if IsMoonFull() and (hour >= 18 or hour < 6) then
        local moongate1 = GetNPCByID(ID.npc.MOONGATE_OFFSET)
        local moongate2 = GetNPCByID(ID.npc.MOONGATE_OFFSET + 1)
        if moongate1:getLocalVar("romaeveActive") == 0 then

            -- Loop over the affected NPCs: Moongates, bridges and fountain
            for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET + 7 do
                GetNPCByID(i):setAnimation(tpz.anim.OPEN_DOOR) -- Open them
            end
            moongate2:untargetable(true)
            moongate1:untargetable(true)
            moongate1:setLocalVar("romaeveActive", 1) -- Make this loop unavailable after firing
        end

    -- Clean up at 6am
    elseif hour == 6 then
        local moongate1 = GetNPCByID(ID.npc.MOONGATE_OFFSET)
        local moongate2 = GetNPCByID(ID.npc.MOONGATE_OFFSET + 1)
        if moongate1:getLocalVar("romaeveActive") == 1 then
            for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET + 7 do
                GetNPCByID(i):setAnimation(tpz.anim.CLOSE_DOOR)
            end
            moongate2:untargetable(false)
            moongate1:untargetable(false)
            moongate1:setLocalVar("romaeveActive", 0) -- Make loop available again
        end
    end
end
```
fixed version of what I wrote before

MarianArlt: Sorry, was sleeping myself there. Changed as proposed, kept `vanadielHour` though.

ibm2431 (approved):

TeoTwawki (approved):

scripts/zones/RoMaeve/Zone.lua
```diff
@@ -35,21 +35,39 @@ function onRegionEnter(player, region)
...
+    -- Make Ro'Maeve come to life between 6pm and 6am during a full moon
+    if IsMoonFull() and (vanadielHour >= 18 or vanadielHour < 6) then
+        local moongate1 = GetNPCByID(ID.npc.MOONGATE_OFFSET)
+        local moongate2 = GetNPCByID(ID.npc.MOONGATE_OFFSET + 1)
 
-    if IsMoonFull() and hour == 18 then
-        for i = ID.npc.MOONGATE_OFFSET, ID.npc.MOONGATE_OFFSET+7 do
-            GetNPCByID(i):openDoor(1728)
-            GetNPCByID(i):untargetable(true)
+        if moongate1:getLocalVar("romaeveActive") == 0 then
```
TeoTwawki, line 45: note to self: look into making variables on the zone object itself a thing. if for some reason this one NPC gets lost by something as trivial as an ID shift, the whole thing breaks.

PR #1357, opened 2020-10-14
Add proper events to ??? in Garlaige Citadel for THF AF feet

MarianArlt:
**_I affirm:_**

For THF AF feet quest "Hitting the Marquisate" in Garlaige Citadel:

When clicking a ??? and it's the right one in order, the player should receive the proper event which asks him/her whether or not to use the Bomb Incense key item. After choosing yes that ??? should become unusable for this purpose and the player can search for the next one. This way the player gets a proper clue on which ??? is the next.

Also the bomb NM should spawn with a certain delay or else it will most probably instantly self-destruct (since the only way to avoid this is to know this mechanic and run instantly after the CS into the correct corner), in which case the player won't get the necessary item dropped. Guides, and even the event's dialogue suggest this behavior.


MarianArlt:
So luckily I found [a video from 2016](https://www.youtube.com/watch?v=X3Ou2nm5qCI) specific to this quest. There were even more messages missing. I added those, for which there were no JSON dumps and manually searching through the scenes in game didn't yield anything, so I used `messageSpecial`. The spawn seems to be around 5 seconds after the cutscene. I think that's a fair value.

- [x] Tested working locally

Doesn't build with various connection errors to OpenSuse..., is that from my end?

zach2good (approved):
Seems fine to me 👍

PR #1358, opened 2020-10-14
Fix: legacy of the lost

bope12:
Currently the legacy of the lost bcnm can be entered with anyone who is currently on the mission and anyone who is in the party will be skipped ahead to the next mission on a win. This at least puts a check on battlefield win for the credit, but does not address people being able to join on a lesser mission.

**_I affirm:_**

### DIFF
--- a/scripts/zones/Talacca_Cove/bcnms/legacy_of_the_lost.lua
+++ b/scripts/zones/Talacca_Cove/bcnms/legacy_of_the_lost.lua
@@ -31,7 +31,7 @@ function onEventUpdate(player, csid, option)
 end
 
 function onEventFinish(player, csid, option)
-    if csid == 32001 and not player:hasCompletedMission(TOAU, tpz.mission.id.toau.LEGACY_OF_THE_LOST) then
+    if csid == 32001 and player:getCurrentMission(TOAU) == tpz.mission.id.toau.LEGACY_OF_THE_LOST then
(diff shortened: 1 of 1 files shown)


scripts/zones/Talacca_Cove/bcnms/legacy_of_the_lost.lua
```diff
@@ -31,7 +31,7 @@ function onEventUpdate(player, csid, option)
 end
 
 function onEventFinish(player, csid, option)
-    if csid == 32001 and not player:hasCompletedMission(TOAU, tpz.mission.id.toau.LEGACY_OF_THE_LOST) then
+    if csid == 32001 and player:getCurrentMission(TOAU) == tpz.mission.id.toau.LEGACY_OF_THE_LOST then
         player:completeMission(TOAU, tpz.mission.id.toau.LEGACY_OF_THE_LOST)
         player:setTitle(tpz.title.GESSHOS_MERCY)
```
TeoTwawki, line 35: you probably get the title even if just helping, so we should probably move the mission check to a conditional inside the csid check and then put the mission progression commands inside that

bope12: Changed unless someone can confirmed otherwise

TeoTwawki (approved):

ibm2431 (approved):

PR #1363, opened 2020-10-15
Make NM Lottery parameters configurable

Kreidos:
I'm sure there are plenty of servers who believe the lottery rates are too high or too low for their liking. This adds 2 flags to adjust the Chance and Cooldown on Lottery NMs.

This allows for some interesting options, such as turning the rates down and setting cooldowns to 0 thus having all NMs be a rare true-lottery. Or of course you could disable all Lottery spawns entirely if you want, I ain't your boss. :shrug: 

**_I affirm:_**

### DIFF
--- a/scripts/globals/mobs.lua
+++ b/scripts/globals/mobs.lua
@@ -61,6 +61,13 @@ end
 -- potential lottery placeholder was killed
 tpz.mob.phOnDespawn = function(ph, phList, chance, cooldown, immediate)
     if type(immediate) ~= "boolean" then immediate = false end
+    
+    if NM_LotteryChance then
+        chance = NM_LotteryChance >= 0 and (chance * NM_LotteryChance) or 100
+    end
+    if NM_LotteryCooldown then
+        cooldown = NM_LotteryCooldown >= 0 and (cooldown * NM_LotteryCooldown) or cooldown
+    end
 
     local phId = ph:getID()
     local nmId = phList[phId]
--- a/scripts/globals/mobs.lua
+++ b/scripts/globals/mobs.lua
@@ -76,8 +76,9 @@ tpz.mob.phOnDespawn = function(ph, phList, chance, cooldown, immediate)
         local nm = GetMobByID(nmId)
         if nm ~= nil then
             local pop = nm:getLocalVar("pop")
-
-            if os.time() > pop and not lotteryPrimed(phList) and math.random(100) <= chance then
+            
+            chance = math.ceil(chance * 10) -- chance / 1000.
+            if os.time() > pop and not lotteryPrimed(phList) and math.random(1000) <= chance then
 
                 -- on PH death, replace PH repop with NM repop
                 DisallowRespawn(phId, true)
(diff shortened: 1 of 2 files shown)


TeoTwawki:
This is similar to what I do on my server, except that I imposed a minimum time before applying the reduction on repop time (didn't want any lotto less than 20 min long while universally cutting it)

Kreidos:
> This is similar to what I do on my server, except that I imposed a minimum time before applying the reduction on repop time (didn't want any lotto less than 20 min long while universally cutting it)

I like this idea myself. I was concerned about having too many flags (wanted it short and simple), but if it's something enough people care about I could add a minimum time as another flag.

**edit** - I know some servers are masochists, so I kind of assumed most people would be using it to lower the rates, ha ha.

Apocarypse (approved):

zach2good (approved):
Idea is good, maths looks good, that settings file is a nightmare!

Since these lua settings are mega-global, could you change the settings name to be ALL_CAPS? `NM_LOTTERY_CHANCE` etc.

If you're feeling _really_ generous you could change the other non-conforming unclear settings to be the same, but that's a lot of work and totally up to you

ibm2431 (commented):
Seconding logic looking fine and ALL_CAPS~

_I'd_ be okay with min and max time settings, but they can be added later (ie: not cause a debate in this PR).

ibm2431:
Since I saw some discussion on Discord.

For future reference, _my_ general opinion (others may vary) is:

Measurement is:
1. How many lines of existing code does the setting effect? Lower is better. _versus_
2. How many in-game events does the setting have an effect on? Higher is better.

Examples:
- Bad: Setting which requires a huge code switch with 20-50+ lines of code. Not a good setting.
- Bad: Setting which [only affects one thing an NPC does that players may not even interact with](https://github.com/project-topaz/topaz/blob/release/scripts/globals/settings.lua#L153). Also not a good setting.
- Good: Setting which modifies only 2 lines of code and has an effect any time a mob is killed. Potentially a good setting.

EXP_RATE, NM_LOTTERY_CHANCE, ENABLE_ROE, are all good settings. They have a huge effect on gameplay, yet are lightweight from a code maintenance perspective.

[AQUAVEIL_COUNTER](https://github.com/project-topaz/topaz/blob/release/scripts/globals/settings.lua#L134) and [BLINK_SHADOWS](https://github.com/project-topaz/topaz/blob/release/scripts/globals/settings.lua#L131) are bad settings. They do only affect one line, but they also [only affect one _event_](https://github.com/project-topaz/topaz/blob/release/scripts/globals/spells/blink.lua#L16) (casting of a single spell).

[USE_ADOULIN_WS_CHANGES](https://github.com/project-topaz/topaz/blob/release/scripts/globals/settings.lua#L82) is right on the cusp. It affects a lot (all weaponskill uses), but it's borderline cumbersome to maintain (if checks [in each WS file](https://github.com/project-topaz/topaz/blob/release/scripts/globals/weaponskills/aeolian_edge.lua#L22-L31), if checks inside [WS calculation functions](https://github.com/project-topaz/topaz/blob/release/scripts/globals/weaponskills.lua#L79-L83)). At some point the logic needs restructuring so that scripts treat SOA changes as _default_ and pre-SOA under a legacy setting. But it's probably for the best we don't breathe on WS too hard until we're ready to redo them from scratch.

Basically, I measure the worth of a setting as the biggest bang for its lines of code.

But I will insist that anything to do with a setting must be capable of being used by the vanilla current client. If you need an old client to be able to make use of a setting, it'd be a hard no from me.

ibm2431 (approved):

PR #1365, opened 2020-10-15
Exploits List Part 1

zach2good:
**DON'T PANIC**

Most servers have had these fixes for 12 or so days. If you're a server operator who wasn't previously aware of this, you need to contact me on Discord.

We were very kindly donated a list of common exploits that can be used to crash a server, warp around freely, cheese craft attempts etc. We've been working on patching these, and this is the first part - **the ones easiest for a bad actor to achieve**. These are all reproduced, and then verified fixed.

_The crashes can be forced on a server with a single operation from either of the two common windowing programs._
 
Any questions -> Discord.

EDIT: Before I merge I'll swap out the `ShowWarning` for `ShowExploit`

**_I affirm:_**

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -441,7 +441,7 @@ void SmallPacket0x00D(map_session_data_t* session, CCharEntity* PChar, CBasicPac
             // The broken rod can never be lost in a normal failed synth. It will only be lost if the synth is
             // interrupted in some way, such as by being attacked or moving to another area (e.g. ship docking).
 
-            ShowWarning(CL_YELLOW "SmallPacket0x00D: %s attempting to zone in the middle of a synth, failing their synth!\n" CL_RESET, PChar->GetName());
+            ShowExploit(CL_YELLOW "SmallPacket0x00D: %s attempting to zone in the middle of a synth, failing their synth!\n" CL_RESET, PChar->GetName());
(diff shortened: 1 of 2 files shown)


src/map/packet_system.cpp
```diff
@@ -2862,6 +2906,19 @@ void SmallPacket0x05D(map_session_data_t* session, CCharEntity* PChar, CBasicPac
     const auto EmoteID = data.ref<Emote>(0x0A);
     const auto emoteMode = data.ref<EmoteMode>(0x0B);
 
+    // Rate limit emotes
+    auto lastEmoteTime = PChar->GetLocalVar("LastEmoteTime");
+    auto timeNowSeconds = std::chrono::time_point_cast<std::chrono::seconds>(server_clock::now());
+    if (lastEmoteTime == 0 || (timeNowSeconds.time_since_epoch().count() - lastEmoteTime) > 5)
```
TeoTwawki: whats the basis for it being 5 seconds? I know retail does limit the spammage speed of command sin general, just wondering what 6 seconds as opposed to say 3, or 9, or whatever (the source/reason)

EDIT:
>Any questions -> Discord.

Oops...

zach2good: Good question! I chose 5s entirely arbitrarily.

ibm2431 (approved):

PR #1371, opened 2020-10-16
Fix Supply Run Mission Crash

Era-Lusiphur:
CP value was removed from outposts table in error, which caused crashes when adding the CP reward for doing supply run missions. Simply restoring the old version of the table, which does not affect the recent fix to Outpost warping costs or any other function.

**_I affirm:_**


zach2good:
Whats the relation to the PR that removed these?
https://github.com/project-topaz/topaz/pull/1020

Era-Lusiphur:
> Whats the relation to the PR that removed these?
> #1020

Was discovered that the removal of the CP values causes a crash when finishing a supply run here:
https://github.com/project-topaz/topaz/blob/f3eabf8efd92216b018c37753a88bdf878b8af72/scripts/globals/conquest.lua#L1112

zach2good:
CI is dead because... Microsoft? Seems fine to me

PR #1372, opened 2020-10-16
Prefill Trust and Intermediate ROE Records

zach2good:
**_I affirm:_**

Need these for my trust branch - hitting level and speaking to ROE NPC when you have x sparks aren't implemented yet (I think?)

### DIFF
--- a/scripts/globals/roe_records.lua
+++ b/scripts/globals/roe_records.lua
@@ -1,4 +1,4 @@
--- This file is called by roe.lua and require()'s no file, it should not be require()'d by any 
+-- This file is called by roe.lua and require()'s no file, it should not be require()'d by any
 -- other lua scripts, which should instead require() roe.lua directly.
 
 local triggers = tpz.roe.triggers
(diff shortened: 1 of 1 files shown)


Kreidos:
I might recommend leaving them commented out until they're actually completeable, as their presence in the table is what allows people to take them. And just removing the `--` in a Trust branch commit when you are ready for them should still avoid any possible conflict issues from `release` if others want to work on other parts of the file.

zach2good:
Ah, I see what you mean!

Kreidos (approved):
I can't really argue with code that doesn't do anything! :man_shrugging: 
Table entries all look good.

ibm2431 (commented):
`Allman Braces` 👀

zach2good:
_Ooof_

At this point in `roe_records` life, changing the already consistent style to something else is too much of a pain in the ass. While I've personally seen the benefit of all this data being in a flexible lua table that we can refresh, I don't think it should _always_ be that way. Since it is a somewhat static (but large) set of data I'd like to one day see it plopped into the db. 

How that happens in the future is anyone's guess. Thats a problem for future developers.

Kreidos:
if someone can offer of a way of storing something as highly varied as RoE record data in a db without it turning into a nightmare, I'm all ears. (Look at the magian table for just a tiny taste of that life.) Particularly records that don't behave like all the other records (take/deal damage, complete X mission, timed/scheduled 4-hour records, etc.) Plus the fact that records can have arbitrary number of requirements... Do we just make 6 requirement columns and hope that's enough? One row per requirement and have thousands of more rows? 

I also don't feel the overhead of using a Lua table is as high as may be indicated.

Kreidos:
PS. use vim! I can fix the table pretty quick if it's that bothersome. I was under the impression that same-line open braces were pretty standard Lua style but I see that it's in the new style guide for Lua too.

PR #1373, opened 2020-10-16
Trust ROE Conditions

zach2good:
**_I affirm:_**

### DIFF
--- a/scripts/globals/spells/trust/adelheid.lua
+++ b/scripts/globals/spells/trust/adelheid.lua
@@ -1,6 +1,7 @@
 -----------------------------------------
 -- Trust: Adelheid
 -----------------------------------------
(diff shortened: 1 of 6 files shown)


Kreidos:
It's your own branch, but yup looks good! 
Having these lines without any accompanying record table entries in `roe_records.lua` also wouldn't cause any issues, as it'd be impossible for a player to have or take those records #'s in the first place.

Kreidos (approved):

zach2good:
> Having these lines without any accompanying record table entries in roe_records.lua also wouldn't cause any issues, as it'd be impossible for a player to have or take those records #'s in the first place.

Thats the confirmation I was hoping you'd say, thanks!

PR #1381, opened 2020-10-17
Player's main job must be THF to use thief's tools

rude-jerk:
**_I affirm:_**


zach2good (approved):
Yeah, docs all confirm using `Thief's Tools` should be THF-main only 👍

ibm2431 (approved):

PR #1389, opened 2020-10-18
Missing drops for Hydra

jarmengaud:
Missing drops for Hydra

Drop rates per ffxidb

**_I affirm:_**

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -16603,6 +16603,15 @@ INSERT INTO `mob_droplist` VALUES (3146,0,0,1000,15123,10); -- Valor Breeches
 INSERT INTO `mob_droplist` VALUES (3146,0,0,1000,15140,10); -- Monster Gaiters
 INSERT INTO `mob_droplist` VALUES (3146,0,0,1000,15142,10); -- Scout's Socks
 INSERT INTO `mob_droplist` VALUES (3146,0,0,1000,16360,10); -- Etoile Tights
+INSERT INTO `mob_droplist` VALUES (3147,0,0,1000,5564,1000); -- Hydra ; Hydra meat  210%
+INSERT INTO `mob_droplist` VALUES (3147,0,0,1000,5564,1000); -- Hydra ; Hydra meat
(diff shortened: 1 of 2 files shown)


sql/mob_droplist.sql
```diff
@@ -16603,6 +16603,15 @@ INSERT INTO `mob_droplist` VALUES (3146,0,0,1000,15123,10); -- Valor Breeches
 INSERT INTO `mob_droplist` VALUES (3146,0,0,1000,15140,10); -- Monster Gaiters
 INSERT INTO `mob_droplist` VALUES (3146,0,0,1000,15142,10); -- Scout's Socks
 INSERT INTO `mob_droplist` VALUES (3146,0,0,1000,16360,10); -- Etoile Tights
+INSERT INTO `mob_droplist` VALUES (3147,0,0,1000,5564,1000); -- Hydra ; Hydra meat  210%
```
zach2good: Purely a style thing, but using ';' as a delimiter here makes me a little uneasy because it's a valid character for SQL. Can you change this to one of the other styles in the file?
`(Hydra) Hydra Meat  210%`
`Hydra -- Hydra Meat  210%`

zach2good (changes requested):
When you say ffxidb, can you also provide the link that you're using to make reviewers time a little easier:
http://www.ffxidb.com/zones/51/hydra

For others passing by, what @jarmengaud has done here is good: they've looked at the average and TH0 values and ROUNDED DOWN, rather than being generous. 

If it turns out to be wrong, people will complain and it'll get fixed. If it was set too high, nobody will complain about a generous drop rate 🤷

ibm2431:
[24% is the highest TH0 bucket (Common)](https://www.bg-wiki.com/bg/Treasure_Hunter), they've got nothing to round up to~! 😉

The TH2 data that FFXIDB has is also useful for the torque and axe. They're right around 56%, which is what the TH2 Common bucket _should_ be. So I'm in full agreement that those are Common drops.

sql/mob_droplist.sql
jarmengaud: K done

zach2good (approved):

PR #1390, opened 2020-10-18
added hp/mp/tp absorb to twilight knife

sleepingkirby:
Adds twilight_knife.llua file and adds absorb twilight knife into it.

**_I affirm:_**

### DIFF
--- a/scripts/globals/items/twilight_knife.lua
+++ b/scripts/globals/items/twilight_knife.lua
@@ -1,0 +1,71 @@
+-----------------------------------------
+-- ID: 19132
+-- Item: Twilight Knife
+-- Additional Effect: HP, MP or TP Drain
+-----------------------------------------
+require("scripts/globals/status")
+require("scripts/globals/magic")
+require("scripts/globals/msg")
+-----------------------------------
+
+function onAdditionalEffect(player,target,damage)
+    -- no proc on undead
+    if (target:isUndead()) then
+        return 0,0,0
+    end    
+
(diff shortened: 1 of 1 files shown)


TeoTwawki (changes requested):
Please target https://github.com/project-topaz/topaz/tree/wip/effect-rework branch instead, and add the relevant modifiers instead of a new script. per item scripting is being done away with. The branch cannot be merged to canary and then release until the item mods are added.

```sql
NSERT INTO `item_mods` VALUES (19132,431,7); -- Twilight Knife: AddEffect Type HPMPTP_DRAIN
INSERT INTO `item_mods` VALUES (19132,499,21); -- SUBEFFECT_HP_DRAIN  (Teo will fix alternate animations later)
INSERT INTO `item_mods` VALUES (19132,500,10); -- AddEffect damage 10 (On my own knife on retail, I get same drain amount range for all 3 types)
INSERT INTO `item_mods` VALUES (19132,501, 100); -- AddEffect chance 100% (your script was always firing 1 of the 3, so I set 100 here)
INSERT INTO `item_mods` VALUES (19132,950,8); -- AddEffect element dark (for resist rates)
```
There is a cosmetic issue I did not think of in that only one mod can be set for the subeffect but on retailMP drain has a different color. I will resolve that in the global. I will also make independent rates for multi effects on items and their damages/drains amounts handled. For now, the above makes the item "work" in the additional effect branch.

PR #1395, opened 2020-10-19
Standardized POS destinations of players at all Survival Guides

EpicTaru:
This is to ensure that when a Survival Guide teleport is used, the player is placed directly next to a Guide (or ON the guide in navmesh limitations prohibit the placing next to the Guide).

Multiple times were noticed when using a Survival Guide the player was placed at a position near the Survival Guide but not directly next or ON the Guide.

This all began because of the below issue with the placement of the player at the Survival Guide in Romaeve.  In my post-code fix of that placement, I audited all of the other Survival Guide placements (to make sure that they were all working) and the oddity and non-standardness of player placement. At some, players were placed near, at some, players were placed far away, at some, players were placed directly ON the survival guide. This all /triggered me so I decided to standardize the player placement at all Survival Guides.

Biggest change: Position of Survival Guide in Romaeve was a full 36 points different on the X-axis from where the Survival Guide book actually is (most likely an accidental inverse of numbers (51 instead of 15)).

**_I affirm:_**


EpicTaru:
~~I've decided to convert this to draft and work on rotation of player when placed at the target guide.~~

Edit: DONE!!!

By default, I'll set it so they are placed in front of the guide, facing it. If navmesh issues prevent this, you'll be placed on the left or right side, facing the guide. If neither of these are able to be used, I'll place you on the guide, facing out from it (as if you were the open guide book, but you had eyes, looking out at people that are standing in front of the guide).....but don't think I'll need this last option at all. 

Any thoughts/objections?

zach2good (approved):

PR #1398, opened 2020-10-20
Adjust char timestamp migration for older Mariadb

Kreidos:
It appears some older versions of maria require a specific default value.

Thanks to Wren + Kain for reports.

**_I affirm:_**


zach2good (approved):
Noooo MySQL is meant to be the fussy one! Maria is meant to be chill :(

TeoTwawki:
technically, the timestamps were always fussy on older mysql also! Maria just got nice over time so older maria's had some of mysql's baggage for being a drop in replacement.

PR #1403, opened 2020-10-21
Add !immortal command

ibm2431:
Who wants to live forever?

Well, with this command we can be princes of the universe!

So while we're fighting to survive, we never fall to the darkest powers!

**_I affirm:_**

### DIFF
--- a/scripts/commands/immortal.lua
+++ b/scripts/commands/immortal.lua
@@ -14,7 +14,7 @@ function error(player, msg)
     player:PrintToPlayer("!immortal {player}")
 end
 
-function onTrigger(player, hp, target)
+function onTrigger(player, target)
     -- validate target
     local targ
     local cursor_target = player:getCursorTarget()
@@ -37,10 +37,10 @@ function onTrigger(player, hp, target)
             local immortal = false
             if targ:isPC() then
                 if targ:getCharVar("Immortal") == 1 then
-                    targ:delStatusEffect(tpz.effect.ERGON_MIGHT)
+                    targ:delStatusEffectSilent(0)
                     targ:setCharVar("Immortal", 0)
                 else
-                    targ:addStatusEffect(tpz.effect.ERGON_MIGHT, 0, 0, 0)
+                    targ:addStatusEffectEx(0, tpz.effect.TRANSCENDENCY, 0, 0, 0)
                     targ:setCharVar("Immortal", 1)
                     immortal = true
                 end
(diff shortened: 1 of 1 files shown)


zach2good (approved):
Cool!

TeoTwawki:
Q: does unkillable prevent doom death and other instakills? :thinking: 

unkillable style instead of massive regen and 2hjhrs is legit how retail has their GMs not die. Thank you for thinking of this, it never occurred to me that flag would work on a player entity, I've been using a stripped down godmode instead for years when I want to test things.

TeoTwawki (approved):

zach2good:
image

scripts/commands/immortal.lua
```diff
@@ -0,0 +1,69 @@
...
+        targ = player
+    end
+
+    if not targ:isNPC() then
+        if targ:isAlive() then
+      
+            local immortal = false
+            if targ:isPC() then
+                if targ:getCharVar("Immortal") == 1 then
+                    targ:delStatusEffect(tpz.effect.ERGON_MIGHT)
```
TeoTwawki: um didn't catch the use of a status effect my first look over. lets not do that. 

Lets just used addStatusEX to make its a duplicate of brew effect instead, for the lols :+1:

TeoTwawki (changes requested):

ibm2431:
> Q: does unkillable prevent doom death and other instakills?

Looking like it doesn't. Unkillable just prevents your HP from falling down to 0 in `addHP`. Doom just calls `setHP(0)` outright. So you're still technically killable.

TeoTwawki:
> 
> 
> > Q: does unkillable prevent doom death and other instakills?
> 
> Looking like it doesn't. Unkillable just prevents your HP from falling down to 0 in `addHP`. Doom just calls `setHP(0)` outright. So you're still technically killable.

K I still need doom screen and deathres if I test mobs with access to those then :( that sucks

ibm2431:
Actually! I may have lied! I didn't read `setHP` very closely!

```cpp
    ((CBattleEntity*)m_PBaseEntity)->health.hp = 0;
    auto value = (int32)lua_tointeger(L, 1);
    ((CBattleEntity*)m_PBaseEntity)->addHP(value);
```

Since `addHP` was called and the entity's HP is currently zero, unkillable bumps it back up to 1 (in `addHP`).

Just tested with `!immortal` and `!hp 0`!

TeoTwawki (approved):
re-approved

PR #1407, opened 2020-10-22
General combat calculation improvements for Avatars

NassirAlawar:
Too much changed to go into detail here. I have summarized changes in each commit and my code is packed full of comments and links to references and research.

Super short summary: 
Updated battleutils and summon.lua with equations for pDIF, fSTR, Hit Rate, Level Correction.
I have updated the base weapon computation for avatars.
I have updated how crit rate from dex works.
Level correction is not computed for newer zones/content since the release of SoA.
Some minor bug fixes.

Note: I have not updated the BP damage equations and plan to tackle those in the future, likely one avatar per PR or maybe even one BP per PR.

Note: I have not added the Damage Limit+ trait or "Physical Damage +%" to the computations as I do not believe the project supports them yet. I have noted how they were and where they need to be added once they are implemented.

**_I affirm:_**


scripts/globals/summon.lua
```diff
@@ -10,112 +10,278 @@ function getSummoningSkillOverCap(avatar)
...
+    hitrateFirst = hitrateFirst / 100
 
-    -- start the hits
-    local hitsdone = 1
-    local hitslanded = 0
-    local hitdmg = 0
+    hitrateSubsequent = utils.clamp(hitrateSubsequent, minHitRate, maxHitRate)
+    hitrateFirst = utils.clamp(hitrateFirst, minHitRate, maxHitRate)
+
+
```
zach2good: No need for double line breaks here

src/map/utils/battleutils.cpp
```diff
@@ -2229,14 +2229,84 @@ namespace battleutils
...
+            //* Owing to this change, the maximum accuracy of abilities that rely on main weapon accuracy has also been raised from 95% to 99%.
+            //Further, some monster damage types have been changed from hand-to-hand to blunt.* Fellows and alter egos enjoy this benefit as well.
+            //The maximum accuracy of beastmaster familiars, wyverns, avatars, and automatons has been increased from 95% to 99%.
+            //* In line with this change, familiars summoned using the following items have had their damage types changed from hand-to-hand to blunt.
+            //Carrot Broth / Famous Carrot Broth / Bug Broth / Quadav Bug Broth / Berbal Broth / Singing Herbal Broth / Carrion Broth / 
+            //Cold Carrion Broth / Meat Broth / Warm Meat Broth / Tree Sap / Scarlet Sap / Fish Broth / Fish Oil Broth / Seedbed Soil / Sun Water / 
+            //Grasshopper Broth / Noisy Grasshopper Broth / Mole Broth / Lively Mole Broth / Blood Broth / Clear Blood Broth / Antica Broth / Fragrant Antica Broth
+
+            int32 maxHitRate = 99;
+            auto targ_weapon = dynamic_cast<CItemWeapon*>(PAttacker->m_Weapons[SLOT_MAIN]);
```
zach2good, line 2299: This cast can still produce `nullptr`, which will get used by `targ_weapon->isTwoHanded();` later, you can change that to:
`bool isTwoHanded = targ_weapon && targ_weapon->isTwoHanded();` to make it safe

src/map/utils/battleutils.cpp
```diff
@@ -2331,99 +2443,213 @@ namespace battleutils
...
+        if(isCritical)
         {
-            cRatio += 1;
+            wRatio += 1;
         }
 
-        cRatio = std::clamp<float>(cRatio, 0, ratioCap);
+        float qRatio = wRatio;
+        float upperLimit = 0.0;
+        float lowerLimit = 0.0;
```
zach2good: Floats should be marked as `0.0f` or else they'll be created as doubles and then converted. I'm sure the compiler probably handles this, but better to be clear

src/map/utils/battleutils.cpp
```diff
@@ -2331,99 +2443,213 @@ namespace battleutils
...
+        //H2H & GK : 3.5 : 4.5
+        //2H : 3.75 : 4.75
+        //Scythe : 4 : 5
+        //Archery & Throwing : 3.25 : 3.25*1.25
+        //Marksmanship : 3.5 : 3.5*1.25
+        
+        //https://www.bluegartr.com/threads/114636-Monster-Avatar-Pet-damage
+        //Monster pDIF = Avatar pDIF = Pet pDIF
+
+        auto targ_weapon = dynamic_cast<CItemWeapon*>(PAttacker->m_Weapons[SLOT_MAIN]);
```
zach2good, line 2544: Same here, you're going to be using `targ_weapon` later and there isn't a guarantee that it'll be populated, so either bail all of those checks if this is nullptr or put a nullptr check inline with each of those down below

src/map/utils/battleutils.cpp
```diff
@@ -2331,99 +2443,213 @@ namespace battleutils
...
         {
+            //Crit Attack Bonus caps at +100% and is a flat increase to final crit damage
+            //so this is change to increase pDIF and not the qRatio
             int16 criticaldamage = PAttacker->getMod(Mod::CRIT_DMG_INCREASE) - PDefender->getMod(Mod::CRIT_DEF_BONUS);
+
+            //Avatars get the Crit. Atk. Bonus II trait for a +8% crit damage increase
+            if(attackerType == TYPE_PET)
+            {
+                CPetEntity* petEntity = dynamic_cast<CPetEntity*>(PAttacker);
+                if(petEntity)
```
zach2good: Can be changed to:
```cpp
if (CPetEntity* petEntity = dynamic_cast<CPetEntity*>(PAttacker))
{
    ...
```

src/map/utils/battleutils.cpp
```diff
@@ -2487,6 +2713,25 @@ namespace battleutils
...
+            //Since I can only confirm Avatars and this has a much larger impact on balance I will
+            //Only change this logic for Avatars pending further testing.
+
+            ENTITYTYPE attackerType = PAttacker->objtype;
+            bool isAvatar = false;
+
+            if(attackerType == TYPE_PET)
+            {
+                CPetEntity* petEntity = dynamic_cast<CPetEntity*>(PAttacker);
+                isAvatar = petEntity->getPetType() == PETTYPE_AVATAR;
```
zach2good: petEntity can be nullptr here
```cpp
if (CPetEntity* petEntity = dynamic_cast<CPetEntity*>(PAttacker))
{
    isAvatar = petEntity->getPetType() == PETTYPE_AVATAR;
}
```

src/map/utils/petutils.cpp
```diff
@@ -1608,13 +1608,17 @@ namespace petutils
...
-            ((CItemWeapon*)PPet->m_Weapons[SLOT_MAIN])->setDamage((uint16)(floor(PPet->GetMLevel() * 0.74f)));
 
-            if (PetID == PETID_CARBUNCLE)
+            //In a 2014 update SE updated Avatar base damage
+            //Based on testing this value appears to be Level now instead of Level * 0.74f
+            uint16 weaponDamage = 1 + PPet->GetMLevel();
+            if (PetID == PETID_CARBUNCLE || PetID == PETID_CAIT_SITH)
             {
-                ((CItemWeapon*)PPet->m_Weapons[SLOT_MAIN])->setDamage((uint16)(floor(PPet->GetMLevel() * 0.67f)));
+                weaponDamage = (uint16)(floor(PPet->GetMLevel() * 0.9f));
```
zach2good: Use `static_cast` instead of C-Style casts please

zach2good (commented):
I can't really speak for the maths content of the PR, this sort of balancing and stats stuff is _way_ beyond what I understand. Thanks for all the links to sources and explanations! They should help out people with a better understanding of whats going on here.

Lua all looks nice and tidy.

I have a couple of C++ dings:
General nullptr checking (because paranoia is good in core)
https://github.com/project-topaz/topaz/blob/release/CONTRIBUTING.md#breakbeforebraces-allman
https://github.com/project-topaz/topaz/blob/release/CONTRIBUTING.md#spacebeforeparens-controlstatements
https://github.com/project-topaz/topaz/blob/release/CONTRIBUTING.md#casting---static_cast-over-c-style

And maybe an excuse to introduce a new rule:
In lua you're leaving a space after starting a comment: `-- Comment`
Can you do the same with C++: `// Comment` instead of `//Comment`

Otherwise, everything looks spot-on! 👍 

EDIT: Added the comment thing as a rule
https://github.com/project-topaz/topaz/blob/release/CONTRIBUTING.md#general-code-guidlines-all-languages

TeoTwawki:
> And maybe an excuse to introduce a new rule:
In lua you're leaving a space after starting a comment: -- Comment
Can you do the same with C++: // Comment instead of //Comment

new? that was always a rule. I know coz it was my rule before topaz happened. :stuck_out_tongue:

scripts/globals/summon.lua
```diff
@@ -10,112 +10,278 @@ function getSummoningSkillOverCap(avatar)
...
-    elseif ratio < 3.6 then
-        maxRatio = 2.4 * ratio - 2.52
-        minRatio = 5 * ratio / 3 - 2
-    else
-        maxRatio = 4.2
-        minRatio = 4
+
+    local zoneId = avatar:getZone():getID()
+
+    local shouldApplyLevelCorrection = (zoneId < 256) and not (zoneId == 183)
```
ibm2431, line 130: [Should throw in the source for this](https://www.bg-wiki.com/bg/PDIF#Level_Correction_Function_.28cRatio.29), because I was gearing up to fight it. 😉 

Although we should probably be setting up and using a new zone flag instead of checking hardcoded zone IDs.

NassirAlawar:
Sure, I'll fix these today 👍 
Just to be clear on the thread name: "General combat calculation improvements for Avatars"
This will impact players as well, I started with the summon.lua but then that drifted into the general combat equations for the melee hits.

scripts/globals/summon.lua
```diff
@@ -10,112 +10,278 @@ function getSummoningSkillOverCap(avatar)
...
+            end
+            -- get a random ratio from min and max
+            local qRatio = getRandRatio(wRatio)
+
+            --Final pDif is qRatio randomized with a 1-1.05 multiplier
+            local pDif = qRatio * (1 + (math.random() * 0.05))
+            --Avatars have Crit Atk Bonus II for +8% crit dmg
+            if isCrit then
+                pDif = pDif * 1.08
+            end
```
ibm2431: Let's give/use a mod for the Crit Bonus via trait~!

scripts/globals/summon.lua
```diff
@@ -10,112 +10,278 @@ function getSummoningSkillOverCap(avatar)
...
+            finaldmg = avatarHitDmg(weaponDmg, fSTR, pDif) * dmgmod
+            numHitsProcessed = 1
+        end
+
+        -- If the first hit kills the mob other hits are not computed in retail
+        -- This can easily be seen by using BPs on low level mobs where the first hit kills them
+        while numHitsProcessed < numHitsLanded and finaldmg < targetHP do
```
ibm2431: Problem with the `finaldmg < targetHP` check is that `AvatarPhysicalMove` is called _before_ defense adjustments for the defender `AvatarFinalAdjustments`, which can cause these cycles to break out early under the assumption the mob will be dead.

scripts/globals/summon.lua
```diff
@@ -10,112 +10,278 @@ function getSummoningSkillOverCap(avatar)
...
+        end
 
-function avatarHitDmg(dmg, pdifMin, pdifMax, fstrMin, fstrMax, critrate)
-    local fstr = math.random(fstrMin, fstrMax)
-    local pdif = math.random(pdifMin * 1000, pdifMax * 1000) / 1000
-    if math.random() < critrate then
-        pdif = math.min(pdif + 1, 4.2)
+        if finaldmg > 0 then
+            target:wakeUp()
+        end
```
ibm2431: I know this was here before, but waking up targets is now baked into lua_baseentity:takeDamage(); you can remove it from the raw damage calculation step.

scripts/globals/summon.lua
genoxd: Oh interesting okay, maybe down the road we need a final display damage and an internal total damage. Good catch 👍

scripts/globals/summon.lua
genoxd: lol sure I'll add it, I left out the link because I'd been linking to pDIF so often. I thought I was safe!
I would agree with a zone flag, the first place I looked was the DB to see if there was any flag per expansion and or anything else I could use. Luckily they're all clumped together except for Legion A

scripts/globals/summon.lua
genoxd: Sure, I'll look into doing that

NassirAlawar:
I believe all the noted issues have been fixed and I have merged and retested

NassirAlawar:
Just realized I forgot to implement the request to give avatars crit att bonus from db and get rid of the hard coding. Will try to get that done in a day or two

NassirAlawar:
Changed the crit bonus to a modifier and added the modifier logic to summon.lua as well

I have retested and confirmed I see no issues. As always, let me know if there's anything else 👍

zach2good (approved):

zach2good:
We were talking about this PR yesterday; Avatars will eventually need a full rewrite, but this lays out a well-researched framework for when that happens 👍

NassirAlawar:
Let me know if there's anything I can do to help regarding avatars or if there's anything you'd prefer to have prioritized. My next plan was to move onto blood pact equations including modifiers. I have been collecting some Pet: DEX+/STR+ gear for math stuff on retail.

zach2good:
I honestly know nothing about combat balancing and rely on the others for sanity checking, but if you continue testing things and documenting your sources, we'll have all of that available when someone "fixes" avatars 👍

PR #1408, opened 2020-10-23
Shift Bastok Quest IDs

kaincenteno:
added TRUST_BASTOK. I know it has its own branch but adding it here so it doesn't get lost

**_I affirm:_**

### DIFF
--- a/scripts/globals/quests.lua
+++ b/scripts/globals/quests.lua
@@ -234,6 +234,8 @@ tpz.quest.id =
         FULLY_MENTAL_ALCHEMIST          = 88,
         SYNERGUSTIC_PURSUITS            = 89,
         THE_WONDROUS_WHATCHAMACALLIT    = 90,
+        SYNERGISTIC_SUPPORT             = 91,
+        TRUST_BASTOK                    = 92,
     },
 
     -----------------------------------


ibm2431:
Others:
Trust branch currently has Trust: Bastok as 91. kain is saying that there's a quest that _should_ have been there of Synergistic Support, which is 8 years old.

I'm hoping someone made a mistake when adding Trust Bastok.

If not, this could be quite bad for us.

TeoTwawki:
@ibm2431[ its part of a "deleted" quest-line on retail.](https://ffxiclopedia.fandom.com/wiki/Synergistic_Pursuits) (removed on 3-27-12) its still in the dats but no longer can be flagged. Still shows in my "completed" list though:

image

Polutils shown to confirm it does come right before the trust quest in the index, though our IDs do not match that.

The Japanese shown just says "BS Quest 93           Requester" - placeholder entries for future things. I like to think some SE devs was like _"sigh.. Bull Sh--- quest 98..  Bull Sh-- Quest number 99.."_

TeoTwawki:
Support was a repeatable btw, The 2 that came right before are not.

kaincenteno:
dunno if this got changed recently or if the quest was always flagged wrong for bastok. Would be good if someone had an old client. for like last year maybe?

kaincenteno:
@TeoTwawki  @ibm2431 

No issue with progression. After further investigation it seems the issue is being cause by another issue will be open up.

This only fixes the missing questID.

@zach2good  tagging you so you merge this onto trust branch

ibm2431:
@TeoTwawki 

> Trial-size Trial by Earth
> Trial by Earth

> All By Myself

> The Naming Game

This just keeps looking worse. I was really hoping it was just Jeuno.

kaincenteno:
Changed name of this PR as it only fixes the missing quests.

Issue with trusts quest progression in bastok can be explained in https://github.com/project-topaz/topaz/issues/1411

zach2good:
If this just makes the quest in Bastok appear correctly in the logs, then this is fine to merge.
People _might_ get caught in a position where a GM has to use GM wizardry to fix their quest status, but it's so small on such minor quests that I think that's OK.
```
        SYNERGISTIC_SUPPORT             = 91,
        TRUST_BASTOK                    = 92,
```

This has unearthed a lof of evil that we should track outside of this PR.

PR #1412, opened 2020-10-23
Byakko drops up to 2 hides, but only 15% each (common)

jarmengaud:
Byakko drops 2 hides but only 15% (common) each

https://ffxiclopedia.fandom.com/wiki/Byakko

**_I affirm:_**


TeoTwawki:
At least one of these is likely group with another item or is set to fall in an empty slot after other drops if enough room is left. other wise we could exceed max loot pool size making the rows that come after the hides accidentally rarer than intended.

Consider this comment a future todo/issue to be solved with some screenshots of loot pools to figure out the order/slots the items come in.

zach2good (approved):

PR #539, opened 2020-04-27
Mazurka Fix

Omnione:
• Added the ability to cast mazurka in appropriate zones.
• Added effect flags to wear on action against mob and wear on mob hit.

This fixes issue #528

BRD Mazurka should be allowed in dungeons (and removal on hit or attack)
issued by @kaincenteno

**_I affirm:_**
- [X] that I agree to Project Topaz's [Limited Contributor License Agreement](http://project-topaz.com/blob/release/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [X] that I've _tested my code_ since the last commit in the PR, and will test after any later commits

PR #1413, opened 2020-10-23
Adjust Trust and Mount flags in Adoulin and ToAU zones

jarmengaud:
Missing mount flag in some Adoulin open fields
Removed mount flag in Adoulin dungeons
Adding trust flag in ToAU event zones

**_I affirm:_**


kaincenteno:
ouch wasn't aware that people were able to use mounts in dungeons lol. Glad this fixed it

ibm2431:
May wanna go over these again; SOA dungeons are losing Escape and Fellow.

Mazurka might be okay; [Omni deliberately didn't touch SOA zones when he was adjusting Mazurka flags last](https://github.com/project-topaz/topaz/pull/539).

flags

jarmengaud:
Escape is intended, release branch does not have it either, because it takes more work to allow escape (need to define escape zones).
I will add back fellow

jarmengaud:
did a quick test on retail, you can use mazurka in SOA dungeons

jarmengaud:
Ok flag FELLOW added for dungeons, even though this flag is not consistent across all zones, which has never been detected as an issue since... fellow is not implemented :-)

PR #1414, opened 2020-10-23
Harvest festiva l2005

bope12:
**_I affirm:_**

Basic 2005 Harvest Festival event, all feedback is greatly welcome as I am going to be adding events as they come up in the year with content up to time permitting.
There is most likely easy ways to clean this up even more but I wanted to get this pr up before the start of the event on retail.
There are npc's that I did did not have captures of what their costumed look was so I made up based on the zones I did have they are marked with Missing in the zones ID file, I went with only 4 patrol paths one for each zone and two in waters as I did not know their exact paths but wanted them to at least wonder around some. Currently if you talk to a npc it will freeze them.
I included a new binding to change npc's look.size value as that was needed to change their look without requiring a sql change and reboot.
atPoint() was using float values and failed the check so I clamped the check to three decimal points allowing for paths not limited to whole digits.
pathfinds patrol method looks like it was meant to start from any point on the path so I cleaned it up to make that work


src/map/entities/baseentity.cpp
```diff
@@ -153,3 +153,13 @@ uint16 CBaseEntity::GetModelId()
 {
     return look.modelid;
 }
+
+void CBaseEntity::SetLookSize(uint16 size)
+{
+    look.size = size;
+}
+
+uint16 CBaseEntity::GetLookSize()
```
TeoTwawki, line 162: That byte in the look value on these NPCs is not really size and we should stop calling it that.

src/map/lua/lua_baseentity.cpp
```diff
@@ -4622,7 +4623,32 @@ inline int32 CLuaBaseEntity::checkNameFlags(lua_State* L)
...
+    lua_pushinteger(L, m_PBaseEntity->GetLookSize());
+
+    return 1;
+}
 
+inline int32 CLuaBaseEntity::setLookSize(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+
+    if (m_PBaseEntity->objtype == TYPE_PC)
```
TeoTwawki, line 4639: check not check for != TYPE_PC instead, and then avoid having the empty conditional?

scripts/globals/events/harvest_festivals.lua
```diff
@@ -5,6 +5,231 @@ require("scripts/globals/settings")
...
+        [234] = { 17735939, 17735940, 17735941, 17735962, 17735961 },
+        [235] = { 17740139, 17740140, 17740141, 17740142, 17740143 },
+        [238] = { 17752504, 17752505, 17752506, 17752507, 17752548, 17752545 },
+        [241] = { 17764742, 17764743, 17764744},
+    },
+    paths =
+    {
+        bastok =
+        {
+        -341.947, -10.004, -161.233,
```
TeoTwawki: indents off

scripts/globals/events/harvest_festivals.lua
```diff
@@ -5,6 +5,231 @@ require("scripts/globals/settings")
...
+        goblin = 8,
+    },
+    decorations =
+    {
+        [230] = { 17719828, 17719829, 17719830},
+        [231] = { 17723774, 17723775, 17723776, 17723777, 17723812, 17723811 },
+        [234] = { 17735939, 17735940, 17735941, 17735962, 17735961 },
+        [235] = { 17740139, 17740140, 17740141, 17740142, 17740143 },
+        [238] = { 17752504, 17752505, 17752506, 17752507, 17752548, 17752545 },
+        [241] = { 17764742, 17764743, 17764744},
```
TeoTwawki: brace spacing inconsistent. some lines got spaces, some didn't.
@zach2good whats our convention now for spacing?

scripts/globals/events/harvest_festivals.lua
```diff
@@ -5,6 +5,231 @@ require("scripts/globals/settings")
...
+        applyHalloweenNpcCostumes(zone:getID(),true)
+        applyHalloweenDecorations(zone:getID(),true)
+    end
+    events.harvest.registerRegions(zone)
+end
+
+function events.harvest.registerRegions(zone)
+    switch (zone:getID()): caseof
+    {
+        [230] = zone:registerRegion( 7,  122.602,  0.0,   72.500,  125.000,  0.0,   75.625 ), --Southern_San_dOria Pitchfork
```
TeoTwawki: inconsistency here trips my OCD. functions contents isn't indented, the comments aren't up to standard `-- space` instead of `--nospace` and while it may not technically be a problem having some be function contents and others not really throws me off reading it.

I'll let someone else weigh in because unfortunately my er..condition..causes me to zero in on formatting and miss the bigger picture, which isn't fair to you.

bope12: It's all fair I'll fix up all that you mentioned when I get home tonight

PR #1415, opened 2020-10-23
Fix mobskill flag 0x010 (SKILLFLAG_REPLACE_ATTACK)

### LINKED ISSUE #1406: SetMobSkillAttack & tpz.mobMod.ATTACK_SKILL_LIST use TP
**_I have:_**

I observed and discussed this issue in https://github.com/project-topaz/topaz/pull/1388#issuecomment-712293072 and even made a [debug video](https://streamable.com/ojrve0) of this behavior.
`SetMobSkillAttack(list_id)` and `setMobMod(tpz.mobMod.ATTACK_SKILL_LIST, list_id)` are supposed to replace the default attack of a mob with the skill list ID passed to them and are currently mainly used for flying wyrms during their flight phase. The latter is actually not used at all but it serves the same purpose, and has the same flaw.

It can be observed however that **both of these** actually **use up all TP of the mob** on every regular hit as if it still was a mobskill instead of a regular attack. This is extremely bad, as the mob now does not generate any TP at all and does not use any mobskills.

This affects all flying wyrms like Jormungand, Tiamat, Ouryu and Wyrm and makes their flying phase extremely boring and well,...simply bugged.

MarianArlt:
**_I affirm:_**

Adds `isAttackReplacement()` to `mobskill.cpp` to prevent mobskill flag `SKILLFLAG_REPLACE_ATTACK = 0x010` from using up TP.

Fixes https://github.com/project-topaz/topaz/issues/1406
- [x] Successfully built and tested working locally

### DIFF
--- a/src/map/mobskill.cpp
+++ b/src/map/mobskill.cpp
@@ -69,9 +69,14 @@ bool CMobSkill::isTwoHour() const
   return m_Flag & SKILLFLAG_TWO_HOUR;
 }
 
+bool CMobSkill::isAttackReplacement() const
+{
+  return m_Flag & SKILLFLAG_REPLACE_ATTACK;
+}
+
 bool CMobSkill::isTpSkill() const
 {
-    return !isSpecial();
+    return !isSpecial() && !isAttackReplacement();
 }
 
 bool CMobSkill::isSpecial() const
--- a/src/map/mobskill.h
+++ b/src/map/mobskill.h
@@ -50,6 +50,7 @@ public:
     bool        isSingle() const;
     bool        isTwoHour() const;
     bool        isSpecial() const;
+    bool        isAttackReplacement() const;
     bool        isTpSkill() const;
 
     uint16      getID() const;


src/map/mobskill.cpp
```diff
@@ -69,9 +69,14 @@ bool CMobSkill::isTwoHour() const
   return m_Flag & SKILLFLAG_TWO_HOUR;
 }
 
+bool CMobSkill::isAttackReplacement() const
+{
+  return m_Flag & SKILLFLAG_REPLACE_ATTACK;
```
zach2good, line 74: This is probably from copy/pasting my comments ;)
4 spaces here please

zach2good:
Putting this into a branch so I can make that change before I go to bed, thanks for this 👍

src/map/mobskill.cpp
MarianArlt: Yeah it was, sorry ^^
Thanks for correcting. There's a lot of inconsistency in that file regarding indentation.

PR #1417, opened 2020-10-23
Validate wide scan tracking

InoUno:
Adds validation to wide scan tracking, such that only characters with an appropriate range can track entities, and only entities that show up on wide scan can be tracked.

**_I affirm:_**


ibm2431 (approved):
Seems sane. Thanks, Ino~!

TeoTwawki (approved):

zach2good (approved):

PR #1422, opened 2020-10-24
Yagudo NMs in dynamis remove doom on death

rude-jerk:
**_I affirm:_**


TeoTwawki:
While the dyna yagudo may do this at a 100% rate, there has also always been a random chance for this to occur for other mobs doom's as well that has never been implemented in the project. It may be worth making the this a little less yag specific and having a way to define the chance. There is also a message "PlayerName narrowly escapes impending doom." (its msg basic id 359)

kaincenteno:
Awesome work @rude-jerk  :smile: 

Something from FFXIcyclopedia that they should also receive a message when the status gets deleted:
```(when this occurs, "[Player] narrowly escapes impending doom" will be displayed in the chat log).```

Dunno if you have played with data inside ID.lua inside each zone. If that would be a textid that is the same on each dynamis zone that can be prompted. If it is then you would have to store the textID inside each ID.lua for each zone that the yagudo dynmis are =)

TeoTwawki:
@kaincenteno its a messageBasic not a messageSpecial, so it'd go in the msg.lua global instead. :)

rude-jerk:
Whoops! Missed the message because it was in a different color than I was expecting in my reference video.

kaincenteno:
Awesome @TeoTwawki  that makes it even easier =)

But yeah @rude-jerk  i agree that you would have to keep track on WHICH yagudo uses it so the correct yagudo that casted it when killed can remove the effect of Doom (which is a technicality but would make it more accurate to retail =) )

But so far your work looks amazing =) thank you for spending some time on making our dear yagudo army great again =)

rude-jerk:
Generalized the mixin and allowed for removal chance to be set per-mob using local vars. Tracking the mob that applied the debuff can probably be added as a follow-up for mobs where that's of a concern.

wrenffxi (changes requested):
I tested this PR via the following steps:

1. Enter Dynamis-Beaucedine and `!mobhere 17326289` to bring [Xaa Chau](https://ffxiclopedia.fandom.com/wiki/Xaa_Chau_the_Roctalon) to me.
2. Engage mob.  This mob is supposed to use Doom, but doesn't, so I `!exec target:useMobAbility(1073)` to make it doom me.
3. While doomed, `!hp 0` to kill the mob.  This successfully cleared doom status from my character, with message `Wren narrowly escapes impending doom.`

This all looks good to me.

~~However, I believe this line `player:messageBasic(tpz.msg.basic.NARROWLY_ESCAPE)` should be `player:messagePublic(tpz.msg.basic.NARROWLY_ESCAPE, player)` so that the message appears to nearby players as well.~~ *edit: OP fixed with follow-up commit.  Approved.*

wrenffxi (approved):

PR #1424, opened 2020-10-24
Ranged Attacks for Trusts & Trust: Semih Lafihna & Ayame WS Logic

zach2good:
**_I affirm:_**

**TODO**
- Ranged WS animations for Semih are messed up.

```
        [3487] = {['name']="Sidewinder",                     ['category']=11, ['id']=3487, ['animation']=429,  ['message']=185, },
        [3489] = {['name']="Stellar Arrow",                  ['category']=11, ['id']=3489, ['animation']=430,  ['message']=185, },
        [3490] = {['name']="Lux Arrow",                      ['category']=11, ['id']=3490, ['animation']=431,  ['message']=185, },
```

- Actionview shows the category, id, animation and message are all same between retail and my changes...
- Model is correct, capped from retail...

**DONE**
- Basic RA for trusts
- AI for Semih
- Trusts stop running are 3 ticks of being chased 💀
- I think I'm going to have to move all the RA stuff down into BattleEntity, it just makes sense. Even if it's stubbed.
- Sane damage and delay for RA
- Throwing support for Joachim
- Bindings to tell Trusts to use RA
```lua
    mob:addSimpleGambit(ai.t.TARGET, ai.c.ALWAYS, 0,
                        ai.r.RA, 0, 0) -- Try RA all the time, but bound by your weapon's delay
```


ibm2431:
> I think I'm going to have to move all the RA stuff down into BattleEntity, it just makes sense.

I'm surprised it already isn't. Do Goblin Ambushers currently not Ranged Attack?

TeoTwawki:
There are 2 types of ranged attack mob type objects can use. one is a "real" ranged attack, the other is a mobskill faking a ranged attack. A case of SE having multiple devs who didn't communicate to each other how they implemented. In game on retail, you can tell them apart by how the message displays: 1 line of text with no line break, or 2.

_both_ types should be handled at battelentity imo.

PR #1426, opened 2020-10-24
Enable (and fix) basic warnings on all platforms

zach2good:
**_I affirm:_**

Clang:
```
-Wlogical-not-parentheses
Example: !PMob->getMobMod(MOBMOD_NO_DESPAWN) != 0

-Wenum-compare-switch
-Wabsolute-value
Using abs() on an unsigned value (negative isn't possible)

-Wtautological-constant-out-of-range-compare
Checking to see if a condition is true on a type where it isn't possible.  Example: is a uint8 larger than 10000?

-Wnon-virtual-dtor
-Wunused-lambda-capture
-Wmissing-field-initializers
-Wunused-private-field
-Wmissing-field-initializers
```


src/common/utils.cpp
```diff
@@ -155,7 +155,7 @@ uint8 relativeAngle(uint8 world, int16 diff)
     uint8 angle = world + diff;
     if (angle < 0)
     {
-        angle = 256 - abs(angle);
+        angle = 256 - angle;
```
ibm2431, line 158: I may have made a mistake and the uint8 angle above needs to be int16.

zach2good:
Closing for now, don't have time to do this audit :(

PR #1427, opened 2020-10-24
Update table of BCNMs in bcnm global

wrenffxi:
**_I affirm:_**

* Add new battlefields to table in bcnm.lua (discovered via `!cs 32000 0 0 0 2147483647`)
* Use tpz.zone keys rather than zone numbers
* Braces on new lines


ibm2431 (approved):
> Use tpz.zone keys rather than zone numbers

❤️

TeoTwawki (approved):

PR #1430, opened 2020-10-25
Correct wyvern breath element selection

rude-jerk:
**_I affirm:_**

https://www.bg-wiki.com/bg/Wyvern_(Dragoon_Pet)

> 
> The selected breath is determined exclusively by an opponent's elemental resistances at the time of activation
> 
> Magic damage reduction, elemental damage reduction and absorption, elemental magic accuracy buffs such as Klimaform, or equipment such as the Drachen Armet/+1 or Vishap Armet/+1/+2/+3 have no bearing on a wyvern's choice of breath.
> 
> Effects that lower enemy resistances such as Threnody or Skillchains are factored into this choice.
> 
> In the event of a tie for lowest resistance, the element is selected in the following order of Fire → Ice → Wind → Earth → Thunder → Water.


scripts/globals/pets/wyvern.lua
```diff
@@ -116,13 +116,25 @@ function onMobSpawn(mob)
...
+                local lowest = resistances[tpz.jobAbility.FLAME_BREATH]
+                for key, value in pairs(resistances) do
+                    if value < lowest then
+                        lowest = value
+                    end
+                end
+                for key, value in pairs(resistances) do
+                    if value == lowest then
+                        table.insert(breaths, key)
+                        break
```
ibm2431: You can cut the second loop by assigning your `breath` as you're stepping through the table the first time. This will be guaranteed to only be a single breath when you exit the loop because your resistance comparisons are strictly less than and don't include equal to. Then you can `table.insert(breaths, breath)` directly.

rude-jerk: In making this change I noticed that I needed to convert to using ipairs, as the table order was arbitrary using pairs. Corrected this at the same time!

scripts/globals/pets/wyvern.lua
```diff
@@ -116,15 +116,32 @@ function onMobSpawn(mob)
...
+                local breathList =
+                {
+                    tpz.jobAbility.FLAME_BREATH,
+                    tpz.jobAbility.FROST_BREATH,
+                    tpz.jobAbility.GUST_BREATH,
+                    tpz.jobAbility.SAND_BREATH,
+                    tpz.jobAbility.LIGHTNING_BREATH,
+                    tpz.jobAbility.HYDRO_BREATH,
+                }
+                local resistances = {
```
ibm2431: Gotta a non-newline brace here~

ibm2431 (approved):
Others: It's important that resistances are examined in elemental order - this is the same selection that skillchain element happens in.

ibm2431:
When elements are fixed in release we might be able to remove the two tables altogether and just use Flame Breath and Fire element as offsets.
```lua
local lowest = 0
local lowest_res = target:getMod(tpz.mod.FIRERES)
for i = 1, i <= 5, 1 do
    local res = target:getMod(tpz.mod.FIRERES + i)
    if res < lowest_res then
        lowest = i
        lowest_res = res
    end
end
table.insert(breaths, tpz.jobAbility.FLAME_BREATH + lowest)
```

But for now the tables are needed because the elements in release follow the wrong order.

PR #1432, opened 2020-10-25
Tracy Profiling

zach2good:
This is the power of CMake!

Build with `TRACY_ENABLE` to turn on the profiling, otherwise, everything will get stripped away and has no effect on the build.

This PR adds support for the Tracy client in topaz_game. Once the game is built and running, you can connect the server to it to see the beautiful stats in realtime:

image

We are hardcoded to v0.7.3, different versions of client and server are not compatible!

https://github.com/wolfpld/tracy/releases/tag/v0.7.3

**GUIDE**

https://github.com/project-topaz/topaz/wiki/Performance-Profiling-with-Tracy

**_I affirm:_**

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -255,6 +255,10 @@ void CCharEntity::clearPacketList()
 
 void CCharEntity::pushPacket(CBasicPacket* packet)
 {
+    TracyZoneScoped;
+    TracyZoneIString(GetName());
+    TracyZoneHex16(packet->id());
+
     std::lock_guard<std::mutex> lk(m_PacketListMutex);
     PacketList.push_back(packet);
 }
@@ -516,6 +520,7 @@ void CCharEntity::ClearTrusts()
 
 void CCharEntity::Tick(time_point tick)
 {
+    TracyZoneScoped;
     CBattleEntity::Tick(tick);
     if (m_DeathTimestamp > 0 && tick >= m_deathSyncTime)
     {
@@ -532,6 +537,7 @@ void CCharEntity::Tick(time_point tick)
 
 void CCharEntity::PostTick()
 {
+    TracyZoneScoped;
     CBattleEntity::PostTick();
     if (m_EquipSwap)
     {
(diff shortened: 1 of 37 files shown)


src/map/entities/charentity.cpp
```diff
@@ -255,6 +255,12 @@ void CCharEntity::clearPacketList()
 
 void CCharEntity::pushPacket(CBasicPacket* packet)
 {
+    ZoneScoped;
+    const char* cstr = reinterpret_cast<const char*>(GetName());
+    ZoneText(cstr, std::strlen(cstr));
+    auto packet_id = fmt::format("{:2x}", packet->id());
+    ZoneText(packet_id.c_str(), packet_id.size());
```
zach2good: Note to self:
```cpp
std::string Hex8ToString(uint8 hex)
{
    return fmt::format("0x{:2X}", hex);
}

std::string Hex16ToString(uint16 hex)
{
    return fmt::format("0x{:4X}", hex);
}
```

zach2good:
```cpp
#define ZoneString(str) ZoneText(str.c_str(), str.size())
#define ZoneCString(cstr) ZoneText(cstr, std::strlen(cstr))
#define ZoneIString(istr) ZoneCString(reinterpret_cast<const char*>(istr))
```

zach2good:
Note to self:
```
To report data, use the TracyPlot(name, value) macro.
To configure how plot values are presented by the profiler, you may use the TracyPlotConfig(name,
format) macro, where format is one of the following options:
• tracy::PlotFormatType::Number – values will be displayed as plain numbers.
• tracy::PlotFormatType::Memory – treats the values as memory sizes. Will display kilobytes,
megabytes, etc
```

```cpp
//what:
//LUA_GCCOUNT: returns the current amount of memory (in Kbytes) in use by Lua.
//LUA_GCCOUNTB: returns the remainder of dividing the current amount of bytes of memory in use by Lua by 1024.
//LUA_API int (lua_gc) (lua_State *L, int what, int data);

#ifdef TRACY_ENABLE
TracyPlotConfig("Lua Memory Usage", tracy::PlotFormatType::Memory);
TracyPlot("Lua Memory Usage", lua_gc(L, LUA_GCCOUNT, 0) / 1024); // from kb to b
#endif
```

ibm2431 (approved):
Let us embark on a new era of performance testing!

PR #1437, opened 2020-10-27
Fix off by one error in item_mods.sql

TeoTwawki:
**_I affirm:_**
- [x] that I've _tested my code_ since the last commit in the PR, and will test after any later commits. No really. I swear.

_/em takes the walk of shame over to the PR button_


kaincenteno:
what was the error? o.o
#mysterious

zach2good:
I was wondering why this didn't grant the skill 😆

PR #1438, opened 2020-10-27
Hook up BCNM Timer Packet

zach2good:
**_I affirm:_**

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -44,6 +44,7 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "status_effect_container.h"
 #include "treasure_pool.h"
 
+#include "utils/charutils.h"
 #include "utils/itemutils.h"
 #include "utils/zoneutils.h"
 #include "zone.h"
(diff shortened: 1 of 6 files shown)


scripts/globals/bcnm.lua
```diff
@@ -973,7 +973,7 @@ function EventTriggerBCNM(player, npc)
     elseif not player:hasStatusEffect(tpz.effect.BATTLEFIELD) then
         local mask = findBattlefields(player, npc, 0)
 
-        -- mask = 268435455 -- uncomment to open menu with all possible battlefields
+        mask = 268435455 -- uncomment to open menu with all possible battlefields
```
rude-jerk: Whoops!

zach2good: image

src/map/battlefield.cpp
```diff
@@ -288,6 +289,7 @@ bool CBattlefield::InsertEntity(CBaseEntity* PEntity, bool enter, BATTLEFIELDMOB
                 ApplyLevelRestrictions(PChar);
                 m_EnteredPlayers.emplace(PEntity->id);
                 luautils::OnBattlefieldEnter(PChar, this);
+                charutils::sendTimerPacket(PChar, m_TimeLimit);
```
zach2good: This should be SendTimerPacket

zach2good:
This has existed (and been used in Full Speed Ahead) since the summer, probably fine 🤷

PR #1444, opened 2020-10-28
Passing party leader dismisses Trusts

zach2good:
**_I affirm:_**

EDIT: Tested, works as expected


Kreidos (approved):
Really?? I'd no idea. :shrug:

TeoTwawki:
huh. didn't know you _could_ pass lead with them out.

zach2good:
Will need to check in retail

zach2good:
Tested on retail, this works fine. Trust disappear without an animation, but that's fixed in this branch: https://github.com/project-topaz/topaz/pull/1424

PR #1454, opened 2020-10-29
 Add Plumbing and Scripts to Support Fools/Fanatics type Items

neuromancerxi:
Modifies/Reorders magic_shield and physical_shield effects and adds new tiers.
Moves Rampart IronWill check to rampart effect.
Modifies the MAGIC_SHIELD check to check which power tier is active before flooring magic.
Adds scripts for Fanatics/Fools Drink/Tonics.
Adds scripts for Carnal/Spiritual Incense.
Adjusts timers to 1s where applicable.
Adjusts scripts to new values where applicable.

**SPECIAL NOTE TO SERVER OPERATORS**: Make sure you check your custom scripts when merging this change. This adjusts the scripts that existed in Topaz at the time of submission, but due to the change in how power is now applied, please make sure any scripts that use apply (and possibly remove) MAGIC_SHIELD and/or PHYSICAL_SHIELD effects use the new values in the effects.

**_I affirm:_**


scripts/globals/effects/magic_shield.lua
```diff
@@ -1,13 +1,23 @@
 -----------------------------------
+-- Magic Shield
+-- Blocks magic damage and effects depending on power
 --
--- Magic Shield BLOCKS all magic attacks
---
+-- Power Notes:
+--  0 - 50%  DMGMAGIC (e.g. Fool's Tonic)
```
TeoTwawki, line 6: Just tossing you some info/possibilities here:

We do have additional field sin the effect we could be using, which could make this simpler.

`hasStatusEffect(effectID, subid)` is also a thing. one of the lines you replaced was using that to tell apart 2 versions of magic shield. We could also use subpower and tiers. Frankly what we're doing with power even before your PR should have been tiers but way back when somebody used power so we stuck with it. We could use subpower to decide if we're blocking status effects or not as an example instead of using an invalid dmg reduction of -101.

scripts/globals/magic.lua
```diff
@@ -347,8 +347,15 @@ end
 
 function getMagicHitRate(caster, target, skillType, element, percentBonus, bonusAcc)
     -- resist everything if magic shield is active
-    if (target:hasStatusEffect(tpz.effect.MAGIC_SHIELD, 0)) then
```
TeoTwawki, line 350: We have 2 versions of this overloaded function:
 - https://github.com/project-topaz/topaz/blob/45c97efb1fdb128570a223b3bea8cb40ed281182/src/map/status_effect_container.cpp#L1110
```cpp
bool CStatusEffectContainer::HasStatusEffect(EFFECT StatusID, uint16 SubID)
{
    for (CStatusEffect* PStatusEffect : m_StatusEffectSet)
    {
        if (PStatusEffect->GetStatusID() == StatusID &&
            PStatusEffect->GetSubID() == SubID &&
            !PStatusEffect->deleted)
        {
            return true;
        }
    }
    return false;
}
```

- https://github.com/project-topaz/topaz/blob/45c97efb1fdb128570a223b3bea8cb40ed281182/src/map/status_effect_container.cpp#L835
```cpp
bool CStatusEffectContainer::HasStatusEffect(EFFECT StatusID)
{
    for (CStatusEffect* PStatusEffect : m_StatusEffectSet)
    {
        if (PStatusEffect->GetStatusID() == StatusID &&
            !PStatusEffect->deleted)
        {
            return true;
        }
    }
    return false;
}
```

I believe the original intent was to have this only apply to versions that did not have a subid, and use subid to tell apart differing versions. An unfinished implementation looked like to me (or they mistook subid for power?)

neuromancerxi: Ah, that makes much more sense now, I think I'm following your line of reasoning. I'll take a look at applying a "isfakemagicshield" sub for Fanatic's Tonic and Spiritual Incense since these offer the MDT UMDT adjustments, but not the actual magic immunity.

PR #1457, opened 2020-10-30
Add flag to disable timed RoE records + minor fix

Kreidos:
This patch is good to go into release and should be the last of the straight-to-release fixes. Future PRs will have more recordsets and features which may warrant their own branch once again. :)

Some have mentioned they don't mind RoE, but don't want or don't like the CHALLENGE SPLASH SPAM. These people are crazy, but regardless here's a settings flag to ~~turn just those off~~ turn all timed records off.

**_I affirm:_**

### DIFF
--- a/scripts/globals/roe_records.lua
+++ b/scripts/globals/roe_records.lua
@@ -17,7 +17,9 @@ local timedSchedule = {
     {     0,  4009,  4015,  4011,  4017,  4014}, -- Saturday
 }
 -- Load timetable for timed records
-RoeParseTimed(timedSchedule)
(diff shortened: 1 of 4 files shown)


ibm2431:
Do they not want the records _period_, or just the blaring trumpets?

Kreidos:
I assume the records themselves too. I didn't think it unreasonable that it be disabled, but I can remove the flag; someone can just edit the timetable to all 0's if they want them off.

ibm2431:
Just wanting to make sure the setting does what people _expect_ it to do~!

zach2good (approved):

PR #1463, opened 2020-10-31
Resolve BLU spell setting exploits

ibm2431:
Most public servers should already have this fix. If you're a public server operator who didn't know this PR was coming, get in touch with either zach2good or myself on Discord.

As this is an exploit fix, I'll wait a while before going into further details.

If you're a server operator who doesn't have this patched already, don't sleep on this. I'll eventually demonstrate how a BLU player on any DSP derived server can do this:
cheat

**_I affirm:_**


zircon-tpl:
It has been two weeks since this resolution was merged into `release`.

The following video was previously shared with me by ibm2431 to demonstrate how this exploit functioned:
https://youtu.be/RskQy8i-LPY

PR #1464, opened 2020-11-01
Change Nidhoggs method of using super_buff

MarianArlt:
**_I affirm:_**

This PR is tightly tied to https://github.com/project-topaz/topaz/pull/1388 and came to be because that PR was found to be the legitimate use case for skill ID 956
(KS99 Wyrm airborne Hurricane Wing).

In the discussion of that PR it was decided to remove the use of any non-accurate IDs for now and instead encourage the re-factoring of the random defense/resistance buff that Nidhogg can use at will.

I commented out rather than deleting anything for later reference.


Era-Lusiphur:
I haven't looked into it really but was it used that way to make it true AoE, hitting everyone nearby?

ibm2431:
Oh, Nidhogg was already set up in a local var 2hr format. That makes this a lot easier than I thought.

If proper animation / skill can't be found, then we can leave all of onMobFight as is (active), but replace the `useMobAbility` call with a direct invocation of `mob:addStatusEffectEx(tpz.effect.SUPER_BUFF, 0, 50, 0, 30)`. Then [the "super buff" skill can be deleted altogether](https://github.com/project-topaz/topaz/blob/release/scripts/globals/mobskills/super_buff.lua).

If we don't have a capture, we can go without the animation if it means fixing this wonky skill problem. 

I'd be willing to bet that Nidhogg uses VULTURE2 on retail. 😉

MarianArlt:
I searched tons of functions today to see if there's a method to simply play the animationID 432 without having to call a skill but couldn't find any. Played around with inject functions (all player or NPC targeted) and did some extremely funny things with `independantAnimation()` but none of these can do this. Is there no such function?
(Would be perfectly appropriate here)

ibm2431:
From @59blargedy's capture, which I won't link here on GitHub. It's on the Captures Discord. You can see the effect go off at 2:36 in her video. (Also note the animation ID.)
super_buff

ActionView didn't display anything because its check [against Windower's resource table failed](https://github.com/Windower/Resources/blob/master/resources_data/monster_abilities.lua#L794-L795).

scripts/globals/mobskills/super_buff.lua
```diff
@@ -0,0 +1,19 @@
+---------------------------------------------
+-- Spike Flail
+-- Deals extreme damage in a threefold attack to targets behind the user.
```
ibm2431: Be sure to change the copy-paste comment block 😉

scripts/globals/mobskills/super_buff.lua
```diff
@@ -0,0 +1,19 @@
+---------------------------------------------
+-- Spike Flail
```
TeoTwawki: fix the comment for us if you would :+1:

MarianArlt:
Awesome. Thank you ibm I have tried to spot that in the videos, looking at that panel but couldn't see it. I appreciate the support.
PR now simply adds the appropriate ID and changes the ID in the script.

scripts/globals/mobskills/super_buff.lua
```diff
@@ -0,0 +1,19 @@
...
+-- Deals extreme damage in a threefold attack to targets behind the user.
+---------------------------------------------
+require("scripts/globals/monstertpmoves")
+require("scripts/globals/settings")
```
ibm2431: I'm not sure why this was in the original Super Buff, but settings aren't used in this one!

scripts/globals/mobskills/super_buff.lua
MarianArlt: Will do.

scripts/globals/mobskills/super_buff.lua
ibm2431: Teo beat me to my comment again. ☹️

TeoTwawki: jinx

ibm2431 (approved):

TeoTwawki (approved):

PR #1465, opened 2020-11-01
Implements the Retrace spell and related usable items

Aeshur:
**_I affirm:_**

Implements the retrace spell into the game, along with related usable items:
- Scroll of Instant Retrace
- Cobra staff
- Fourth staff
- Ram staff

The staves have separate teleport IDs from Retrace because you can use their teleports after joining a different allied nation.

Retrace was changed from an 8 second cast to a 5 second cast recently according to wiki so I adjusted that.


Nireya:
Apologize for the multiple commits, I have no idea how to squash I'm too noob, feel free to advise 😹

scripts/globals/spells/retrace.lua
```diff
@@ -0,0 +1,28 @@
...
+        result = 0
+    else
+        result = 48
+    end
+    return result
+end
+
+function onSpellCast(caster, target, spell)
+    if (target:getCampaignAllegiance() > 0) then
+        target:addStatusEffectEx(tpz.effect.TELEPORT, 0, tpz.teleport.id.RETRACE, 0, 3.4)
```
TeoTwawki: 3.4 isn't actually valid because the function doesn't support floating point values. it will become just 3 at that point. I am aware we have some existing code making the same mistake that has not been corrected yet.

scripts/globals/items/scroll_of_instant_retrace.lua
```diff
@@ -0,0 +1,24 @@
...
+        result = 0
+    else
+        result = 56
+    end
+    return result
+end
+
+function onItemUse(target)
+    if (target:getCampaignAllegiance() > 0) then
+        target:addStatusEffectEx(tpz.effect.TELEPORT, 0, tpz.teleport.id.RETRACE, 0, 1)
```
TeoTwawki: 1 second? seems kinda short for that animation. But if you already checked and its retail accurate leave it.

scripts/globals/teleports.lua
```diff
@@ -271,6 +278,21 @@ tpz.teleport.toHomeNation = function(player)
...
+-----------------------------------
+
+tpz.teleport.toAlliedNation = function(player)
+    local Allegiance = player:getCampaignAllegiance()
+    if Allegiance == tpz.alliedNation.SANDORIA then
+        player:setPos(-98, 1, -41, 224, 80)
+    elseif Allegiance == tpz.alliedNation.BASTOK then
+        player:setPos(-291, -10 , -107, 212, 87)
+    elseif Allegiance == tpz.alliedNation.WINDURST then
+        player:setPos(-31.442, -5 , 129.202, 128, 94)
```
ibm2431: I know the rest of this file sets a bad example, but does anyone have thoughts on this using the tables Nireya is defining above for the teleport destinations, rather than defining the same positions twice?

Nireya: `unpack(<table>)` is a lua method which takes a table and breaks it apart into multiple arguments. The order of both needs to line up - so it's not used too often - but it could theoretically be used in this case.

scripts/globals/spells/retrace.lua
```diff
@@ -0,0 +1,28 @@
...
+-- Transports player to their Allied Nation. Can cast on allies.
+-----------------------------------------
+require("scripts/globals/teleports")
+require("scripts/globals/status")
+require("scripts/globals/msg")
+-----------------------------------------
+
+function onMagicCastingCheck(caster, target, spell)
+    local result = 0
+    if (target:getCampaignAllegiance() > 0) then
```
ibm2431: Standard "remove needless outer parens" comment 😉

scripts/globals/items/scroll_of_instant_retrace.lua
```diff
@@ -0,0 +1,24 @@
...
+-----------------------------------------
+
+function onItemCheck(target)
+    local result = 0
+    if (target:getCampaignAllegiance() > 0) then
+        result = 0
+    else
+        result = 56
+    end
+    return result
```
ibm2431: Slightly fewer lines if you want them:
```lua
if not (target:getCampaignAllegiance() > 0) then
    return 0
else
    return 56
end
```

TeoTwawki: outer parens not needed :P

ibm2431: Lua can be odd with `not` at times 💦

ibm2431: not

scripts/globals/spells/retrace.lua
Aeshur: I used Warp II as a reference since this spell also is cast on other people, it appears 4 is more correct though yeah. Wasn't aware of that, thanks.

scripts/globals/items/scroll_of_instant_retrace.lua
Aeshur: I was using the Scroll of Instant Warp item as a base line reference, but 1 did feel too short for this.

scripts/globals/items/scroll_of_instant_retrace.lua
Aeshur: I changed both item and spell to `if not` statements, and yeah the spell didn't work without the parentheses 😹

scripts/globals/teleports.lua
Aeshur: I re-ordered the tables to line up to allow for this method,  but after taking a look at examples and reading a bunch about it I was still pretty stumped at how `unpack` works exactly.

ibm2431: An easy way to think about `pack` and `unpack` is that they add brackets or remove them.

```lua
some_table = pack(1, 2, 3, 4) -- results in a table of {1, 2, 3, 4}
unpack(some_table) -- results in a return value of 1, 2, 3, 4 as separate arguments
```

Unpack is useful in the sense that lua lets you assign multiple values at once, because functions can return multiple values by default.

```lua
local pos = {123, 456, 789}
x, y, z = unpack(pos)
```
Would result in x being assigned to 123, y to 456, and z to 789

Since you'd be calling setPos directly, you can unpack it in the function call:
```lua
local pos = destinations[[ids.BASTOK_MARKETS_S]]
player:setPos(unpack(pos))
```
Should result in a function call to `setPos` equivalent to:
`player:setPos(-291, -10 , -107, 212, 87)`

zach2good (approved):
All seems fine to me, thanks for acting on the others' feedback so promptly 👍 

Today I learned about `upack()`, neat!

PR #1468, opened 2020-11-01
Add !addallweaponskills command

zach2good:
**_I affirm:_**

Needed this during testing, pretty useful

### DIFF
--- a/scripts/commands/addallweaponskills.lua
+++ b/scripts/commands/addallweaponskills.lua
@@ -1,0 +1,36 @@
+---------------------------------------------------------------------------------------------------
+-- func: addallweaponskills
+-- desc: Adds all learned weaponskills to the given target. If no target then to the current player.
+---------------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


ibm2431 (approved):

scripts/commands/addallweaponskills.lua
```diff
@@ -0,0 +1,36 @@
...
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!addallweaponskills {player}")
+end
+
+function onTrigger(player, target)
+
+    -- validate target
+    local targ
+    if (target == nil) then
```
TeoTwawki: suggestion:

```lua
    -- validate target
    local targ
    if target then
        targ = GetPlayerByName(target)
        if not targ then
            error(player, string.format("Player named '%s' not found!", target))
            return
        end
    else
        targ = player
    end
```
ditches outer parens and the == nil

Issue #1456, opened 2020-10-30
Dynamis bosses excessively cast death


MarianArlt:
**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

I noticed this doing classic Dynamis. The spell "death" has no further restrictions on itself, so bosses who have it on their spell list can select it on par with any other spell as if it was Rasp or Aero. I feel this is generally speaking an issue in the selection of magic spells on mobs, where there's no difference being made in the tier of spells.

What's even worse though, the Dynamis bosses seem to absolutely spam it and give it preference over other spells they should cast, like Breakga, Graviga etc. It's not that they don't cast anything else (after all their lists are correct), but about eight out of ten times or maybe even nine, they'll cast death as long as they have mp > 666. This is weird to me, because there is no obvious reason they should choose death more often than any other spell. Yet they clearly do.

Apparently there's different conditions on different mobs for the use of this spell to begin with. Some get access to it after a certain amount of HP, some after rereasing several times and others just have free access at all times. Diabolos for example can cast it quite frequently supposedly. I have watched a lot of classic Dynamis videos today though, and it's clear that these don't. ~~Instead they will absolutely [spam TP moves](https://www.youtube.com/watch?v=OfpFEbL6q70) probably due to very high regain and/or low TP trigger (video from april last year, only three people constantly hitting, doesn't even get to cast because of TP preference...)~~(wrong reference)

I also checked Dynamis Lord and he uses the rest of his list a bit more often, but still uses death way too frequently.

I wanted to propose a math.random condition right inside of "death.lua" but this would then just return and not cast any other spell when called, and as mentioned, death gets called almost always. Super weird.

TeoTwawki:
its a combination of our lack of any per-spell cooldowns for mobs (thus allowing same spell to be chosen again when the mobs global cast timer is up) and the spell selection code not matching how retail chooses spells.

Here is the logic for spell selection:
https://github.com/project-topaz/topaz/blob/e5750c30b8685a198661a8743ea92a13c91fee83/src/map/ai/controllers/mob_controller.cpp#L392

https://github.com/project-topaz/topaz/blob/8648c4071c1885ad1a388d74e7f868e39df2f75e/src/map/mob_spell_container.cpp#L123

[...]

PR #1469, opened 2020-11-01
Add SEVERE_SPELL_CHANCE mob modifier

MarianArlt:
**_I affirm:_**

This PR is a result of https://github.com/project-topaz/topaz/issues/1456 and Discussions on Discord about how to handle it.
It adds a new mob modifier called `SEVERE_CHANCE` that can be individually de- or increased where necessary.

We need to agree on:

- [x] default values (I set it to 5% on both any mob and BLM type [kinda superfluous])
- [x] spells that should go in this list (I added death, impact, meteor, meteor_ii and comet)
- [x] Maybe a more descriptive name (renamed to SEVERE_SPELL_CHANCE)

- [x] Built successfully today on top of freshly pulled upstream release branch and tested working as expected locally
(increased/decreased modifier from lua script on the fly)


scripts/globals/status.lua
```diff
@@ -2292,6 +2292,7 @@ tpz.mobMod =
     ALLI_HATE           = 68, -- Range around target to add alliance member to enmity list.
     NO_LINK             = 69, -- If set, mob cannot link until unset.
     NO_REST             = 70, -- Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
+    SEVERE_CHANCE       = 71, -- % chance to use a severe spell like death or impact
```
TeoTwawki: there were a few gaps in the list we could fill in, such as 
```lua

    SIGHT_RANGE         = 4,  -- sight range
    SOUND_RANGE         = 5,  -- sound range
    BUFF_CHANCE         = 6,  -- % chance to buff (combat only)
    GA_CHANCE           = 7,  -- % chance to use -ga spell
    HEAL_CHANCE         = 8,  -- % chance to use heal
    HP_HEAL_CHANCE      = 9,  -- can cast cures below this HP %
    SUBLINK             = 10, -- sub link group
    LINK_RADIUS         = 11, -- link radius
    DRAW_IN             = 12, -- 1 - player draw in, 2 - alliance draw in -- only add as a spawn mod!
    -- 13 Available for use
```
13 would put it closer to the existing chance mods

src/map/mob_modifier.h
```diff
@@ -98,6 +98,7 @@ enum MOBMODIFIER : int
     MOBMOD_ALLI_HATE          = 68, // Range around target to add alliance member to enmity list.
     MOBMOD_NO_LINK            = 69, // If set, mob cannot link until unset.
     MOBMOD_NO_REST            = 70, // Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
+    MOBMOD_SEVERE_CHANCE      = 71, // % chance to use a severe spell like death or impact
```
TeoTwawki: don't forget this one

src/map/mob_spell_container.cpp
```diff
@@ -60,7 +60,12 @@ void CMobSpellContainer::AddSpell(SpellID spellId)
         m_gaList.push_back(spellId);
 
     }
-    else if(spell->canTargetEnemy()){
+    else if(spell->isSevere()){
```
TeoTwawki: brace on newline please

src/map/mob_spell_container.cpp
```diff
@@ -60,7 +60,12 @@ void CMobSpellContainer::AddSpell(SpellID spellId)
...
+        // select spells like death and impact
+        m_severeList.push_back(spellId);
+
```
TeoTwawki: got an extra blank line here

src/map/mob_spell_container.cpp
```diff
@@ -136,6 +141,11 @@ std::optional<SpellID> CMobSpellContainer::GetSpell()
         }
     }
 
+    // try something really destructive
+    if(HasSevereSpells() && tpzrand::GetRandomNumber(100) < m_PMob->getMobMod(MOBMOD_SEVERE_CHANCE)){
```
TeoTwawki: hmm just realized the whole file has inline bracing before you got here. makes me want to auto-format the whole dang thing...We can do that separately from this pr >.>

src/map/mob_modifier.h
MarianArlt: Hmmm. Slot 13 is occupied by MOBMOD_RAGE // [...] spawn mod! (OBSOLETE - use rage mixin)
Even though it's obsolete, is it safe to overwrite this? Did a search for either MOBMOB_RAGE and tpz.mobmod.RAGE and neither has any occurrences. So I'd guess there won't be an issue? Please confirm.

TeoTwawki: then 13 should have been mobmod rage in status.lua also.

@ibm2431 @zach2good more desync between core and lua..

TeoTwawki: looks like ALL the ones marked "available for use" are in fact NOT :rage:

scripts/globals/status.lua
MarianArlt: As discussed I'll leave this at 71 for now. The obsolete mods should be discussed independently.

src/map/mob_modifier.h
MarianArlt: As discussed I'll leave this at 71 for now. The obsolete mods should be discussed independently.

src/map/mob_modifier.h
```diff
@@ -27,77 +27,78 @@ This is a list of mob specific modifiers. They can be added to pools / families
...
+    MOBMOD_HP_SCALE            = 61, // Scale the mobs max HP. ( hp_scale / 100 ) * maxhp
+    MOBMOD_NO_STANDBACK        = 62, // Mob will never standback
+    MOBMOD_ATTACK_SKILL_LIST   = 63, // skill list to use in place of regular attacks
+    MOBMOD_CHARMABLE           = 64, // mob is charmable
+    MOBMOD_NO_MOVE             = 65, // Mob will not be able to move
+    MOBMOD_MULTI_HIT           = 66, // Mob will have as many swings as defined.
+    MOBMOD_NO_AGGRO            = 67, // If set, mob cannot aggro until unset.
+    MOBMOD_ALLI_HATE           = 68, // Range around target to add alliance member to enmity list.
+    MOBMOD_NO_LINK             = 69, // If set, mob cannot link until unset.
+    MOBMOD_NO_REST             = 70, // Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
```
MarianArlt, line 101: Just for clarification all these simply got their padding increased by one column to improve verbosity of the new modifier at position 71.

TeoTwawki: if you could infill the "missing" ones into status.lua so we don't mistake them for having being free like the comment incorrectly says, I would appreciate it.

MarianArlt: See c3aaaa415a3c6a82d71503f95e22af45aefc591e

MarianArlt: After dialogue the new mod now occupies again slot 13, obsolete mods have been removed and two TODO mods marked as such. A comment has been added to the source containing the mirroring table and a comment about always editing both. See 611a174349f77bbd594a77881d7e528ed8da0b64

MarianArlt: After dialogue: fixed only for the affected lines by this PR for cleanliness of the PR itself. See ff24165b3b3b2dcc6f2b07f85ffad18954bbdbdb

src/map/mob_spell_container.cpp
MarianArlt: After dialogue: fixed only for the affected lines by this PR for cleanliness of the PR itself. See ff24165b3b3b2dcc6f2b07f85ffad18954bbdbdb

src/map/mob_spell_container.cpp
MarianArlt: After dialogue: fixed only for the affected lines by this PR for cleanliness of the PR itself. See ff24165b3b3b2dcc6f2b07f85ffad18954bbdbdb

src/map/mob_spell_container.cpp
MarianArlt: After dialogue: fixed only for the affected lines by this PR for cleanliness of the PR itself. See ff24165b3b3b2dcc6f2b07f85ffad18954bbdbdb

TeoTwawki (approved):
If nobody has a problem with the default % then I think this is fine

zach2good:
As always, I lean in favour of "more difficult is better; too easy doesn't get fixed" - so I would want it to go up to something like 20%, but I'm not particularly bothered.  If it _feels_ about right, that's good enough. Someone like @wrenffxi probably has a feeling of where this rate should be

EDIT: I haven't been keeping up in Discord, if this has been discussed

zach2good (approved):

MarianArlt:
I really like the idea behind "too easy doesn't get fixed". But I also think 1 out of 5 rolls is massively frequent for what this category should reflect *by default*. I remember doing Dynamis Lord back in era days and if he did that back then, the fight would've been pretty much impossible, burning through the tanks (I mean, we never filled 4 alliances and even less with each party having one tank, that sounds like legit japanese onry ^^;). I could be totally wrong though. Discord sadly didn't yield any opinions on this number. Would be nice to have  some more though.

TeoTwawki:
Another thing to consider is these rolls being independent of each other lets their order effect the outcome - we're rolling for a thing, then if it doesn't happen the next one gets rolled. By even adding another we've artificially reduced the chances of other outcomes ever so slightly. Sooo I guess its good that we're checking severe first in addition to excluding them from the other list, but in the future we should revisit spell selection so that we can be confident that 40% really means 40% (just using 40 as the example)

TeoTwawki:
5% felt low,. 20% felt high. I though 10% was just right /goldilocks

MarianArlt:
> [...]in the future we should revisit spell selection so that we can be confident that 40% really means 40%[...]

I'm glad you mention that Theo, that was one of my initial worries.

As for this: today I [watched](https://www.youtube.com/watch?v=D9noDsJGhJo) [again](https://www.youtube.com/watch?v=jQWKUWJWka0) a [lot](https://www.youtube.com/watch?v=Igot1O9pxWw) of Dynamis [captures](https://www.youtube.com/watch?v=lmpSL0Yw9qI) from the [capture](https://www.youtube.com/watch?v=jQWKUWJWka0) [channel](https://www.youtube.com/watch?v=6H-beKoFDwg) [and](https://www.youtube.com/watch?v=ywPUasgxe8I) [was](https://www.youtube.com/watch?v=q0JgVMHuDyE) [quite](https://www.youtube.com/watch?v=uevyCKUMN4U) [surprised](https://www.youtube.com/watch?v=Jl0JtpozXWA).
Turns out the non-Arch bosses use death at will and rather frequently when the Arch bosses will not at all.

|                   | Dynamis Captures        |             |
|-------------------|----------------|-------------|
|                   | Regular Bosses | Arch Bosses |
| Pops              | 5              | 5           |
| Total casts       | 27             | 60          |
| of which "severe" | 5              | 2           |
| % "severe" spells | 18.52%         | 3.33%       |

This is only of 10 fights in total but the tendency is quite obvious (even though it's all just math.random(), of course). The two severe spells from Arch came both from Arch Gu'Dha Effigy by the way and were Death and Impact. Additionally this includes 2 captures of NQ Dynamis Lord and 3 of Arch Dynamis Lord and none of these used Death even once (shocking!). Given this, the rate for the three city bosses might be even higher.

So I guess **the question is if you want a higher default and adjust down or have a lower default and adjust up**, since this category will not only serve Dynamis I guess.

TeoTwawki:
Higher and adjust down seems safer due to people tending to not want us to fix anything they were benefiting from and loudly proclaiming we nerfed it.

MarianArlt:
That's a really good point...
...Slap those 20% on there!

TeoTwawki (approved):

zach2good (approved):

PR #1472, opened 2020-11-03
Fix broken mob animations if interrupted during death

zach2good:
**_I affirm:_**

If `ActionType` isn't set, there is an area of empty space left behind. The bitpacking at the end of this packet sees that empty space and thinks "Ohh, free real estate" and puts information in there - resulting in a shift of the rest of the packet - which leads to broken animations.

Thanks to Setzor for the explanation!

### DIFF
--- a/src/map/packets/action.cpp
+++ b/src/map/packets/action.cpp
@@ -100,6 +100,8 @@ CActionPacket::CActionPacket(action_t& action)
         ref<uint8>(0x0C) = 0x1C;
         ref<uint8>(0x0D) = 0x5D;
         ref<uint8>(0x0E) = 0x19;
+
+        ActionType = ACTION_WEAPONSKILL_START;
     }
     break;
     case ACTION_MOBABILITY_FINISH:


src/map/packets/action.cpp
```diff
@@ -100,6 +100,8 @@ CActionPacket::CActionPacket(action_t& action)
         ref<uint8>(0x0C) = 0x1C;
         ref<uint8>(0x0D) = 0x5D;
         ref<uint8>(0x0E) = 0x19;
+
+        ActionType = ACTION_WEAPONSKILL_START;
```
ibm2431, line 104: I take exception to assigning the action type of [weaponskill start](https://github.com/Windower/Lua/wiki/Action-Category-03) inside a [mobskill interrupt](https://github.com/Windower/Lua/wiki/Action-Category-07) case.

But we apparently use `ACTION_WEAPONSKILL_START` for mobskill starts too, so I'm just going to throw my hands in the air and pretend I didn't see this. 🤷

ibm2431 (approved):
I guess we can save accurate enum names for when action are rewritten.

zach2good:
>  when action are rewritten.
😱

PR #1477, opened 2020-11-04
Created setallskills.lua

Laserstrahl:
Script to set all skills to the given level.

**_I affirm:_**

### DIFF
--- a/scripts/commands/setallskills.lua
+++ b/scripts/commands/setallskills.lua
@@ -1,0 +1,64 @@
+---------------------------------------------------------------------------------------------------
+-- func: setallskills <skill level> <target>
+-- desc: sets target's level for all skills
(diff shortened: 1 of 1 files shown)


TeoTwawki:
it would be better to modify the one that caps all skills to have an optional parameter, and default to maxing them out (its current behavior) than to add yet another command that manipulates skills. Since they cap at different levels this isn't all that useful to warrant a new stand alone command.

edit: or even edit the one that sets one skill to accept "all" as a skill name to do this. because naming.

zach2good:
EDIT: Agree with above, I think the functionality in this script might be useful to some, but not enough to warrant a whole new script.

I'm kinda into modifying the existing `!capallskills`, so that it sets every skill to its max value, and `!capallskills 100` _attempting_ to set every skill to 100, if it can get there.

PR #1480, opened 2020-11-05
Fixes Gobbie Mystery Box "You must wait another 0 days..."

neuromancerxi:
unknown

Adds 1 to final amount printed to the player to prevent "You must wait another 0 days to perform that action." Retail count starts at 45, not 44. 

Thanks again Nyu for the retail confirmation of count starting at 45 days on a new character. Character created just before JST Midnight showed 45 days before JST midnight and after JST midnight.

**_I affirm:_**

### DIFF
--- a/scripts/globals/gobbiemysterybox.lua
+++ b/scripts/globals/gobbiemysterybox.lua
@@ -126,7 +126,7 @@ tpz.mystery.onTrigger = function (player, npc, events)
             player:startEvent(event.DEFAULT, specialDialUsed, adoulinDialUsed, pictlogicaDialUsed, wantedDialUsed, 0, 0, hideOptionFlags, dailyTallyPoints)
         end
     else
-        player:messageSpecial(zones[player:getZoneID()].text.YOU_MUST_WAIT_ANOTHER_N_DAYS, GOBBIE_BOX_MIN_AGE - playerAgeDays)
+        player:messageSpecial(zones[player:getZoneID()].text.YOU_MUST_WAIT_ANOTHER_N_DAYS, GOBBIE_BOX_MIN_AGE - playerAgeDays + 1)
     end
 end
 


scripts/globals/gobbiemysterybox.lua
```diff
@@ -126,7 +126,7 @@ tpz.mystery.onTrigger = function (player, npc, events)
             player:startEvent(event.DEFAULT, specialDialUsed, adoulinDialUsed, pictlogicaDialUsed, wantedDialUsed, 0, 0, hideOptionFlags, dailyTallyPoints)
         end
     else
-        player:messageSpecial(zones[player:getZoneID()].text.YOU_MUST_WAIT_ANOTHER_N_DAYS, GOBBIE_BOX_MIN_AGE - playerAgeDays)
+        player:messageSpecial(zones[player:getZoneID()].text.YOU_MUST_WAIT_ANOTHER_N_DAYS, GOBBIE_BOX_MIN_AGE - playerAgeDays + 1)
```
cocosolos, line 129: ~~This will fix the display but it's still not ticking on JST midnight like it should. Instead of changing this I think `playerAgeDays` needs to account for JST midnight since it currently uses `os.time()`. I think you can just switch it to `JstMidnight()`.~~

neuromancerxi: Based on the data from Nyu, it's not based on JP midnight, but character creation time stamp.

How this was tested:

1. Character created same day and spoke with Gobbie at ~23:14 JST
2. Character spoke with gobbie, "You must wait another 45 days to..."
3. Waited until after JST midnight (12:01 JST)
4. Character spoke with gobbie, "You must wait another 45 days to..."

This debunked my theory that it was based on JST Midnight, seems to confirm that it's based on character creation time stamp, and lends credence to the fix as submitted. (Screenshots below are taken in Time Zone UTC+1)

image

image

cocosolos: Oh I misunderstood or misread your first post, my bad! 😫

neuromancerxi: All good, I can see how my last sentence in the PR could be ambiguous. Hopefully this does a better job of painting the picture. :+1:

zach2good:
I was about to try and argue that there might be some rounding shenanigans going on here, but we're already dealing with days, so the rounding has already happened. I'm fine with the `+1`, rather than trying to dig in and find the epoch  times and `ceil` them to the next day so the maths sorts itself out 👍

zach2good (approved):

PR #1486, opened 2020-11-08
Trust Cipher CS Args

zach2good:
**_I affirm:_**

Shoutout to @ibm2431 for the advice about packing more info into item subIds


TeoTwawki (approved):
looks more gooder to me :+1:

ibm2431 (approved):

ibm2431:
I'll let you have the merge when the subIDs are adjusted~

zach2good:
After discussion, we've decided to just add the text for "non-player race" ciphers, since we don't yet have proof of who gets which text. There is a free period coming up, so we might be able to extract patterns at that time

PR #1488, opened 2020-11-09
[WIP] Trusts: More accurate HP/MP stats

zach2good:
**_I affirm:_**

image

But in all seriousness, I'll be using the calculations for player health:

http://ffxi-stat-calc.sourceforge.net/cgi-bin/ffxistats.cgi?mode=document

and the calculations for monster health (go see `mobutils.cpp`).

The ratios are currently right (Paladins have significantly less MP than White Mages, for instance), but the numbers seem low. So I need to dig into the differences between players and trusts for this (be it stats, gear, race contribution etc.)

**HEFTY REMINDER** That the options to boost trust hp, mp, stats, and skills have existed in `map.conf` for six months. If you don't like the stats becoming more accurate you are encouraged to tweak them to your liking :)

### DIFF
--- a/src/map/utils/trustutils.cpp
+++ b/src/map/utils/trustutils.cpp
@@ -3,6 +3,7 @@
 #include "../../common/timer.h"
 #include "../../common/utils.h"
 
+#include <algorithm>
 #include <cmath>
 #include <cstring>
 #include <vector>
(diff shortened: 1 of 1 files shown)


ibm2431:
To be clear: Making something more accurate to retail is not a nerf.

If I see complaints about our pursuit of accuracy, I _will_ start throwing out bans.

The - to be frank, outright whining - from Zach previously trying to make trust MP pools more accurate was toxic, mentally exhausting, and even slowed down development of _completely unrelated_ things.

It will _not_ be tolerated this time.

If there are complaints about trust stat adjustments, they should be based on being inaccurate to retail. If you are unable to show a retail inaccuracy, then you have nothing to say on this matter.

src/map/utils/trustutils.cpp
```diff
@@ -345,43 +346,128 @@ void LoadTrustStatsAndSkills(CTrustEntity* PTrust)
...
+    auto scaleToGrade = [mapRanges](float input) -> unsigned int
+    {
+        unsigned reverseMappedGrade = mapRanges(80U, 140U, 1U, 7U, static_cast<unsigned int>(input * 100U));
+        reverseMappedGrade += 1U; // Boost grades
+        return std::clamp(7U - reverseMappedGrade, 1U, 7U);
+    };
+
+    // HP/MP ========================
+    // This is the same system as used in charutils.cpp, but modified
+    // to use parts from mob_family_system instead of hardcoded player
```
TeoTwawki, line 373: Thank you. I almost balked at the removal of the above todo before seeing this comment. I still believe there will be more redoes of the stat calcs in the future, but _mobs themselves_ need a general cleanup too so accuracy would be debatable either way.

PR #1490, opened 2020-11-10
Add MAX_SWINGS, ADDITIONAL_SWING_CHANCE mod and JOB_MULTIPLE latent

MarianArlt:
**_I affirm:_**

This PR seeks to add:
• a **latent** of name **`JOB_MULTIPLE_8`** for weapons that trigger an effect while the players job is a multiple of 8 _(Octave Club)_
• a **modifier** of name **`ADDITIONAL_SWING_CHANCE`** for weapons that can trigger an additional attack *on top of* other multiple hits like Double Attack as a latent.
• a **modifier** of name **`MAX_SWINGS`** for weapons that increase their hitCount while under a latent effect.
• the latent database entry for Octave Club

As a consequence of this, this PR also had to:
• create conditions for the additional swings in `CAttackRound::CreateAttacks()`
• slightly restructure the conditions of additional swings and how they block each other in that same function
• See 4ea931fc3cf1e8db9e73e5982d087af1bee47341
(Mythic Aftermath was not blocked behind Quad/Triple/Double Attack and latent modifiers for OAT not considered)

In Detail:
• If `MAX_SWINGS` exists (should pretty much only be flagged by a latent) it will overwrite the hitCount of that weapon from the database
• The default swing now functions as a fallback after all else is considered. This improves recognition of priorities
• Quad/Triple/Double attack got their num check removed, they should be rolled in any case, also when a weapon or monster can attack several times. It is the top most priority and does not stack with regular multiple hits or mythic modifiers
• Triple/Double attack from Mythic weapons gets rolled after this if Aftermath is in effect
• Equipment modifiers from Empyrian armor is then considered
• Only if any of these fail will a weapon or monster use regular multiple hits
• In preparation for Amood Great Axes a modifier and a condition has been added especially for these, since they do indeed stack
• After all these considerations, ammo triggered additional swings from weapons like the Jailer weapons are considered, the check itself has not been changed
• All of the above are possible *additional swings*, care has been taken to make sure they can not stack up to more than 8 hits
• At the very end a default hit will always occur

This makes implementation of most weapons that increase their hitCount with a latent or some other effect now possible and secure with these mods. (Although there are extremely few to begin with)

I tested this successfully with the following setups and it works as expected.
• RDM99/PLD49
Regular Weapon: one hit
Joyeuse: occ. 2-3 hits
Kraken Club: occ. 2-8 hits
Octave Club: one hit
• RDM96/PLD48
Octave Club: occ. 2-8 hits
• WAR99/PLD49
Regular Weapon: occ. Double Attack
Joyeuse: occ. 2-3 hits with more 2's than 3's
Kraken Club: occ. 2-8 hits with a few more 2' than others
Octave Club: occ. Double Attack
• THF99/PLD49
All of the above in the same manner but with triple attack
Justice Sword + Virtue Stones: Successfully triggers an additional hit even on top of Triple Attack, making it 4 hits
Vajra + 3000TP Aftermath: occ. 2-3 hits with frequent 2's
• SAM99/PLD49
Hardwood Katana + Zanshin: occ. hits again after missing
_(Zanshin is not a member of the function in question but I wanted to be sure it works)_
Hardwood Katana + Virtue Stone: stones won't get accidentally get used
Hardwood Katana + Zanshin + Triple/Double Attack equip: Zanshin swing stacks

Things I did not implement or I could not figure out but should not be part of this PR:
• Empyrian armor modifiers have various effects on attack swings but there is little people talking about whether it stacks or not
(for now I think it is sensible to treat the Iga Garb condition that was there as if it was a roll on Double Attack)
• Somebody before me who implemented this condition mentioned something about creating a lua function, but it was not very clear what "nitty gritty stuff" refers to. I suppose its the idea of outsourcing these?
• I did not implement the missing condition for Dynamis [D] Follow-Up attack chances
• Octave Club is probably the only item in the game (?) that checks on two different lvl multipliers at the same time. This simply interferes with how the according latent switch statements are written and neither will flag in that condition.
(for now comment the multiplier of 2 out)

I'm well aware that this messes with one of the most basic mechanics of the game. Please take the time to read what I wrote as carefully as I wrote and tested this PR. Much appreciated.

### DIFF
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -2946,8 +2946,8 @@ INSERT INTO `item_latents` VALUES(18850, 366, 5, 6, 1000);  -- DMG+5 while TP <1
 -- -------------------------------------------------------
 -- Octave Club
 -- -------------------------------------------------------
-INSERT INTO `item_latents` VALUES(18852, 977, 2, 38, 2);    -- Occasionally attacks 2 times when mjob multiple of 2
-INSERT INTO `item_latents` VALUES(18852, 977, 6, 38, 8);    -- Occasionally attacks 2 to 8 times when mjob multiple of 8
+INSERT INTO `item_latents` VALUES(18852, 978, 2, 38, 2);    -- Occasionally attacks 2 times when mjob multiple of 2
+INSERT INTO `item_latents` VALUES(18852, 978, 6, 38, 8);    -- Occasionally attacks 2 to 8 times when mjob multiple of 8
 
 -- -------------------------------------------------------
 -- Kerykeion
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -2947,7 +2947,7 @@ INSERT INTO `item_latents` VALUES(18850, 366, 5, 6, 1000);  -- DMG+5 while TP <1
 -- Octave Club
 -- -------------------------------------------------------
 INSERT INTO `item_latents` VALUES(18852, 977, 2, 38, 2);    -- Occasionally attacks 2 times when mjob multiple of 2
-INSERT INTO `item_latents` VALUES(18852, 977, 8, 38, 8);    -- Occasionally attacks 2 to 8 times when mjob multiple of 8
+INSERT INTO `item_latents` VALUES(18852, 977, 6, 38, 8);    -- Occasionally attacks 2 to 8 times when mjob multiple of 8
 
 -- -------------------------------------------------------
 -- Kerykeion
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -2946,8 +2946,7 @@ INSERT INTO `item_latents` VALUES(18850, 366, 5, 6, 1000);  -- DMG+5 while TP <1
 -- -------------------------------------------------------
 -- Octave Club
 -- -------------------------------------------------------
--- TODO: Optimize latent_efffect_container.cpp to allow a mod be overwritten with a higher value
--- INSERT INTO `item_latents` VALUES(18852, 977, 2, 38, 2); -- Occasionally attacks 2 times when mjob multiple of 2
+INSERT INTO `item_latents` VALUES(18852, 977, 2, 38, 2);    -- Occasionally attacks 2 times when mjob multiple of 2
 INSERT INTO `item_latents` VALUES(18852, 977, 8, 38, 8);    -- Occasionally attacks 2 to 8 times when mjob multiple of 8
 
 -- -------------------------------------------------------
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -78,14 +78,9 @@ INSERT INTO `item_latents` VALUES(10975, 25, 13, 52, 8);   -- DARK WEATHER:ACC
 INSERT INTO `item_latents` VALUES(10975, 23, 13, 52, 8);   -- DARK WEATHER:ATT
 
 -- -------------------------------------------------------
--- Eerie Cloak +1
--- -------------------------------------------------------
-INSERT INTO `item_latents` VALUES(11300, 369, 1, 40, 0);    -- Level multiple of 13 and Nighttime: Refresh +1MP/tick
-
--- -------------------------------------------------------
 -- Eerie cloak +1
 -- -------------------------------------------------------
-INSERT INTO `item_latents` VALUES(11301, 369, 1, 40, 0);    -- Refresh+1 at night when the level of the player's main job is a multiple of 13.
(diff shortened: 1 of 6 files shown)


TeoTwawki:
Some things you should no: multi hit weapons like Joyeuse were already implemented, it was a field in item_weapon table_**however**_ I do want to see that column die and it migrate to a mod as you have here. Octave club is a good example of why that should be the case.

I'd like to also see the job_multple_lv latents changed slightly: we have a parameter field we can use on a latent. Instead of making it a static 8 for this one item, make it multiple of the specified parameter. Then we don't need a new latent ID used every time we need a different multiple number (there are several right now). MAINJOB_MULTIPLE set param and done.

TeoTwawki:
nope fk all this editing, reposting

TeoTwawki:
Some things you should know before reading this pr: multi hit weapons like Joyeuse were already implemented, it was a field in item_weapon table however I do want to see that column die and it migrate to the mod as is done here. Octave club is a good example of why that should be the case.

I'd like to also see the `job_multple_lv latents` changed slightly: we have a parameter field we can use on a latent. Instead of making it a static 8 for this one item, make it multiple of the specified parameter. Then we don't need a new latent ID used every time we need a different multiple number (there are several right now). `MAINJOB_MULTIPLE` set param and done.

kaincenteno:
Awesome job =)

MarianArlt:
As suggested by Teo I also reduced the various JOB_MULTIPLE_X latents into one single latent of name `JOB_MULTIPLE` that takes a divisor (2 for even anyways) or 0 for odd.
Entries in item_latents.sql have been adjusted accordingly. (very few thank god...)
The night check for Eerie cloak has to be separate and was named `JOB_MULTIPLE_AT_NIGHT`.
Eerie cloak had a false duplicate in the database where the NQ version was also given the latent effect. This was removed.

The force pushes corrected a comment and a missing semicolon.

TeoTwawki (approved):

TeoTwawki:
I wish I could per-emptively re-approve for the force pushes I know are coming :x


###### something somethings sql rows and hit count overflows

MarianArlt:
No need to force push.
• `MAX_SWINGS` has been manually limited to 8 inside of `CAttackRound::CreateAttacks()` because it wouldn't get clamped by `getHitCount()` for some reason. This led to no hit getting added at all. It is a safety measure and doesn't hurt.
• Now that this is sorted Octave Club has been activated for both latent values in the database.
• `JOB_MULTIPLE_AT_NIGHT` was missing its check for odd levels. (is not used currently but makes it future proof)

- [x] Tested locally (not multiple_at_night to be perfectly honest with you)

This closes #1498 and makes me comfortable saying that I consider this done now. (except for any requests of course)

zach2good (approved):
All seems reasonable to me, but its a very core-gameplay change, so I'll wait for @ibm2431 to check it over 👍

sql/item_latents.sql
```diff
@@ -2946,8 +2946,7 @@ INSERT INTO `item_latents` VALUES(18850, 366, 5, 6, 1000);  -- DMG+5 while TP <1
 -- -------------------------------------------------------
 -- Octave Club
 -- -------------------------------------------------------
--- TODO: Optimize latent_efffect_container.cpp to allow a mod be overwritten with a higher value
--- INSERT INTO `item_latents` VALUES(18852, 977, 2, 38, 2); -- Occasionally attacks 2 times when mjob multiple of 2
+INSERT INTO `item_latents` VALUES(18852, 977, 2, 38, 2);    -- Occasionally attacks 2 times when mjob multiple of 2
 INSERT INTO `item_latents` VALUES(18852, 977, 8, 38, 8);    -- Occasionally attacks 2 to 8 times when mjob multiple of 8
```
TeoTwawki: Technically, that should be 6 not 8 still. The clamp stops it from turning into 10, but it might not be obvious to new people looking back at this in the future that the modifier is additive.

MarianArlt: I thought it'd make sense to be more verbose so to make it use the actual desired number, but you're right. Adjusted it accordingly.

ibm2431:
> so I'll wait for @ibm2431 to check it over

Nothing particularly stands out to me.

So long as it caps at 8 and additional hits are proccing off the correct hand... 🤷 

I think I've telegraphed what might be early targets the next time I touch core after my current work. I'm not gonna get fussed over virtue stones and Amood~

MarianArlt:
last push was a request by @neuromancerxi to avoid a merge conflict with #1478 
modifier identifiers were increase by one since that PR will occupy `977` .
New identifiers for these mods will occupy `978` and `979`. Octave club was adjusted accordingly.
This will leave SPAREs 980-982 after this PR.

_(force push cause I forgot to edit status.lua , sorry)_

ibm2431:
Merged into release.

PR #1494, opened 2020-11-10
Fragile Code Cleanup

zach2good:
**_I affirm:_**

**NOT YET SANITY TESTED**

I set up static analysis with Sonar, and started fixing all the things it complained about. Each commit explains what I'm doing.

The biggest thing is in the Packet System:
- `session` was poorly named, and shadowed a global session var -> renamed to `PSession`
- Every packet handler made copies of `CBasicPacket`, which isn't a big deal - `CBasicPacket` is just a bag of pointers, but we still shouldn't be doing that in the hot path 🤷‍♂️ Swapped it for move semantics gubbins

### DIFF
--- a/src/map/linkshell.cpp
+++ b/src/map/linkshell.cpp
@@ -484,8 +484,9 @@ namespace linkshell
                     LinkshellList.erase(PItemLinkshell->GetLSID());
                 }
             }
-            catch (std::out_of_range&)
+            catch (std::out_of_range& exception)
             {
+                ShowError("linkshell::DelOnlineMember caught exception: %s\n", exception.what());
             }
         }
         return false;
--- a/src/common/zlib.cpp
+++ b/src/common/zlib.cpp
@@ -4,6 +4,7 @@
 #include <string>
 #include <cstring>
 #include <cassert>
+#include <memory>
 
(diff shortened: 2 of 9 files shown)


src/map/linkshell.cpp
```diff
@@ -484,8 +484,9 @@ namespace linkshell
                     LinkshellList.erase(PItemLinkshell->GetLSID());
                 }
             }
-            catch (std::out_of_range&)
+            catch (std::out_of_range& e)
```
TeoTwawki: can we use more than one letter for the exception, so it becomes obvious to newbies what this is? I just don't think "e" will cut it, and that will also stop them from googling what this is.

TeoTwawki (approved):
wrong button

TeoTwawki (changes requested):
only thing I caught complain worthy was that little e

Kreidos (approved):
`e` is a pretty common variable convention to me, but maybe it's just my Java and Python experience has poisoned me. :stuck_out_tongue:

TeoTwawki:
> `e` is a pretty common variable convention to me, but maybe it's just my Java and Python experience has poisoned me. stuck_out_tongue

knee deep in the venom of experience, the young'uns need time to build up resistance

PR #1497, opened 2020-11-11
Add new chat channels to the enums

TeoTwawki:
**_I affirm:_**

### DIFF
--- a/src/map/packets/chat_message.h
+++ b/src/map/packets/chat_message.h
@@ -61,7 +61,7 @@ enum CHAT_MESSAGE_TYPE
     MESSAGE_UNKNOWN_32    = 32, // Looks the same as 31
     MESSAGE_UNITY         = 33,
     MESSAGE_JP_ASSIST     = 34,
-    MESSAGE_NA_ASSIST     = 35
+    MESSAGE_NA_ASSIST     = 35,
 };
 
 /************************************************************************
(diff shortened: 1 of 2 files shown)


TeoTwawki:
Errata:
- Mentors may speak in these channels
- New players may speak in these channels, until such time as their new player status wears off (removing via the helpdesk menu does not remove access
- Returning players may speak in these channels, until 48 hours have passed
- For everyone else its just spam
- access is via `/assistj` and `/assiste` have fun confusing them with the existing `/assist` command that has nothing to do with chat :+1:

zach2good:
Are all of those conditions enforced by the client? Or are we going to have to do something with them?

zach2good (approved):

src/map/packets/chat_message.h
```diff
@@ -59,7 +59,9 @@ enum CHAT_MESSAGE_TYPE
     MESSAGE_LINKSHELL3    = 30, // Yes really, it looks like a 3rd LS may have been planned at some point.
     MESSAGE_NS_LINKSHELL3 = 31, // (assumed, as it follows pattern and color)
     MESSAGE_UNKNOWN_32    = 32, // Looks the same as 31
-    MESSAGE_UNITY         = 33
+    MESSAGE_UNITY         = 33,
+    MESSAGE_JP_ASSIST     = 34,
+    MESSAGE_NA_ASSIST     = 35
```
zach2good: Please leave `,`'s at the end of enums so there are less line changes later - should SE decide we need more useless channels in the future 🙏

zach2good (approved):

TeoTwawki:
> Are all of those conditions enforced by the client? Or are we going to have to do something with them?

I would think at a min the 48 hr timer has to be purely server side and we’d want the rest enforced there anyway when we get around to full implementation because packets can be sent by tools. 

Right now the server ignores these because there is no case entry in the switch.

src/map/packets/chat_message.h
TeoTwawki: Will do shortly

TeoTwawki: done

PR #1502, opened 2020-11-11
Auras: Fill in logic for mob-based auras

zach2good:
**_I affirm:_**

Not yet tested, will have time at the weekend.

Side note; iterating a mob's enmity container suuuuuucks.
```cpp
            for (auto [idx, entry] : *PMob->PEnmityContainer->GetEnmityList())
            {
                CBattleEntity* PTarget = entry.PEnmityOwner;
```


ibm2431:
Wondering if some genericism with var names / loops might result in us being able to combine the logic for the four branches into one, with the if check just determining which list is getting iterated.

zach2good:
I was thinking the same thing, maybe a "generate lists" section, and then "iterate lists" after

zach2good:
Another option is for me to make `EnmityList` iterable, but I'm sure I'll fight `EnmityContainer` and friends another time

scripts/globals/geo.lua
```diff
@@ -8,23 +8,27 @@ require("scripts/globals/status")
...
--- TODO: After elements are aligned in the codebase, this should become:
--- tpz.geo.spawnLuopan = function(player, target, spell, tick_effect, tick_power, target_type)
-tpz.geo.spawnLuopan = function(player, target, modelID, tick_effect, tick_power, target_type, spell)
+tpz.geo.spawnLuopan = function(player, target, spell, tickEffect, tickPower, targetType)
 
     tpz.pet.spawnPet(player, tpz.pet.id.LUOPAN)
     local luopan = player:getPet()
 
+    local element = spell:getElement()
+    local visibleAura = element - 1
```
ibm2431, line 17: I greatly approve.

ibm2431 (approved):

PR #1503, opened 2020-11-11
Fix BST Ready timer does not accumulate charges after Call Beast.

neuromancerxi:
Fixes Ready Recast timer would not add charges, and would sit at 0 charges
without ever refilling. Adds 1s recast timer to trigger charges and regular
refilling based on values in sql/abilities_charges.

**_I affirm:_**

### DIFF
--- a/scripts/globals/abilities/call_beast.lua
+++ b/scripts/globals/abilities/call_beast.lua
@@ -25,4 +25,5 @@ end
 function onUseAbility(player, target, ability)
     tpz.pet.spawnPet(player, player:getWeaponSubSkillType(tpz.slot.AMMO))
     player:removeAmmo()
+    player:addRecast(tpz.recast.ABILITY, 102, 1)
 end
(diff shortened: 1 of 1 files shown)


scripts/globals/abilities/call_beast.lua
```diff
@@ -25,4 +25,5 @@ end
 function onUseAbility(player, target, ability)
     tpz.pet.spawnPet(player, player:getWeaponSubSkillType(tpz.slot.AMMO))
     player:removeAmmo()
+    player:addRecast(tpz.recast.ABILITY, 102, 1)
```
zach2good, line 32: Is this 102 the entry in `abilities_charges`? I'd say leave a small comment explaining where to find it, it isn't a well known table

zach2good (changes requested):
Looks good, ready to merge after adding one comment for clarity 👍

zach2good (approved):

zach2good:
Don't worry about the CI, it's just complaining about nothing:
```
Error: ;31m[Error] bind failed with error: 98
Error: Process completed with exit code 255.
```

PR #1504, opened 2020-11-11
Add Lua Hook for character's lastOnline timestamp

Kreidos:
Needed for @kaincenteno and his in-progress work on Login Points. Simple Lua hook to return the Unix timestamp of the last online time.

**_I affirm:_**

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -672,6 +672,30 @@ inline int32 CLuaBaseEntity::resetLocalVars(lua_State* L)
 }
 
 /************************************************************************
+*  Function: getLastOnline()
+*  Purpose : Returns the unix timestamp of the last time the char logged off or zoned
+*  Example : player:getLastOnline()
+*  Notes   :
+************************************************************************/
+
+inline int32 CLuaBaseEntity::getLastOnline(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+
+    if (auto PChar = static_cast<CCharEntity*>(m_PBaseEntity); PChar->objtype == TYPE_PC)
+    {
+        lua_pushinteger(L, PChar->lastOnline);
+    }
+    else
+    {
+        lua_pushnil(L);
+    }
+
+    return 1;
+}
+
+/************************************************************************
 *  Function: injectPacket()
 *  Purpose : Injects a packet to the player's client
 *  Example : player:injectPacket(packet)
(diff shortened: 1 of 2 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -671,6 +671,30 @@ inline int32 CLuaBaseEntity::resetLocalVars(lua_State* L)
...
+*  Example : player:getLastOnline()
+*  Notes   :
+************************************************************************/
+
+inline int32 CLuaBaseEntity::getLastOnline(lua_State* L)
+{
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+
+    if (auto PChar = static_cast<CCharEntity*>(m_PBaseEntity); PChar->objtype == TYPE_PC)
```
zach2good: `static_cast` forces the cast without any checking, so if debug breaks are off and a mob is passed in here, this check will still pass. Granted, it'll get caught by `PChar->objtype == TYPE_PC`, but it's a practice we shouldn't be relying on;

```cpp
if (auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity))
{
    ....
}
```

will honour nullptr, and check that the pointer is in the same inheritance tree, rather than hard-casting the pointer by virtue of them being the same size.

zach2good (changes requested):
Seems fine, nice and easy - single casting ding then ready to go 👍

src/map/lua/lua_baseentity.cpp
```diff
@@ -683,7 +683,7 @@ inline int32 CLuaBaseEntity::getLastOnline(lua_State* L)
     TPZ_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
     TPZ_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
 
-    if (auto PChar = static_cast<CCharEntity*>(m_PBaseEntity); PChar->objtype == TYPE_PC)
+    if (auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity); PChar && PChar->objtype == TYPE_PC)
```
zach2good: Almost, after `dynamic_cast`, these checks are redundant; as they're implied by `dynamic_cast`: `PChar && PChar->objtype == TYPE_PC`

All you need is:
```cpp
if (auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity))
{
...
```

zach2good: Ah, just saw your comment:

> TYPE_PC check needs to stay, since lua_baseentities could be other types and only PCs have valid online times.

`dynamic_cast<CCharEntity*>` will only return a 'usable' pointer if the target has vtable entries and memory allocated for the type it's trying to cast to, otherwise it will fail. Like you mention; `static_cast` will just smush the pointer into the new type and hope for the best. It would have the base class available, but would fail when it tries to get anything from `CharEntity`.

zach2good (approved):

src/map/lua/lua_baseentity.cpp
Kreidos: Good to know

PR #1505, opened 2020-11-12
Add introductory dialogues for the Dynamis currency goblins

MarianArlt:
**_I affirm:_**

After getting access to Dynamis and having been instructed to go get a Prismatic Hourglass, any Dynamis currency goblin in Beadeaux, Davoi or Castle Oztroja will [talk to the player about the key item](https://youtu.be/mD91QXsDvSM?t=410) at this point. Once.
After that all goblins will introduce their currency to the player. Once.
After that they will all show the menu.

This PR adds the missing events and makes it so that:
• either of the goblins will introduce the hourglass. After doing so, the others won't.
• the player may skip this dialogue by trading 50k gil right away without talking to the goblin and receiving the hourglass
• before offering the dialogue menu each goblin will introduce the currency he trades once and then never again
• after all three goblins have introduced their currency to the player the charVar responsible for this is deleted
• the player may move freely between these three in any state of this logic without messing anything up

A lot of the lines in the commit are just parenthesis clean up. The important bit is the `onTrigger()` function in each goblins script and the resulting `onEventFinish()`.

### DIFF
--- a/scripts/zones/Davoi/npcs/Lootblox.lua
+++ b/scripts/zones/Davoi/npcs/Lootblox.lua
@@ -37,23 +37,24 @@ local maps = {
 function onTrade(player, npc, trade)
     local gil = trade:getGil()
     local count = trade:getItemCount()
+    local introMask = player:getCharVar("currencyGoblinIntro")
 
-    if (player:hasKeyItem(tpz.ki.VIAL_OF_SHROUDED_SAND)) then
+    if player:hasKeyItem(tpz.ki.VIAL_OF_SHROUDED_SAND) then
 
         -- buy prismatic hourglass
-        if (gil == PRISMATIC_HOURGLASS_COST and count == 1 and not player:hasKeyItem(tpz.ki.PRISMATIC_HOURGLASS)) then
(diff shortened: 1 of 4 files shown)


scripts/zones/Davoi/npcs/Lootblox.lua
```diff
@@ -123,22 +147,37 @@ function onEventUpdate(player, csid, option)
...
 function onEventFinish(player, csid, option)
+    local introMask = player:getCharVar("currencyGoblinIntro")
+
+    if csid == 131 then
+        -- make prismatic hourglass introduction unavailable to the other goblins
+        player:setCharVar("currencyGoblinIntro", utils.mask.setBit(introMask, 4, true))
+
+    elseif csid == 132 then
+        -- remove if this was the last of the three goblins that introduced currency
+        if utils.mask.getBit(introMask, 2) and utils.mask.getBit(introMask, 3) then
```
TeoTwawki: can just check if the mask is full right after setting this goblin's bit, theres a function for it along with the get's for the bits.

TeoTwawki: it works either way I just want you to know that exists. I was looking for it and missed your zero'ing out the var at 1st.

MarianArlt: I very much like this pointer and implemented it. It's a cleaner check and less error prone.

PR #1510, opened 2020-11-12
Corrected CS ID value for NPC Dhea Prandoleh involvement in beginning of quest The Tigress Strikes

EpicTaru:
…igress Strikes

-- Usage of CS ID 135 here doesn't allow onEventFinish portion of script to fire off and add quest The Tigress Strikes.
---- This subsequently causes NPC Rotih Moalghett in Fort Karugo Narugo [S] to not update a variable needed to proceed further and then the snowball starts forming...
-- CS ID 135 appears to just be dialogue that is used as part of the dialogue of CS ID 133.
-- No snowballs were harmed in this submission.

**_I affirm:_**


ibm2431 (commented):
Confirming from capture that 133 is the proper event to start.

GitHub still won't let me comment on arbitrary lines that weren't diffed. This sequence:
```lua
        if player:getCharVar("TigressStrikesProg") < 3 then
            player:startEvent(131)
```
Is where event 135 _should_ be, not 131.

EpicTaru (commented):
Adjustment made, good sir.

Snowball 131 _**was harmed**_ in this adjustment.

ibm2431 (commented):
Was originally merged into a safekeeping branch, but I'll "approve" this and merge it into release~

PR #1523, opened 2020-11-14
Adds missing stat_wsc MODs and Logic

neuromancerxi:
November 2020 Version Update added two items which have WS AGI+ and WS INT+
Using the same logic from Utu Grip (WS DEX +), added MODs and checks to apply them.

Creates MODs and checks for future use cases outside of AGI, INT, and/or DEX.
Updates comments to site sources for calculation details.

Note: MOD Ids have been based on two pending PRs at the time of submission:

1478 - Uses MOD 977
1490 - Uses MOD 978 and 979

**_I affirm:_**


ibm2431:
Should we go ahead and reserve similar mods for all base attributes now?

neuromancerxi:
> Should we go ahead and reserve similar mods for all base attributes now?

Wouldn't be opposed to expanding the PR to do so. Though with the use of ENUMs it does seem rather trivial to add later. Either way, I'll add them if they're likely to be accepted.

ibm2431:
Wren says it looks sane and that it works! Merged into release.

Issue #1524, opened 2020-11-14
Client Version 30201103_1 / 30201103_3 Crash when Status -> Profile is accessed


neuromancerxi:
Issue:

Accessing the Profile Menu causes the client to crash.

Affected Client Versions:

`30201103_1`
`30201103_3`

Steps to Replicate:

Be on either of the above mentioned client versions.
Log in a character.
Open The Main Menu (- key) > Select Status > Select Profile
Client will crash.

Additional Notes:

This was tested against Canary `ea135512f18bc2e0da51d154883c9c0d8dc1b37c`
Client version `30201006_1` does not exhibit this behavior when tested against the above version of Canary.

**_I have:_**

- [x] searched existing issues (http://project-topaz.com/issues/) to see if the issue has already been opened
- [x] checked the commit log to see if the issue has been resolved since my server was last updated

**_Additional Information_** (Steps to reproduce/Expected behavior) **:**

ghost:
[2020-11-14_18-02.zip](https://github.com/project-topaz/topaz/files/5541582/2020-11-14_18-02.zip)
Attached retail captures for accessing status->profile.

zach2good:
Fixed by: https://github.com/project-topaz/topaz/commit/be34888ce2c2727bf233d8acbaf8c3e2d284a194

PR #1529, opened 2020-11-25
Bugfix/profile crash #1524

ghost:
**_I affirm:_**

This fix hard-codes mastery rank as 1, and resolves issue with 0x63 status update packet sending incorrect gladiator rank.  As a result, the rank is also set to 0 (Gla).  Several references are added in comments from retail, but appear to have no impact on stability.

Job Info packet has increased in size by two bytes, which has resolved the profile crash, along with the above change.


ibm2431 (commented):
Looks reasonable enough to me; will let Zach have a say on 0x0C in `menu_merit` before hitting any buttons.

zach2good (approved):
Content seems fine, additional notes and translations are good. I would remove all the debug notes and the things marked as `November client update`, we have the git log and commit messages to let future readers know why it changes.

Similarly, those commits could be squashed together as something like: `Fix Nov 2020 update profile menu crash` etc. 👍

zach2good:
RE the "hacks", I'm fine with that, we don't have that content programmed anyway and there is a note for future readers 👍

zach2good:
`9 commits` 👀

Content looks fine, I'll just squash these 😆

